PR #5, opened 2020-10-16
Add Javascript target

twyatt:
## 🗒️  Notes

- Initial implementation of the JS target
    - Still has some work left to do (you'll see a number of `TODO()`s)
    - Wanted to lay the initial ground work to allow for integration of additional targets
- Dropped `discoverServices` from public API
    - This will be handled automatically on connect
    - _Might_ re-introduce function on the public API later

## 🔧  Features

- Observations remain active across connections
    - Any subscribers of `observe`s will remain active on connection drop and continue to pick up new characteristic changes upon reconnect (without having to re-subscribe)
- `Peripheral.connect` can be invoked from `Event.Disconnected`
    - Allows for simple reconnect logic


core/src/jsMain/kotlin/Central.kt
```diff
@@ -0,0 +1,44 @@
...
+    object {
+        val acceptAllDevices = true
+        val optionalServices = this@toDynamic.optionalServices
+    }
+} else {
+    object {
+        val optionalServices = this@toDynamic.optionalServices
+        val filters = this@toDynamic.filters
+    }
+}
```
cedrickcooke, line 44: JavaScript is wild.

core/src/jsMain/kotlin/Characteristic.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other == null || this::class.js != other::class.js) return false
+        other as Characteristic
+        if (uuid != other.uuid) return false
+        return true
+    }
+
+    override fun hashCode(): Int = uuid.hashCode()
```
cedrickcooke, line 25: Just checking. Is it correct that we're only using the `uuid` for equality?

core/src/jsMain/kotlin/Descriptor.kt
```diff
@@ -0,0 +1,21 @@
...
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other == null || this::class.js != other::class.js) return false
+        other as Descriptor
+        if (uuid != other.uuid) return false
+        return true
+    }
+
+    override fun hashCode(): Int = uuid.hashCode()
```
cedrickcooke, line 20: Same

core/src/jsMain/kotlin/external/BluetoothDevice.kt
```diff
@@ -0,0 +1,14 @@
+package com.juul.kable.external
+
+import org.w3c.dom.events.EventTarget
+
+// https://developer.mozilla.org/en-US/docs/Web/API/BluetoothDevice
```
cedrickcooke, line 5: Nit: `/** */` style comments please

Sidenote: Love that Mozilla documents this even though they don't support it yet.

core/src/jsMain/kotlin/external/Bluetooth.kt
```diff
@@ -0,0 +1,8 @@
+package com.juul.kable.external
+
+import kotlin.js.Promise
+
+// https://developer.mozilla.org/en-US/docs/Web/API/Bluetooth
```
cedrickcooke, line 5: Nit: `/** */` style comments please

core/src/jsMain/kotlin/external/BluetoothRemoteGATTCharacteristic.kt
```diff
@@ -0,0 +1,27 @@
+package com.juul.kable.external
+
+import com.juul.kable.UUID
+import org.khronos.webgl.DataView
+import org.w3c.dom.events.EventTarget
+import kotlin.js.Promise
+
+// https://developer.mozilla.org/en-US/docs/Web/API/BluetoothRemoteGATTCharacteristic
+// https://webbluetoothcg.github.io/web-bluetooth/#bluetoothgattcharacteristic-interface
```
cedrickcooke, line 9: Nit: `/** */` style comments please

core/src/jsMain/kotlin/external/BluetoothRemoteGATTDescriptor.kt
```diff
@@ -0,0 +1,15 @@
+package com.juul.kable.external
+
+import com.juul.kable.UUID
+import org.khronos.webgl.BufferDataSource
+import org.khronos.webgl.DataView
+import org.w3c.dom.events.EventTarget
+import kotlin.js.Promise
+
+// https://developer.mozilla.org/en-US/docs/Web/API/BluetoothRemoteGATTDescriptor
+// https://webbluetoothcg.github.io/web-bluetooth/#bluetoothremotegattdescriptor
```
cedrickcooke, line 10: Gonna stop commenting on these you get the point.

core/src/jsMain/kotlin/Characteristic.kt
twyatt: Being that these are on the public API (and sadly the `data` `ByteArray` is mutable, thanks Java), then no, it should be more than just UUID. I'll think about the best way to fix this (I'm hoping I can think of a clever way to not have to incur the expense of comparing the `ByteArray` contents.

twyatt: **EDIT**: I misread the code this was referring to.

The equality check really only needs to cover what is public, so, in this case it is still incorrect (as `descriptors` is public), but it won't be as expensive as I originally assumed (we aren't doing a content equality check of an arbitrary length `ByteArray`).

Granted we also won't know how many potential descriptors, but in most cases those will be small, and comparing them will involve comparing their UUIDs.

twyatt: Yikes. Thought about this further. I can see where this could really bite someone.

I'm using `Characteristic` (and friends) to hold a reference to the actual platform characteristic. This is a handy "optimization" as anytime the consumer of the library invokes any I/O function they do it with an object that has the direct reference to the type needed to execute the I/O operation.

Unfortunately, that also means that if the connection drops and is re-established, then this `Characteristic` will likely not work (worse yet, I believe on Android it would still work, but not on JS — inconsistent behavior I definitely want to avoid).

The alternative is the have `Characteristic` (and friends) be basic models (just holding the needed UUIDs). This would greatly simplify these classes (as no need to override `equals`/`hashCode`) **but** would mean that every I/O operation would require a lookup of the actual platform object to perform the I/O operation.

I believe the expense would be: O(n) + O(n) + O(n) = O(n) — a list search for service UUID, characteristic UUID and potentially descriptor UUID. All the `n`s would likely be small, but an expense nonetheless.

Although as @cedrickcooke is always quick to remind me: this is I/O, these tasks are likely significantly smaller than actually performing the I/O.

cedrickcooke: Was surprised to hear my PR comments uncovered anything major 😆  they all were surface level when I wrote them.

Not sure how difficult it would be/if even possible, but can you go for the best of both worlds? Characteristics carry what they need for a fast happy path, and if the connection drops and is re-established it will take the slow route to recalculate everything?

twyatt: Actually, I was thinking of something along those lines: lazy characteristics.

Essentially, you create a lazy `Characteristic` via a function on `Peripheral` (e.g. `Peripheral.characteristic(Uuid, Uuid)`). It would simply hold the UUID for the service and characteristic. When needed, the `Peripheral` object would determine that it was lazy and resolve the actual platform object (and store it on the lazy characteristic object).

On disconnect, would need to invalidate all lazy characteristics so that they would again be in their initial lazy state. This is where the hard part comes in, having an efficient way to invalidate.

Was thinking of having the peripheral keep a list of all "child" lazy characteristics, but that sounds like it could be error prone. As consumers of the library make create lots of characteristic objects not realize that a reference is being held to them.

Alternatively, could have an internal monotonic sequence ID. Then when connection drops mark what the sequence ID was at, any subsequent I/O requests with a characteristic with an "old" sequence ID would know to be re-resolved.

twyatt: Lazy characteristics also carry a huge advantage: you can set them up **before** a connection has been established.

This is a huge win, as setting up everything (including observations) but having to wait until **after** connected (and services discovered) is a huge pain.

It'd be amazing to have an API that allowed an entire peripheral representation (e.g. SensorTag) to be setup ahead of time. You could even setup the `observe` call before connected. When connected the library would take care of getting everything hooked up.

The challenging part will be ensuring that stacktraces (when things go wrong) are helpful. With non-lazy characteristics your stacktrace would be clear that your call to retrieve a non-existent characteristic originated from a specific part of your code. With lazy characteristics, the characteristic resolution will happen automatically so be a bit removed from where the characteristic was setup.

I could have a debug mode (similar to Coroutines), when enabled, it would build a stracktrace for the creation of the characteristic and store that. It could be thrown as the cause of any failures to make it clear where to creation originated.

cedrickcooke: Does Kotlin Multiplatform have an equivalent of `WeakReference`?

cedrickcooke: Love the sequence ID idea. Seems easy enough to implement and cheap enough at runtime.

core/src/jsMain/kotlin/external/BluetoothRemoteGATTDescriptor.kt
```diff
@@ -0,0 +1,15 @@
+package com.juul.kable.external
+
+import com.juul.kable.UUID
+import org.khronos.webgl.BufferDataSource
```
davidtaylor-juul, line 4: Unused import?

davidtaylor-juul (commented):
LGTM, pending sorting out that whole issue under discussion re holding direct references to the underlying system objects.

Lots of debug console logging which is immensely useful but will need to be removed at some point, recommend parking a ticket or issue to clean them out before going full production.

twyatt:
> LGTM, pending sorting out that whole issue under discussion re holding direct references to the underlying system objects.
> 
> Lots of debug console logging which is immensely useful but will need to be removed at some point, recommend parking a ticket or issue to clean them out before going full production.

Good thinking. Logged under #6.

twyatt:
Performed the discussed refactor, sans proper platform type caching.

Basically, **every** read/write will do a lookup for the platform type (e.g. characteristic or descriptor). In other words, all characteristic and descriptors objects on the public API will be lazy (and always lazy)...for now.

Decided it's best to go with the simple _inefficient_ approach at first. In a majority of cases the lookup will be very fast (as most users will only have a few characteristics/descriptors).

I figured it'd be better to optimize later _after_ all the platforms are up and running so we can see how the API settles.

I'll do some updates to the MacOS branch to see how this refactor effects that API and make sure it doesn't kill performance in Native threading land; then I'll re-request review.

twyatt:
> I'll do some updates to the MacOS branch to see how this refactor effects that API and make sure it doesn't kill performance in Native threading land; then I'll re-request review.

Was a bit more work than expected, so I'm going to postpone testing that platform for the time being and focus on it again when I start putting a PR up for it.

core/src/jsMain/kotlin/external/BluetoothRemoteGATTDescriptor.kt
twyatt: Apparently I need to get Kotlinter setup. 😄

core/src/jsMain/kotlin/Observers.kt
```diff
@@ -0,0 +1,108 @@
...
+
+import com.juul.kable.external.BluetoothRemoteGATTCharacteristic
+import kotlinx.coroutines.await
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.flow
+import org.khronos.webgl.DataView
+import org.w3c.dom.events.Event
+
+private typealias EventListener = (Event) -> Unit
```
cedrickcooke, line 11: Nit: I don't like this typealias, I'd rather just see `(Event) -> Unit` in the two places we use it.

core/src/jsMain/kotlin/Options.kt
```diff
@@ -0,0 +1,49 @@
...
+                if (other == null || this::class.js != other::class.js) return false
+                other as Services
+                if (!services.contentEquals(other.services)) return false
+                return true
+            }
+
+            override fun hashCode(): Int {
+                return services.contentHashCode()
+            }
+        }
```
cedrickcooke, line 27: Nit: Would it make sense to just use `List` instead of `Array` for these things? Would save a _lot_ of boilerplate.

cedrickcooke (approved):
Wish I knew more about web bluetooth, but it looks good to me.

core/src/jsMain/kotlin/Options.kt
sdonn3: Does that decisions have something to do with JS and how it's handled there?

sdonn3 (approved):
lgtm

core/src/jsMain/kotlin/Options.kt
twyatt: @sdonn3 is correct, the `Array` is a more natural collection for interop with Javascript (`List` works, but uses transpiled `List` class from stdlib rather than native JS types).

Relevant snippets of documentation below:

---

- `optionalServices[]`: An array of `BluetoothServiceUUID`s.

— [Bluetooth.requestDevice()](https://developer.mozilla.org/en-US/docs/Web/API/Bluetooth/requestDevice)

---

- `kotlin.Array` is mapped to JavaScript Array.
- Kotlin collections (`List`, `Set`, `Map`, etc.) are not mapped to any specific JavaScript type.

— [Representing Kotlin types in JavaScript](https://kotlinlang.org/docs/reference/js-to-kotlin-interop.html#representing-kotlin-types-in-javascript)

core/src/jsMain/kotlin/Observers.kt
twyatt: Being a `nit`, I'll probably keep it as-is. I generally go the `typealias` route when I need to make the high-order function nullable, it just feels like it reads better to me.

e.g. The following is a bit cluttered IMHO:

```
private fun Characteristic.createListener(): ((Event) -> Unit)? = { event ->
```

core/src/jsMain/kotlin/Options.kt
cedrickcooke: 👍  gotcha. A little unfortunate, but since we're doing this in `jsMain` and not `common` I think it's totally fine.

core/src/jsMain/kotlin/Observers.kt
cedrickcooke: Literally just an opinion, if you disagree feel free to totally ignore this and keep the typealias.

Issue #6, opened 2020-10-21
Remove and/or reduce logging


twyatt:
In the early development stages logging is essential for debugging, but prior to a major release it should be removed (or substantially reduced).

Perhaps an opt-in debug-mode option to enable logging?

- [x] Remove all `println` statements

PR #7, opened 2020-11-05
Add Android target

twyatt:
Dropped `wasConnected` from `Event.Disconnected`. Wanting to keep API simple/small and a "was connected" can easily be tracked by the library consumer (e.g. via `Event.Ready`). _Can re-add it back later if we are presented with a good use-case for it._

Android supports the same features that were implemented in the Javascript target (#5), namely:

- Observations remain active across connections
    - Any subscribers of `observe`s will remain active on connection drop and continue to pick up new characteristic changes upon reconnect (without having to re-subscribe)
- `Peripheral.connect` can be invoked from `Event.Disconnected` (or `State.Disconnected`)
    - Allows for simple reconnect logic


core/src/commonMain/kotlin/Descriptor.kt
```diff
@@ -4,13 +4,13 @@ import com.benasher44.uuid.Uuid
 import com.benasher44.uuid.uuidFrom
 
 public fun descriptorOf(
-    serviceUuid: String,
-    descriptorUuid: String,
-    characteristicUuid: String,
+    service: String,
+    characteristic: String,
+    descriptor: String,
```
twyatt, line 9: Renamed to not have `Uuid` postfix, as we'll soon support shorthand (e.g. `"heartbeat"`).

core/src/commonMain/kotlin/Central.kt
```diff
@@ -1,14 +1,6 @@
...
-        services: List<Uuid>? = null,
-    ): Scanner
-
-    public fun peripheral(
-        advertisement: Advertisement,
-    ): Peripheral
+public interface Central {
+    public fun scanner(): Scanner
+    public fun peripheral(advertisement: Advertisement): Peripheral
 }
```
cedrickcooke, line 6: I'm happy to see this, but I thought there was some bug that made interfaces not work. Did I misunderstand something earlier, or did something change?

core/src/commonMain/kotlin/Event.kt
```diff
@@ -2,44 +2,33 @@ package com.juul.kable
...
+    if (this == Event.Disconnected) action.invoke()
 }
 
 public suspend fun Event.onRejected(
-    action: suspend Event.Rejected.() -> Unit,
+    action: suspend () -> Unit,
 ) {
-    if (this is Event.Rejected) action.invoke(this)
+    if (this == Event.Rejected) action.invoke()
 }
```
cedrickcooke, line 34: Another mega-nit: `===` is more efficient than `==` if you're comparing against an `object`, avoids a method call.

core/src/androidMain/kotlin/BluetoothDevice.kt
```diff
@@ -0,0 +1,72 @@
...
+private fun BluetoothDevice.connectApi21(
+    context: Context,
+    state: MutableStateFlow<State>,
+): Connection? {
+    val callback = Callback(state)
+    val bluetoothGatt = connectGatt(context, false, callback) ?: return null
+
+    // Kludge: Explicitly set Connecting state so when Peripheral is suspending until Connected, it doesn't incorrectly
+    // see Disconnected before the connection request has kicked off the Connecting state (via Callback).
+    state.value = State.Connecting
```
cedrickcooke, line 36: I like seeing the comment on this, just honestly not sure if it's kludge -- seems super reasonable to me

core/src/androidMain/kotlin/Observers.kt
```diff
@@ -0,0 +1,67 @@
...
+        } finally {
+            if (observers.decrementAndGet(characteristic) < 1) {
+                peripheral.stopNotifications(characteristic)
+            }
+        }
+    }
+
+    suspend fun rewire() {
+        lock.withLock {
+            if (observers.isEmpty()) return
```
cedrickcooke, line 41: Not really necessary, the `forEach` is effectively a no-op if the list is empty.

core/src/androidMain/kotlin/Peripheral.kt
```diff
@@ -0,0 +1,291 @@
...
+
+    private val context = androidContext.applicationContext
+
+    private val job = SupervisorJob(parentCoroutineContext[Job]).apply {
+        invokeOnCompletion { _connection?.close() }
+    }
+    private val scope = CoroutineScope(parentCoroutineContext + job)
+
+    private val _state = MutableStateFlow<State>(State.Disconnected())
+    public override val state: Flow<State> = _state.asStateFlow()
```
cedrickcooke, line 70:
```suggestion
    public override val state: StateFlow<State> = _state.asStateFlow()
```

Also a Q: what do you get out of the `.asStateFlow()`, does that wrap it or something to prevent checking class?

core/src/androidMain/kotlin/Peripheral.kt
```diff
@@ -0,0 +1,291 @@
...
+        services.forEach { service ->
+            Log.d(TAG, "• Service: ${service.serviceUuid}")
+            service.characteristics.forEach { characteristic ->
+                Log.d(TAG, "    • Characteristic: ${characteristic.characteristicUuid}")
+                characteristic.descriptors.forEach { descriptor ->
+                    Log.d(TAG, "        • Descriptor: ${descriptor.descriptorUuid}")
+                }
+            }
+        }
+    }
```
cedrickcooke, line 158: Ooooh, fancy.

core/src/androidMain/kotlin/Peripheral.kt
```diff
@@ -0,0 +1,291 @@
...
+
+    public suspend fun requestMtu(mtu: Int): Unit = connection.request {
+        requestMtu(mtu)
+    }
```
cedrickcooke, line 162: This looks infinitely recursive, am I missing something?

core/src/androidMain/kotlin/Connection.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    val characteristicChanges = callback.onCharacteristicChanged
+        .map { (bluetoothGattCharacteristic, value) ->
+            CharacteristicChange(bluetoothGattCharacteristic.toLazyCharacteristic(), value)
+        }
+
+    private val lock = Mutex()
+
+    suspend inline fun <reified T> request(
+        crossinline action: BluetoothGatt.() -> Boolean,
```
sdonn3, line 44: Know this is only used internally but some documentation around 'action' seems warranted

sdonn3:
Trying to review this but there's a lot and Im losing concentration - ill take a look tomorrow morning

core/src/androidMain/kotlin/external/Constants.kt
```diff
@@ -0,0 +1,63 @@
+package com.juul.kable.external
+
+/**
+ * https://android.googlesource.com/platform/external/libnfc-nci/+/lollipop-release/src/include/hcidefs.h#447
+ */
```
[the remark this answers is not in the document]
twyatt: Oh, good call. Despite being a `nit`, really appreciate you pointing this out. Will fix. Thanks!

core/src/androidMain/kotlin/BluetoothDevice.kt
twyatt: I felt it was kludge because it's so disjointed from every other `State` propagation.
Right now, `Callback` which acts as the receiver of messages from Android, just converts the callback style state to the `Flow` `State`...**except** for this one particular instance.

Though, I don't expect to ever change this, so I think you're right that "kludge" probably is a mislabelling and the comment is sufficient to explain _why_ it's being done; it can be changed later and the comment will help guide considerations when/if changing it.

👍

core/src/androidMain/kotlin/Peripheral.kt
twyatt: Unless there is a good reason, I'm going to keep this as-is for now. I can open it up (i.e. "change to `StateFlow`" later) and it will be a compatible change, but I can't go the other way and have it be compatible.

`StateFlow` _probably_ is a better choice though, so I suspect the change will be made; but I'm leaning towards a super restrictive API this early on.

core/src/androidMain/kotlin/Peripheral.kt
twyatt: Hehe. Ya. Though I'll probably remove this with #6.

core/src/androidMain/kotlin/Peripheral.kt
twyatt: Oh, that is super confusing.

`connection.request {` makes `this` the underlying Android `BluetoothGatt`; so the `requestMtu` in the lambda is calling `BluetoothGatt`'s function, not the `Peripheral` function here.

I could change `connection.request` signature to have the parameter (rather than `this`) be the `BluetoothGatt`, but that would make everywhere we do this more verbose. Worth it?

core/src/androidMain/kotlin/Connection.kt
twyatt: Agreed. I'll add some documentation. 👍

core/src/commonMain/kotlin/Central.kt
twyatt: I forget exactly the issue I was running into. But some reason changes made this possible (I'm also very happy with being able to use `interface`s — still can't in some places, but the important bits now use `interface`):

I was previously storing the platform characteristic object within `Characteristic`s and then when passed around I'd have to cast to the appropriate type to get access to that `internal` property. With the `LazyCharacteristic` approach that is now used, characteristics have no references to their underlying platform characteristic objects. It is looked up lazily. So that pure Kotlin `Characteristic` can be easily used with an `interface` and no casting is necessary.

I think I was also running into issues with `Uuid` not being seen as `UUID` when on JVM. I reported https://youtrack.jetbrains.com/issue/KTIJ-405 and they renamed the ticket title with the warning ID, so now I can just suppress it!

https://github.com/JuulLabs/kable/blob/2bc92c307450ec3c3fe47bf56e18367566bd849c/core/src/androidMain/kotlin/PlatformDescriptor.kt#L6-L12

core/src/androidMain/kotlin/Peripheral.kt
cedrickcooke: Could you make it `this@request.requestMtu`? That makes it clear at sight that it's not the same `requestMtu` function being called, without leaking into the other uses of request.

core/src/androidMain/kotlin/Peripheral.kt
twyatt: `.asStateFlow()` returns an immutable `Flow`, which you can't cast to get around that safety.

core/src/androidMain/kotlin/Peripheral.kt
twyatt: Perfect, thanks for the suggestion. My train of thought was way too complicated. 🙃

core/src/androidMain/kotlin/Uuid.kt
```diff
@@ -0,0 +1,7 @@
+package com.juul.kable
+
+import com.benasher44.uuid.Uuid
+import com.benasher44.uuid.uuidFrom
+import java.util.UUID
+
+internal fun UUID.toUuid(): Uuid = uuidFrom(toString())
```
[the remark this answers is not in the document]
twyatt: Good catch. 👍

core/src/androidMain/kotlin/Peripheral.kt
cedrickcooke: That makes me so happy. Solves a big problem I had with LiveData

cedrickcooke (approved):

core/src/commonMain/kotlin/Peripheral.kt
```diff
@@ -2,47 +2,109 @@
...
-public expect class Peripheral {
+public interface Peripheral {
 
+    /**
+     * Provides a [Flow] of the [Peripheral]'s [State].
+     *
+     * After [connect] is called, the [state] will typically transition through the following [State]s:
+     *
+     * ```
+     *     connect()
```
sdonn3, line 21: Still a fan of your ascii art

sdonn3:
looks good to me 👍

sdonn3 (approved):

core/src/commonMain/kotlin/Peripheral.kt
cedrickcooke: Right? They're incredible.

PR #8, opened 2020-11-14
Add Mac OS X 64 target

twyatt:
Adds MacOS64 target. Code could definitely use some clean up (e.g. #6) and better mutability management, as it's currently doing a pretty delicate dance with object freezing. It's in a working state though, so I'd like to get this in and iterate on it later.

The API evolved slightly, so expect some follow up PRs soon to bring Android and JS targets into parity.

Dropped `event` Flow as it added significant complexity and didn't provide any functionality that couldn't be created by the consumer of the library (maybe we'll add it back in later if a strong use-case is found). Although `state` is conflated, the `Disconnected` or `Connected` states would be settled on, so you can react to those states and perform actions. If a state is missed (e.g. `Disconnected`) then it means, as an example, that a reconnect is already in progress. And reconnect is the likely use-case for reacting to `Disconnected` state.

Also, documentation is now available in #9.


core/src/macosX64Main/kotlin/Connection.kt
```diff
@@ -0,0 +1,31 @@
...
+import kotlinx.coroutines.sync.withPermit
+
+internal class Connection(
+    private val delegate: PeripheralDelegate,
+) {
+
+    val characteristicChanges: Flow<DidUpdateValueForCharacteristic> =
+        delegate.characteristicChanges
+
+    // Using Semaphore as Mutex never relinquished lock when multiple concurrent `withLock`s are executed.
```
cedrickcooke, line 15: What's the over/under on Mutex being bugged because it accidentally freezes itself somehow?

core/src/macosX64Main/kotlin/CentralManagerDelegate.kt
```diff
@@ -0,0 +1,136 @@
...
+
+    @Suppress("CONFLICTING_OVERLOADS") // https://kotlinlang.org/docs/reference/native/objc_interop.html#subclassing-swiftobjective-c-classes-and-protocols-from-kotlin
+    override fun centralManager(
+        central: CBCentralManager,
+        didFailToConnectPeripheral: CBPeripheral,
+        error: NSError?,
+    ) {
+        println("<- CentralManagerDelegate didFailToConnectPeripheral")
+        _connectionState.value = DidFailToConnect(didFailToConnectPeripheral.identifier, error)
+    }
```
cedrickcooke, line 100: I read the link but I'm still confused on this. Does Swift allow differentiating overloads by only the parameter names?

core/src/macosX64Main/kotlin/PeripheralDelegate.kt
```diff
@@ -0,0 +1,240 @@
...
+        peripheral: CBPeripheral,
+        didDiscoverServices: NSError?,
+    ) {
+        println("<- PeripheralDelegate didDiscoverServices")
+        _response.sendBlocking(DidDiscoverServices(peripheral.identifier, didDiscoverServices))
+    }
+
+    // todo: func peripheral(CBPeripheral, didDiscoverIncludedServicesFor: CBService, error: Error?)
+
+    /* Discovering Characteristics and their Descriptors */
```
cedrickcooke, line 112: When doing this style of comments, have you considered doing `//region Discovering Services` and `//endregion`?  I think they're a little uglier when just reading the code, but they enable folding in IntelliJ which can be worth it sometimes. 🤷  up to you.

core/src/macosX64Main/kotlin/Uuid.kt
```diff
@@ -0,0 +1,18 @@
...
+internal fun NSUUID.toUuid(): Uuid = uuidFrom(UUIDString)
+
+internal fun Uuid.toCBUUID(): CBUUID = CBUUID.UUIDWithString(toString())
+internal fun CBUUID.toUuid(): Uuid {
+    val fromUuid = when (UUIDString.length) {
+        4 -> "0000$UUIDString-0000-1000-8000-00805F9B34FB"
+        else -> UUIDString
+    }
+    return uuidFrom(fromUuid)
+}
```
cedrickcooke, line 18: A little confused what's happening here. Is CBUUID not guaranteed to be an actual UUID?

core/build.gradle.kts
```diff
@@ -15,22 +15,36 @@ kotlin {
...
             dependencies {
-                api(coroutines("core"))
+                api(coroutines("core", version = "1.3.9-native-mt-2"))
                 api(uuid())
             }
         }
 
+        val jsMain by getting {
+            dependencies {
+                implementation(coroutines("core"))
```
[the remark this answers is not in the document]
twyatt: How I understand this works (or at least hope; I'll see if I can dig a little to double check this):
The common code is pulling in `1.3.9-native-mt-2`, so all targets get that version, except in targets that we specify `1.4.1`, then Gradle knows that it can auto-bump to that version.
So targets that need `mt-2` use that version of Coroutines, the others that don't need that special branch of Coroutines we can tell Gradle to let those targets use `1.4.1`.

core/src/macosX64Main/kotlin/CentralManagerDelegate.kt
twyatt: When targeting Apple targets, yes, that is my understanding. You are overloading based on parameter names.

core/src/commonMain/kotlin/Peripheral.kt
```diff
@@ -33,38 +35,44 @@ public interface Peripheral {
...
      */
+    @Throws(CancellationException::class, IOException::class)
     public suspend fun read(
         characteristic: Characteristic,
     ): ByteArray
 
     /**
      * @throws NotReadyException if invoked without an established [connection][connect].
      */
+    @Throws(CancellationException::class, IOException::class)
```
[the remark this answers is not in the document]
twyatt: Was all an oversight. Needed the `@Throws` annotations as they are used for proper interop with Apple platforms (on the consumer side).

I'll go over these and make sure they're all correct.

`@Throws` is the correct approach AFAIK, so I'll drop the KDoc `@throws` comments.

core/src/macosX64Main/kotlin/Central.kt
```diff
@@ -0,0 +1,21 @@
+package com.juul.kable
+
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlin.coroutines.CoroutineContext
+
+public fun CoroutineScope.central(): AppleCentral = AppleCentral(coroutineContext)
+
+public class AppleCentral internal constructor(
```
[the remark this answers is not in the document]
twyatt: Haha...unfortunately, for consistence going to have to pass this opportunity by.

core/src/macosX64Main/kotlin/CentralManager.kt
```diff
@@ -0,0 +1,167 @@
...
+    private val scope = CoroutineScope(parentCoroutineContext + job + dispatcher)
+
+    internal val delegate = CentralManagerDelegate().freeze()
+    private val cbCentralManager = CBCentralManager(delegate, dispatcher.dispatchQueue)
+
+    fun scanForPeripheralsWithServices(
+        services: List<Uuid>?,
+        options: Map<Any?, *>?,
+    ) {
+        println("-> CentralManager.scanForPeripheralsWithServices")
```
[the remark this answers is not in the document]
twyatt: There are a few options, I might look into them, or just remove logging altogether (see #6).

core/src/macosX64Main/kotlin/Connection.kt
twyatt: I honestly don't know. I couldn't reproduce on JVM target.
I couldn't find an issue reported for it.

Before I went ahead and reported the issue, I wanted to have a small code snippet reproducing it but that was more work than trying `Semaphore` (and seeing that it just worked). 🤷 

I really should try to find some time to repro and report, but `Semaphore(1)` is currently faster than `Mutex` anyways. See https://kotlinlang.slack.com/archives/C1CFAFJSK/p1605617248151000 for details.

core/src/macosX64Main/kotlin/PeripheralDelegate.kt
twyatt: I definitely see the benefit, though I personally don't use folding at all — so, happy to have it be `region` but I slightly prefer it how it is just so it looks nicer. 🤷

core/src/macosX64Main/kotlin/Uuid.kt
twyatt: Doh. I'll look into that. This may be a poor choice of copy/paste. UUIDs can have shorthand notation, which Javascript supports; I may have copied this from my sample project on Javascript. Or Apple platforms also supports the shorthand notation; I'd be surprised if they did though, so I'll look into it.

Good catch.

core/build.gradle.kts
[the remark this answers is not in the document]
twyatt: Dependency insights:

<details>
<summary>Android</summary>

```
./gradlew :core:dependencyInsight --configuration androidReleaseCompileClasspath --dependency coroutines

> Task :core:dependencyInsight
org.jetbrains.kotlinx:kotlinx-coroutines-android:1.4.1
   variant "apiElements" [
      org.gradle.category                 = library (not requested)
      org.gradle.dependency.bundling      = external (not requested)
      org.gradle.jvm.version              = 6 (not requested)
      org.gradle.libraryelements          = jar (not requested)
      org.gradle.usage                    = java-api
      org.jetbrains.kotlin.localToProject = public (not requested)
      org.jetbrains.kotlin.platform.type  = jvm (compatible with: androidJvm)
      org.gradle.status                   = release (not requested)
   ]

org.jetbrains.kotlinx:kotlinx-coroutines-android:1.4.1
\--- androidReleaseCompileClasspath

org.jetbrains.kotlinx:kotlinx-coroutines-core:1.4.1
   variant "jvm-api" [
      org.gradle.libraryelements         = jar (not requested)
      org.gradle.usage                   = java-api
      org.jetbrains.kotlin.platform.type = jvm (compatible with: androidJvm)
      org.gradle.status                  = release (not requested)
   ]
   Selection reasons:
      - By conflict resolution : between versions 1.4.1 and 1.3.9-native-mt-2

org.jetbrains.kotlinx:kotlinx-coroutines-core:1.4.1
\--- org.jetbrains.kotlinx:kotlinx-coroutines-android:1.4.1
     \--- androidReleaseCompileClasspath

org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.9-native-mt-2 -> 1.4.1
\--- androidReleaseCompileClasspath

org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.4.1
   variant "jvm-api" [
      org.gradle.libraryelements         = jar (not requested)
      org.gradle.usage                   = java-api
      org.jetbrains.kotlin.platform.type = jvm (compatible with: androidJvm)
      org.gradle.status                  = release (not requested)
   ]

org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.4.1
\--- org.jetbrains.kotlinx:kotlinx-coroutines-core:1.4.1
     +--- androidReleaseCompileClasspath (requested org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.9-native-mt-2)
     \--- org.jetbrains.kotlinx:kotlinx-coroutines-android:1.4.1
          \--- androidReleaseCompileClasspath
```
</details>

<details>
<summary>Mac OS X 64</summary>

```
./gradlew :core:dependencyInsight --configuration macosX64MainImplementationDependenciesMetadata --dependency coroutines

> Task :core:dependencyInsight
org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.9-native-mt-2 (forced)
   variant "metadata-api" [
      org.gradle.usage                   = kotlin-metadata
      org.jetbrains.kotlin.platform.type = common
      org.gradle.status                  = release (not requested)
   ]

org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.9-native-mt-2
\--- macosX64MainImplementationDependenciesMetadata

org.jetbrains.kotlinx:kotlinx-coroutines-core:{strictly 1.3.9-native-mt-2} -> 1.3.9-native-mt-2
\--- macosX64MainImplementationDependenciesMetadata

org.jetbrains.kotlinx:kotlinx-coroutines-core-metadata:1.3.9-native-mt-2 (forced)
   variant "metadata-api" [
      org.gradle.usage                   = kotlin-metadata
      org.jetbrains.kotlin.platform.type = common
      org.gradle.status                  = release (not requested)
   ]

org.jetbrains.kotlinx:kotlinx-coroutines-core-metadata:1.3.9-native-mt-2
\--- org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.9-native-mt-2
     \--- macosX64MainImplementationDependenciesMetadata
```
</details>

<details>
<summary>JavaScript</summary>

```
./gradlew :core:dependencyInsight --configuration jsCompileClasspath --dependency coroutines

> Task :core:dependencyInsight
org.jetbrains.kotlinx:kotlinx-coroutines-core:1.4.1
   variant "jsLegacy-api" [
      org.gradle.usage                   = kotlin-api
      org.jetbrains.kotlin.js.compiler   = legacy
      org.jetbrains.kotlin.platform.type = js
      org.gradle.status                  = release (not requested)
   ]
   Selection reasons:
      - By conflict resolution : between versions 1.4.1 and 1.3.9-native-mt-2

org.jetbrains.kotlinx:kotlinx-coroutines-core:1.4.1
\--- jsCompileClasspath

org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.9-native-mt-2 -> 1.4.1
\--- jsCompileClasspath

org.jetbrains.kotlinx:kotlinx-coroutines-core-js:1.4.1
   variant "jsLegacy-api" [
      org.gradle.usage                   = kotlin-api
      org.jetbrains.kotlin.js.compiler   = legacy
      org.jetbrains.kotlin.platform.type = js
      org.gradle.status                  = release (not requested)
   ]

org.jetbrains.kotlinx:kotlinx-coroutines-core-js:1.4.1
\--- org.jetbrains.kotlinx:kotlinx-coroutines-core:1.4.1
     \--- jsCompileClasspath (requested org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.9-native-mt-2)
```
</details>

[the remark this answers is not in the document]
twyatt: The important bit is `!!` (strictly) on the Coroutines dependency for `macosX64`:

https://github.com/JuulLabs/kable/blob/4312acaec9fbc3baf3d352235a918fd692cf0c08/core/build.gradle.kts#L41-L46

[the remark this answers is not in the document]
twyatt: > Is this not considered a version conflict because `1.3.9-etc` is compatible with `1.4.1`?

To directly answer your original question: no, it's **not** because they are compatible. The versions just live completely independent on the platforms.

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -12,3 +12,8 @@ fun uuid(
     artifact: String = "uuid",
     version: String = "0.2.2"
 ): String = "com.benasher44:$artifact:$version"
+
+fun stately(
```
Phoenix7351, line 16: Does kotlin have a dependency manager built in that knows to look for this file? Or is this a convention used by a third party dependency management system?

Phoenix7351: Think I answered my own question. This is being used in gradle?

twyatt: > Think I answered my own question. This is being used in gradle?

Yup, `buildSrc` is a standard Gradle feature that allows us to add functionality to the Gradle build system. Here I'm just adding functions to produce the dependency Strings, and being in `buildSrc` allows them to be global.

core/src/commonMain/kotlin/Peripheral.kt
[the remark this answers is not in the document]
twyatt: Updated to at least have all the needed exceptions in the `@Throws`.

I went to JetBrains repos for guidance on how I should approach this, unfortunately none of the following projects use `@Throws`:
- kotlinx.coroutines
- kotlinx.serialization
- kotlinx.datetime

It looks like Square [uses the annotation in Okio](https://github.com/square/okio/blob/000e2dcbc6d855300afdf49b3e23faa42d56dd57/okio-files/src/commonMain/kotlin/okio/Filesystem.kt#L72-L73), so, I figure that's a good reference.

It looks like they use both `@Throws` and `@throws` to add further clarification in the KDoc.

I figure I'll follow that approach and add the exceptions to `@Throws` so that it's properly exposed on the public API for Apple-based platform consumers, then also have `@throws` KDoc for Exceptions that should have further explanation. For example, `NotReadyException` is specific to Kable so warrants further explanation whereas `IOException` is pretty much universally understood (so no need to add KDoc for it, IMHO).

core/build.gradle.kts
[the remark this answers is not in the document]
cedrickcooke: > Dependency insights:
> 
> Android
> Mac OS X 64
> JavaScript

Today I learned about `<details><summary>...</summary>...</details>` and it's so cool.

core/src/macosX64Main/kotlin/CentralManager.kt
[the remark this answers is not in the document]
cedrickcooke: Glad to see this is tracked. I'll ignore it for now regardless of the final result.

core/src/macosX64Main/kotlin/Connection.kt
cedrickcooke: I tried to do a little digging into the code, and my only insight is "weird that Mutex.kt is 400 lines but Semaphore.kt is 300 lines." Honestly, it might just be an implementation quality thing. 

Also, Semaphore.kt manages to fit a _beautiful_ ASCII-art comment in those 300 lines:

```kotlin

    /*
       The queue of waiting acquirers is essentially an infinite array based on the list of segments
       (see `SemaphoreSegment`); each segment contains a fixed number of slots. To determine a slot for each enqueue
       and dequeue operation, we increment the corresponding counter at the beginning of the operation
       and use the value before the increment as a slot number. This way, each enqueue-dequeue pair
       works with an individual cell. We use the corresponding segment pointers to find the required ones.

       Here is a state machine for cells. Note that only one `acquire` and at most one `release` operation
       can deal with each cell, and that `release` uses `getAndSet(PERMIT)` to perform transitions for performance reasons
       so that the state `PERMIT` represents different logical states.

         +------+ `acquire` suspends   +------+   `release` tries    +--------+                    // if `cont.tryResume(..)` succeeds, then
         | NULL | -------------------> | cont | -------------------> | PERMIT | (cont RETRIEVED)   // the corresponding `acquire` operation gets
         +------+                      +------+   to resume `cont`   +--------+                    // a permit and the `release` one completes.
            |                             |
            |                             | `acquire` request is cancelled and the continuation is
            | `release` comes             | replaced with a special `CANCEL` token to avoid memory leaks
            | to the slot before          V
            | `acquire` and puts    +-----------+   `release` has    +--------+
            | a permit into the     | CANCELLED | -----------------> | PERMIT | (RElEASE FAILED)
            | slot, waiting for     +-----------+        failed      +--------+
            | `acquire` after
            | that.
            |
            |           `acquire` gets   +-------+
            |        +-----------------> | TAKEN | (ELIMINATION HAPPENED)
            V        |    the permit     +-------+
        +--------+   |
        | PERMIT | -<
        +--------+  |
                    |  `release` has waited a bounded time,   +--------+
                    +---------------------------------------> | BROKEN | (BOTH RELEASE AND ACQUIRE FAILED)
                           but `acquire` has not come         +--------+
    */
```

twyatt: Things always work better when they have ASCII art in their documentation!

core/src/macosX64Main/kotlin/CentralManagerDelegate.kt
cedrickcooke: I am _incredibly impressed_ that Kotlin is able to successfully interop with this. It's stuff like that which kills a lot of inter-language compatibility.

core/src/macosX64Main/kotlin/PeripheralDelegate.kt
cedrickcooke: I'd leave it as is then. I know some people who love folding, but if the main dev on the project doesn't need it, no reason to clutter it up.

cedrickcooke (approved):

core/build.gradle.kts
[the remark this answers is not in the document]
twyatt: > Today I learned about `<details><summary>...</summary>...</details>` and it's so cool.

Right?! Just be sure to separate `<summary>..</summary>` with a [blank line before the actual details content](https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab).

QuantumRand (approved):

core/build.gradle.kts
```diff
@@ -15,22 +15,36 @@ kotlin {
     android {
         publishAllLibraryVariants()
     }
+    macosX64()
 
     sourceSets {
         val commonMain by getting {
             dependencies {
-                api(coroutines("core"))
+                api(coroutines("core", version = "1.4.1-native-mt"))
```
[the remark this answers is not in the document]
twyatt: The [`native-mt` is for Native **multithreading** support](https://github.com/Kotlin/kotlinx.coroutines/issues/462). The multithreading support in `Couroutines` is needed in Kable since values are being passed between threads via `Channel`s and `Flow`s (e.g. when we receive values from Mac OS's Core Bluetooth callback for a characteristic change, it is called on a dedicated dispatcher, then we pass it to a `Channel` so it can be picked up on the Coroutine context / Dispatcher being used on the `Flow` to observe the characteristic changes).

From what I understand, the Kotlin team is working on a new memory model for Native, which when available, Coroutines will adopt and we'll get the multithread support baked into Coroutines (and will no longer need a dedicated branch / release version of Coroutines), until then, we'll need to use the specialized Coroutines version.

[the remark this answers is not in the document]
twyatt: For details re: choice of version specification at the `common` source set level, see https://github.com/JuulLabs/kable/pull/8#discussion_r525545873 thread.

PR #11, opened 2020-11-23
Setup binary-compatibility-validator

twyatt:
Sets up API compatibility checks via [binary-compatibility-validator](https://github.com/Kotlin/binary-compatibility-validator).

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -12,6 +12,7 @@ plugins {
     id("com.vanniktech.maven.publish") version "0.13.0" apply false
     id("org.jetbrains.dokka") version "1.4.10.2"
     id("kotlinx-atomicfu") version "0.14.4" apply false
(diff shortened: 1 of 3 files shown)


twyatt:
Note to self: build _should_ be failing (but is passing). Need to investigate why the `binary-compatibility-validator` is not tapping into the `check` task. `apiDump` is also not being registered.

Initial suspicion: without `jvm` target, the Android multiplatform target may not be picked up, see also: https://github.com/Kotlin/binary-compatibility-validator/blob/6b19ba4a3abfa70c5906750f747bff1e0d4faa76/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt#L48-L68

davidtaylor-juul (approved):

QuantumRand (approved):

PR #12, opened 2020-12-02
Improve connection handling after a disconnect

twyatt:
Unifies clean up of connection when no longer needed.

It's a bit of a callback dance, but there are a number of causes for disconnection when we have to be sure we've cleaned up the underlying `BluetoothGatt` (via `close`):

- `STATE_DISCONNECTED` from Android (connection dropped, e.g. device went out of range)
- `Peripheral.disconnect` (developer has explicitly requested a disconnect, in `finally` we `dispose` to be sure we clean up if disconnect process is cancelled)
- `Peripheral`'s parent `CoroutineScope` is cancelled
- `Peripheral.connect` is cancelled (e.g. user leaves current screen and connect was being performed via `viewModelScope`)

### DIFF
--- a/core/src/androidMain/kotlin/Observers.kt
+++ b/core/src/androidMain/kotlin/Observers.kt
@@ -52,7 +52,13 @@ internal class Observers(
             }
         } finally {
             if (observers.decrementAndGet(characteristic) < 1) {
-                peripheral.stopNotifications(characteristic)
+                try {
+                    peripheral.stopNotifications(characteristic)
+                } catch (e: NotReadyException) {
+                    // Silently ignore as it is assumed that failure is due to connection drop, in which case Android
+                    // will clear the notifications.
+                    println("Stop notification failure ignored.")
+                }
             }
         }
     }
--- a/core/src/androidMain/kotlin/Peripheral.kt
+++ b/core/src/androidMain/kotlin/Peripheral.kt
@@ -29,7 +29,6 @@ import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.asStateFlow
-import kotlinx.coroutines.flow.catch
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.onEach
@@ -52,7 +51,7 @@ public class AndroidPeripheral internal constructor(
     private val context = androidContext.applicationContext
 
     private val job = SupervisorJob(parentCoroutineContext[Job]).apply {
-        invokeOnCompletion { _connection?.close() }
+        invokeOnCompletion { dispose() }
     }
     private val scope = CoroutineScope(parentCoroutineContext + job)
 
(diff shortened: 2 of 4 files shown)


core/src/androidMain/kotlin/Observers.kt
```diff
@@ -52,7 +52,13 @@ internal class Observers(
             }
         } finally {
             if (observers.decrementAndGet(characteristic) < 1) {
-                peripheral.stopNotifications(characteristic)
+                try {
+                    peripheral.stopNotifications(characteristic)
+                } catch (e: NotReadyException) {
+                    // Silently ignore as it is assumed that failure is due to connection drop, in which case Android
+                    // will clear the notifications.
+                    println("Stop notification failure ignored.")
```
Phoenix7351, line 60: Do you have a logger available instead of println?

core/src/androidMain/kotlin/Peripheral.kt
```diff
@@ -78,34 +77,39 @@ public class AndroidPeripheral internal constructor(
...
+        ) ?: throw ConnectionRejectedException()
+
+    /** Creates a connect [Job] that completes when connection is established, or failure occurs. */
     private fun createConnectJob(): Job = scope.launch(start = LAZY) {
         _ready.value = false
 
-        val connection =
-            bluetoothDevice.connect(context, _state) ?: throw ConnectionRejectedException()
-        _connection = connection
+        val connection = establishConnection().also { _connection = it }
```
Phoenix7351, line 91: Was the switch to using "also" here stylistic? Or is there a more concrete reason to do this instead of just assigning _connection?

core/src/androidMain/kotlin/Peripheral.kt
```diff
@@ -78,34 +77,39 @@ public class AndroidPeripheral internal constructor(
...
+        connection
+            .characteristicChanges
+            .onEach(observers.characteristicChanges::emit)
+            .launchIn(scope, start = UNDISPATCHED)
```
Phoenix7351, line 95: Are you intentionally not catching the errors that could be thrown from this?

core/src/androidMain/kotlin/Peripheral.kt
```diff
@@ -78,34 +77,39 @@ public class AndroidPeripheral internal constructor(
...
         }
 
         println("Ready")
         _ready.value = true
-    }.apply {
-        // fixme: Clear on disconnect (not at end of connect).
-        invokeOnCompletion { connectJob.value = null }
+    }
+
+    private fun dispose() {
```
Phoenix7351, line 110: Should this function name be a little more verbose to point out that it's the connection being disposed? Looking at this from the outside, I would have guessed that this function would be cleaning up some state for Peripheral at large--not just the Peripheral's connection.

core/src/androidMain/kotlin/gatt/Callback.kt
```diff
@@ -46,7 +46,7 @@ internal class Callback(
 ) : BluetoothGattCallback() {
 
     private var disconnectedAction: DisconnectedAction? = null
-    fun invokeOnDisconnected(action: () -> Unit) {
+    fun invokeOnDisconnected(action: DisconnectedAction) {
```
Phoenix7351, line 49: Just curious, is DisconnectedAction just an alias for () -> Unit?

core/src/androidMain/kotlin/Peripheral.kt
twyatt: The `establishConnection` call shortened the line a bit, so make it so it was still reasonably readable when using `also`.

**tl;dr** yes, pure stylistic

core/src/androidMain/kotlin/Observers.kt
twyatt: Tracking via #6.
https://github.com/JuulLabs/tuulbox/pull/22 might provide the logging capabilities we're looking for. 😉

core/src/androidMain/kotlin/Peripheral.kt
twyatt: Correct.

Before this PR, upstream the `Channel` was being closed with an exception, only to have it catch here as a signal to stop collecting.
Instead, just performing a standard close (without exception) upstream so that this stops collection gracefully.

core/src/androidMain/kotlin/gatt/Callback.kt
twyatt: Yup

https://github.com/JuulLabs/kable/blob/226102b47e271a434ed865e72c72053e35a0c8c1/core/src/androidMain/kotlin/gatt/Callback.kt#L37

core/src/androidMain/kotlin/Peripheral.kt
twyatt: Technically, the `Connection` object internally holds either a dispatcher or thread that needs to be cleaned up (and is done via callback from the `Connection.close`). So maybe `Connection.close` isn't verbose enough? Happy to change any names that make things more readable, not sure what to change it to though. Also not as important for `private` functions. 🤷

Phoenix7351 (approved):

sdonn3 (approved):
lgtm

PR #13, opened 2020-12-02
Support multiple calls to 'connect' for JavaScript target

twyatt:
Properly handles multiple calls to `connect` (to have all calls "suspend" — from JavaScript perspective it will have all `Promise`s waiting) until connected, or failure occurs.

### DIFF
--- a/core/src/jsMain/kotlin/Peripheral.kt
+++ b/core/src/jsMain/kotlin/Peripheral.kt
@@ -88,7 +88,7 @@ public class JsPeripheral internal constructor(
     }
 
     public override suspend fun connect() {
-        check(!job.isCancelled) { "Cannot connect, scope is cancelled for $this" }
+        check(job.isNotCancelled) { "Cannot connect, scope is cancelled for $this" }
         val job = connectJob ?: createConnectJob().also { connectJob = it }
         job.join()
     }
--- a/core/src/jsMain/kotlin/Peripheral.kt
+++ b/core/src/jsMain/kotlin/Peripheral.kt
@@ -11,12 +11,14 @@ import com.juul.kable.external.BluetoothRemoteGATTServer
 import com.juul.kable.external.string
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.CoroutineStart
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.await
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.filterNotNull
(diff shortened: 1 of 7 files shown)


core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -76,9 +80,23 @@ public class JsPeripheral internal constructor(
...
 
+    private fun dispose() {
+        observers.clear()
+        disconnectGatt()
+        unregisterDisconnectedListener()
+        _state.value = State.Disconnected()
+    }
+
+    public override suspend fun connect() {
+        check(!job.isCancelled) { "Cannot connect, scope is cancelled for $this" }
```
Phoenix7351, line 91: What's the style guideline on this? Other languages I've used discourage the leading !

Phoenix7351 (approved):

core/src/jsMain/kotlin/Peripheral.kt
twyatt: I think it's pretty standard in Kotlin.
Sometimes I see people write `.not` extension function, but I feel that is less standard, so less readable.

If you could point me to guides for other languages where leading `!` is discouraged that would be appreciated. Would be great to know _why_ other languages have adopted certain guidelines, to decide if we should adopt them as well: https://github.com/JuulLabs/kotlin-guides

Phoenix7351: I thought it was a Swift style thing, but I can't find it written anywhere. Maybe it's something I've adopted from previous teams. In either case, the reasoning was that it's really easy to miss the "!" when reading through the code--thereby reversing the meaning.

twyatt: > it's really easy to miss the "!" when reading through the code

Ya, totally makes sense. Is a valid reason to look for an alternate way to represent it.

I could add a simple `isNotCancelled` extension function? Having inversely named functions is the common approach I've seen in the Kotlin standard library.

sdonn3 (approved):

core/src/jsMain/kotlin/Peripheral.kt
sdonn3: The leading '!' is pretty standard throughout our Kotlin code; feel like it's totally acceptable here

Phoenix7351: Not trying to fight @sdonn3 on it, but--yeah, personally, I think it'd read more clearly with the extra function.

`check(job.isNotCancelled)`

`check(!job.isCancelled)`

twyatt: In this particular instance, I think the `!` is appropriate, but the extension function is also trivial to add, so I can go ahead and add it here.

PR #16, opened 2020-12-04
Bump Kotlin Coroutines dependency to 1.4.2

twyatt:
`SharedFlow` now works on `-native-mt` variant of Coroutines, so `BroadcastChannel` was dropped in favor of `SharedFlow`.


sdonn3 (approved):

core/src/macosX64Main/kotlin/Flow.kt
```diff
@@ -0,0 +1,9 @@
+package com.juul.kable
+
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.runBlocking
+
+internal fun <T> MutableSharedFlow<T>.sendBlocking(value: T) {
```
cedrickcooke, line 6: Nitpick:
```suggestion
internal fun <T> MutableSharedFlow<T>.emitBlocking(value: T) {
```

cedrickcooke (approved):

core/src/macosX64Main/kotlin/Flow.kt
twyatt: Oh, good catch. Will change. Thanks!

core/src/macosX64Main/kotlin/CentralManagerDelegate.kt
```diff
@@ -24,11 +22,8 @@ import kotlin.native.concurrent.freeze
 // https://developer.apple.com/documentation/corebluetooth/cbcentralmanagerdelegate
 internal class CentralManagerDelegate : NSObject(), CBCentralManagerDelegateProtocol {
 
-    // todo: MutableSharedFlow
-    private val _onDisconnected = BroadcastChannel<NSUUID>(1)
-    internal val onDisconnected: Flow<NSUUID>
-        // todo: Drop `get()` when `_onDisconnected` is a MutableSharedFlow.
-        get() = _onDisconnected.openSubscription().consumeAsFlow()
+    private val _onDisconnected = MutableSharedFlow<NSUUID>()
+    internal val onDisconnected = _onDisconnected.asSharedFlow()
```
Phoenix7351, line 26: Just curious, is a MutableSharedFlow not _already_ a shared flow?

twyatt: `MutableSharedFlow` is a mutable version of `SharedFlow`.

Using `asSharedFlow` exposes the underlying (`_onDisconnected`) **mutable** version of the flow, as a read-only `SharedFlow`.

core/src/macosX64Main/kotlin/Peripheral.kt
```diff
@@ -130,9 +131,12 @@ public class ApplePeripheral internal constructor(
                 _connection.value = it
             }
 
-            connection.characteristicChanges
+            connection
+                .delegate
+                .characteristicChanges
+                .takeWhile { it !== Closed }
```
Phoenix7351, line 137: I don't know what type Closed is, but did you intentionally use the referential operator? (!==)

twyatt: `Closed` is a "singleton" `object`, so the `!==` is a small optimization for that comparison. `!=` would've worked (slightly less efficient though).

https://github.com/JuulLabs/kable/blob/09619e55d9cdd8e62edad873c1d916bdd5aa1fab/core/src/macosX64Main/kotlin/PeripheralDelegate.kt#L79-L93

core/src/macosX64Main/kotlin/CentralManagerDelegate.kt
Phoenix7351: Gotcha. So privately you want it mutable and immutable internally.

core/src/macosX64Main/kotlin/Peripheral.kt
Phoenix7351: Got it. Thanks.

core/src/macosX64Main/kotlin/CentralManagerDelegate.kt
twyatt: Ya, basically isolating that emissions on the `Flow` can only come from within this class (and that downstream consumers can never also push items onto the stream — can only read from the stream).

Basically, just good practice to help keep the overall system simpler and less error prone.

PR #20, opened 2020-12-04
Replace RemoteException with GattRequestRejectedException

twyatt:
`RemoteException` is Android specific. `GattRequestRejectedException` will be thrown instead (which extends `BluetoothLeException`) allowing consumers to catch multiplatform friendly `BluetoothLeException` to handle this failure state.


sdonn3 (approved):

core/src/jsMain/kotlin/Exceptions.kt
```diff
@@ -1,3 +1,5 @@
+// ktlint-disable filename
```
cedrickcooke, line 1: I feel like we should either commit to the filename lint and not make individual exceptions like this, or we should disable the filename lint altogether. Nothing I'm willing to block a PR over though.

cedrickcooke (approved):

core/src/jsMain/kotlin/Exceptions.kt
twyatt: I agree, not excited about the one-off lint disabling. Unfortunately it may be the lesser of the evils here?
We have an `Exceptions.kt` in `common` that holds a variety of exceptions, whereas in the platform specific folders we only need to provide, in this case, `IOException`: so linters want the platform specific filenames to be `IOException.kt` but then that becomes inconsistent with the `common` naming.

In general I like the filename lint check, just feels like there are some edge cases where it's just not helpful, this being one of those cases. So the 1-line disabling in this one instance felt appropriate (lesser of the evils). 🤷

PR #23, opened 2020-12-09
Scanner support for JS target

davidtaylor-juul:
Closes #21 

Implements the (experimental) JS scanner functionality.

Requires recentish version of Chrome with `chrome://flags/#enable-experimental-web-platform-features` feature flag enabled

Issue #22, opened 2020-12-08
Ability to provide filters to the scanner


davidtaylor-juul:
The `Scanner` API presently exposes a flow of unfiltered peripherals which can of course be filtered after the fact. Internally the scan mechanisms typically provide for filtering *during* the scan process (part of the BLE spec) which is intended to be more efficient/save battery etc.

When instantiating the scanner, the API should allow injection of said filters to be fed through to the underlying systems.

burnhamd:
@twyatt I believe passing the service filter is required on iOS to be able to scan in background.  According to apple documentation 
_Your app can scan for Bluetooth devices in the background by specifying the bluetooth-central background mode. To do this, your app must explicitly scan for one or more services by specifying them in the serviceUUIDs parameter. The CBCentralManager scan option has no effect while scanning in the background._

Just wanted to mention this as it also blocks background scanning at least on iOS.

twyatt:
@burnhamd thanks for the heads up. This issue is getting a number of 👍  so I'll likely be re-prioritizing issues to be able to get to this one sooner.

PR #24, opened 2020-12-09
Remove Central and simpify/unify API

twyatt:
Decouples components from `Central` paradigm (`Central` has been removed).

API was simplified and unified by making the following changes:

- MacOS target holds an internal `CentralManager`
- Android target uses AndroidX startup to register an `Initializer` (and obtain application `Context`)

`Scanner` is no longer scoped, so `Peripheral` remains the only Coroutine scoped component. This allows library consumers more flexibility about the lifecycle of the various components.

_Updated README can be viewed [here](https://github.com/JuulLabs/kable/blob/twyatt/scope/README.md)._


twyatt:
Note to self: let #23 merge first, then be sure to properly setup `scanner` function for JS target.

twyatt:
General question for reviewers:

A `Scanner` can be obtained via `scanner()` function, but it is common to have a function "act as a constructor" (e.g. `Json` in kotlinx.serialization and similar examples can be found in Coroutines library). As such, I'd also prefer `Scanner()` function so that calling it "feels" like constructor invocation, but it clashes with interface with same name.

Q: Is `Scanner` function (that feels like a constructor) a good idea?
Q: If so, what should `Scanner` interface be renamed to?

QuantumRand (approved):

core/src/jsMain/kotlin/Bluetooth.kt
```diff
@@ -0,0 +1,26 @@
+package com.juul.kable
+
+import com.juul.kable.external.Bluetooth
+import kotlinx.coroutines.CoroutineScope
+import kotlin.js.Promise
+
+private val bluetooth: Bluetooth
+    get() = js("window.navigator.bluetooth") ?: error("Bluetooth unavailable")
```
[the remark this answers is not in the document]
twyatt: Oh. Good point. I don't believe this will work for Node.js.
Looks like Noble library does it via some kind of Node.js bindings mechanism.

https://github.com/noble/noble/blob/c4cd18a7a429bb832f6c4ca793be3faf9af884e9/lib/webbluetooth/bindings.js#L28-L45

...something to look into in the future. Tracking via #25.

README.md
```diff
@@ -7,60 +7,42 @@
...
-For both JavaScript and Apple-based targets, the `central` extension function is called without arguments:
+Scanning begins when the `advertisements` `Flow` is collected and stops when the `Flow` collection is terminated. A
+`Flow` terminal operator (such as [`first`]) may be used to scan until an advertisement is found that matches a desired
+predicated. 
 
 ```kotlin
-val central = scope.central()
+val example = scanner()
+    .advertisements
+    .first { it.name?.startsWith("Example") }
```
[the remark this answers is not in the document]
twyatt: For now, yes, all filtering is to be done at the `Flow` level. Not ideal, but good enough for MVP.

The plan is to add additional configuration support to `Scanner` (follow the builder pattern), similar to how `Json` works in kotlinx.serialization.

So, for platforms that support it, filtering could be done at the scanner level via an API similar to:

```kotlin
val scanner = Scanner {
    services = arrayOf(...)
    priority = Low
}
```

See also: #22

core/src/androidMain/kotlin/Peripheral.kt
```diff
@@ -108,6 +109,11 @@ public class AndroidPeripheral internal constructor(
     }
 
     private fun dispose() {
+        // Explicitly emit `Disconnecting` despite Android likely triggering the same state emission from `Callback`.
+        // Only distinct states are emitted and this emission ensures library consumers can get the state **before** the
+        // disconnect process begins (to cancel on-going I/O if needed, prior to connection being lost).
+        _state.value = State.Disconnecting
```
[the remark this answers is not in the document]
twyatt: Hmm, very good point. I'll investigate.

cedrickcooke:
> Q: Is `Scanner` function (that feels like a constructor) a good idea?
> Q: If so, what should the current `Scanner` interface be renamed to?

Answering this before I do my review.

I'm totally fine with a `fun Scanner(...): Scanner` to act as the constructor. Complete with the exact name match on the interface. This is exactly the pattern you see on a ton of coroutines stuff, such as `MutableStateFlow`.

twyatt:
> Complete with the exact name match on the interface.

🤦  I'm not sure what I was doing wrong, but when I tried this earlier it wasn't compiling. Seems to be working now, this was my preferred approach. So, I'm glad it seems to be working now. 😄

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -17,3 +17,9 @@ fun stately(
     module: String,
     version: String = "1.1.1-a1"
 ): String = "co.touchlab:stately-$module:$version"
+
+object androidx {
+    fun startup(
+        version: String = "1.0.0"
+    ) = "androidx.startup:startup-runtime:$version"
+}
```
cedrickcooke, line 25: ❤️

twyatt: Loved that this allowed me to remove `Context` from the API (so all primary operations can be defined in `common`). A clean/simple/unified API. ❤️

core/src/androidMain/kotlin/ApplicationContextInitializer.kt
```diff
@@ -0,0 +1,15 @@
+package com.juul.kable
+
+import android.content.Context
+import androidx.startup.Initializer
+
+internal lateinit var applicationContext: Context
```
cedrickcooke, line 6: Nit: add `private set`

core/src/androidMain/kotlin/Peripheral.kt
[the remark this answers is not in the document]
twyatt: This didn't end up accomplishing what I wanted it to anyways. Going to revert this addition.

For now, going to catch the failure in the SensorTag sample app (which shows _why_ I was hoping this change would help, but didn't):

```kotlin
private fun Peripheral.remoteRssi() = flow {
    while (true) {
        val rssi = rssi()
        Log.debug("RSSI: $rssi")
        emit(rssi)
        delay(1_000L)
    }
}.catch { cause ->
    // todo: Investigate better way of handling this failure case.
    // When disconnecting, we may attempt to read `rssi` causing a `NotReadyException` but the hope is that `remoteRssi`
    // Flow would already be cancelled by the time the `Peripheral` is "not ready" (doesn't seem to be the case).
    if (cause !is NotReadyException) throw cause
}
```

core/src/androidMain/kotlin/ApplicationContextInitializer.kt
```diff
@@ -0,0 +1,15 @@
...
+
+public class ApplicationContextInitializer : Initializer<Unit> {
```
cedrickcooke, line 9: Each `T` of `Initializer<T>` can only have one initializer, so providing it like this prevents library consumers from having their own `Initializer<Unit>` -- I feel like we shouldn't be consuming such a fundamental type.

I think my preferred option here is a just adding a marker `object Kable`, and just making this `public class KableInitializer : Initializer<Kable>`. As is, the public class name is kind of exposing the implementation detail.

core/src/androidMain/kotlin/Peripheral.kt
```diff
@@ -37,18 +36,20 @@ import kotlin.coroutines.CoroutineContext
...
+): Peripheral = AndroidPeripheral(coroutineContext, bluetoothDevice)
 
 public class AndroidPeripheral internal constructor(
     parentCoroutineContext: CoroutineContext,
-    androidContext: Context,
     private val bluetoothDevice: BluetoothDevice,
 ) : Peripheral {
 
-    private val context = androidContext.applicationContext
+    private val context = applicationContext
```
cedrickcooke, line 52: Nit: can we just remove this line now that we have it statically available?

core/src/androidMain/kotlin/ApplicationContextInitializer.kt
twyatt: Good callout and great suggestion!

README.md
```diff
@@ -9,60 +9,42 @@ with Bluetooth Low Energy devices.
...
 ```
 
 ### JavaScript
 
-On JavaScript, the standard paradigm is to request a specific peripheral rather than processing a stream of
-advertisements. Criterium (`Options`) such as expected service UUIDs on the peripheral and/or the peripheral's name may
-be specified. When `requestPeripheral` is called with the specified options, the browser shows the user a list of
-peripherals matching the criterium. The peripheral chosen by the user is then returned (as a `Peripheral` object).
+On JavaScript, rather than processing a stream of advertisements, a specific peripheral can be requested using the
+`CoroutineScope.requestPeripheral` extension function. Criterium (`Options`) such as expected service UUIDs on the
```
sdonn3, line 44: super nit: singular of criteria is criterion, though I think criteria is correct here. (Had to google it myself)

twyatt: Good catch. I was intending it to be plural (as it's representing multiple options), so "criteria" is what should be here.
Not sure where I got criterium from, looks to be a [bicycle race](https://en.wikipedia.org/wiki/Criterium). 🤷

core/src/androidMain/kotlin/Scanner.kt
```diff
@@ -13,13 +13,15 @@ public class ScanFailedException internal constructor(
...
+
 public class AndroidScanner internal constructor() : Scanner {
 
     private val bluetoothAdapter = BluetoothAdapter.getDefaultAdapter()
         ?: error("Bluetooth not supported")
 
-    public override val peripherals: Flow<Advertisement> = callbackFlow {
-        check(bluetoothAdapter.isEnabled) { "Bluetooth is disabled" } // todo: Exception type.
+    public override val advertisements: Flow<Advertisement> = callbackFlow {
+        check(bluetoothAdapter.isEnabled) { "Bluetooth is disabled" }
```
sdonn3, line 24: Is the todo no longer necessary?

twyatt: Ya, I thought about it, and `IllegalStateException` seems appropriate here.
We can change it later if we find a compelling reason, but I don't think it warrants a `todo`.

core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -27,6 +28,10 @@ import org.w3c.dom.events.Event as JsEvent
 
 private const val GATT_SERVER_DISCONNECTED = "gattserverdisconnected"
 
+public actual fun CoroutineScope.peripheral(
+    advertisement: Advertisement,
+): Peripheral = peripheral(advertisement.bluetoothDevice)
```
Phoenix7351, line 33: This is the same implementation as the Android version's Peripheral implementation. I haven't seen much expected/actual implementations. Is it possible for disparate platforms to share implementations?

sdonn3 (approved):

core/src/jsMain/kotlin/Peripheral.kt
twyatt: In this particular case. The `bluetoothDevice` property **name** is the same on both platforms, but the **type** is the platform specific type for each platform, so being able to have this as a single function in `common` is not possible unless the `BluetoothDevice` is defined as `expect`/`actual`.

Unfortunately, since the platforms diverge a bit in their representation of the underlying bluetooth device, it's probably better to have this function duplicated than deal with the potential complexity of trying to unify the bluetooth device platform representations.

cedrickcooke (approved):

PR #32, opened 2020-12-11
RSSI support for JS target

### LINKED ISSUE #19: RSSI support for JS target
Currently stubbed:

https://github.com/JuulLabs/kable/blob/09619e55d9cdd8e62edad873c1d916bdd5aa1fab/core/src/jsMain/kotlin/Peripheral.kt#L57-L59

davidtaylor-juul:
Closes #19 

Uses the experimental Web API `watchAdvertisements` to obtain the RSSI value from a peripheral advertisement packet.

### DIFF
--- a/core/src/jsMain/kotlin/Peripheral.kt
+++ b/core/src/jsMain/kotlin/Peripheral.kt
@@ -62,7 +62,7 @@ public class JsPeripheral internal constructor(
     public override suspend fun rssi(): Int = suspendCancellableCoroutine { continuation ->
         check(supportsAdvertisements) { "watchAdvertisements unavailable" }
 
-        var listener: ((JsEvent) -> Unit)? = null
+        lateinit var listener: (JsEvent) -> Unit
         val cleanup = {
             bluetoothDevice.removeEventListener(ADVERTISEMENT_RECEIVED, listener)
             // At the time of writing `unwatchAdvertisements()` remains unimplemented
@@ -75,7 +75,7 @@ public class JsPeripheral internal constructor(
             val event = it as BluetoothAdvertisingEvent
             cleanup()
             if (continuation.isActive) {
-                continuation.resume(event.rssi, null)
+                continuation.resume(event.rssi, onCancellation = null)
             }
         }
 
--- a/core/src/jsMain/kotlin/Peripheral.kt
+++ b/core/src/jsMain/kotlin/Peripheral.kt
@@ -4,6 +4,7 @@ package com.juul.kable
 
 import com.juul.kable.WriteType.WithResponse
 import com.juul.kable.WriteType.WithoutResponse
+import com.juul.kable.external.BluetoothAdvertisingEvent
 import com.juul.kable.external.BluetoothDevice
 import com.juul.kable.external.BluetoothRemoteGATTCharacteristic
 import com.juul.kable.external.BluetoothRemoteGATTDescriptor
@@ -20,6 +21,7 @@ import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.job
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.suspendCancellableCoroutine
 import org.khronos.webgl.ArrayBuffer
 import org.khronos.webgl.DataView
 import org.khronos.webgl.Int8Array
@@ -27,6 +29,7 @@ import kotlin.coroutines.CoroutineContext
 import org.w3c.dom.events.Event as JsEvent
 
 private const val GATT_SERVER_DISCONNECTED = "gattserverdisconnected"
+private const val ADVERTISEMENT_RECEIVED = "advertisementreceived"
 
 public actual fun CoroutineScope.peripheral(
     advertisement: Advertisement,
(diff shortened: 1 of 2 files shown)


core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -54,8 +57,36 @@ public class JsPeripheral internal constructor(
...
+                bluetoothDevice.unwatchAdvertisements()
+            }
+        }
+
+        listener = {
+            val event = it as BluetoothAdvertisingEvent
+            cleanup()
+            if (continuation.isActive) {
+                continuation.resume(event.rssi, null)
+            }
```
twyatt, line 79: There shouldn't be any need to check `isActive`.

Per the docs:

> Resumes this continuation with the specified `value` and calls the specified `onCancellation`
handler when either resumed too late (when continuation was already cancelled) or, although resumed
successfully (before cancellation), the coroutine's job was cancelled before it had a
chance to run in its dispatcher, so that the suspended function threw an exception
instead of returning this value.

So, I believe, that it will silently ignore call to `resume` if already cancelled.

```suggestion
            continuation.resume(event.rssi, onCancellation = null)
```

If `onCancellation` handler had been specified then it would be called if we called `resume` in a cancelled (`!isActive`) state. Since we hold the reference to the resource, we still clean up (so no need for us to use `onCancellation`).

The documentation gives the following example of when `onCancellation` is useful (when we're returned/resuming with a resource that will eventually need clean up — `onCancellation` can prevent leaks):

```kotlin
continuation.resume(resource) {
    resource.close()
}
```

core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -54,8 +57,36 @@ public class JsPeripheral internal constructor(
...
         get() = platformServices?.map { it.toDiscoveredService() }
 
-    public override suspend fun rssi(): Int {
-        TODO("Not yet implemented")
+    private val supportsAdvertisements = js("BluetoothDevice.prototype.watchAdvertisements") != null
+
+    public override suspend fun rssi(): Int = suspendCancellableCoroutine { continuation ->
+        check(supportsAdvertisements) { "watchAdvertisements unavailable" }
+
+        var listener: ((JsEvent) -> Unit)? = null
```
twyatt, line 65: If we want some null safety baked into the transpiled code (or at least on the JVM it compiles to bytecode that has `null` checks at usage sites), then you could change it to:

```suggestion
        lateinit var listener: (JsEvent) -> Unit
```

We're basically promising the compiler that we'll set the variable before we try to use it. On the JVM at least (and I assume on JS) it adds non-`null` checks at usage sites and throws `NullPointerException` if `null` when used.

twyatt (approved):
Quite the callback dance you had to do there. Looks great though!

twyatt:
Some thing that is a bit weird about acquiring RSSI in Web Bluetooth via advertisements is that peripherals stop advertising when a connection is established. So it must switch to some polling mechanism (similar to other platforms) where it communicates with peripheral to determine RSSI (and not advertisement packets)? 🤷

core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -54,8 +57,36 @@ public class JsPeripheral internal constructor(
...
+        val cleanup = {
+            bluetoothDevice.removeEventListener(ADVERTISEMENT_RECEIVED, listener)
+            // At the time of writing `unwatchAdvertisements()` remains unimplemented
+            if (bluetoothDevice.watchingAdvertisements && js("BluetoothDevice.prototype.unwatchAdvertisements") != null) {
+                bluetoothDevice.unwatchAdvertisements()
+            }
```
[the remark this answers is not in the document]
davidtaylor-juul: I really can't say what the behavior will be as `unwatchAdvertisements` is not yet implemented. I'm leaning towards the defensive approach in this circumstance.

core/src/jsMain/kotlin/Peripheral.kt
davidtaylor-juul: I read elsewhere in the docs (which I can't find now...) that `resume` throws an exception when called in the inactive state? Given that this callback can repeat I went with the defensive approach. It should never happen as we always remove the listener anyhow so is *probably* safe to ignore but "should never happen" has a way of happening when one least expects it...

davidtaylor-juul: https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/common/src/CancellableContinuationImpl.kt#L265

davidtaylor-juul:
> One thing that is a bit weird about acquiring RSSI in Web Bluetooth via advertisements is that peripherals stop advertising when a connection is established.

It's an interesting approach that JS has taken that's for sure

core/src/jsMain/kotlin/Peripheral.kt
twyatt: Hmm, the code you link implies that it throws when a `continuation` has already been resumed (not necessarily when inactive). We shouldn't be attempted subsequent resumes, so I don't think we'd be hit by this particular exception being thrown.

It's quite possible we do need the check you have, I just interpreted the documentation that we don't need it.
The documentation has also been wrong in the past.

🤷

twyatt: Spoke on Zoom, makes sense to keep `isActive` check.

QuantumRand (approved):

Issue #38, opened 2020-12-19
Throw appropriate exception when performing I/O on unknown characteristic


twyatt:
Currently, when attempting to write/read a characteristic that was not discovered then it will throw a `NoSuchElementException` (as thrown by `first`):

https://github.com/JuulLabs/kable/blob/c750de57fa6b0ea3dbd148dee3807b95e4ec1719/core/src/jsMain/kotlin/Peripheral.kt#L207-L212

Instead, we should throw an exception that clearly indicates what service or characteristic wasn't found.

twyatt:
Quick (potential) code to replace it with:

```kotlin
internal fun bluetoothRemoteGATTCharacteristicFrom(
    characteristic: Characteristic
): BluetoothRemoteGATTCharacteristic {
    val services = checkNotNull(platformServices) { "Services have not been discovered for $this" }
    val characteristics = services
        .firstOrNull { it.serviceUuid == characteristic.serviceUuid }
        ?.characteristics
        ?: throw <custom_exception>("Service ${characteristic.serviceUuid} not found")
    return characteristics
        .firstOrNull { it.characteristicUuid == characteristic.characteristicUuid }
        ?.bluetoothRemoteGATTCharacteristic
        ?: throw <custom_exception>("Characteristic ${characteristic.characteristicUuid} not found")
}
```

_Need to decide what exception (i.e. what to replace `<custom_exception>` with) to throw._

PR #41, opened 2020-12-23
Expanded BLE Advertisement Properties

davidtaylor-juul:
Expands the Advertisement class to include the majority of the information that can be extracted from BLE advertisement packets.

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -6,7 +6,7 @@ public final class com/juul/kable/Advertisement {
 	public final fun getRssi ()I
 	public final fun getTxPower ()Ljava/lang/Integer;
 	public final fun getUuids ()Ljava/util/List;
-	public final fun manufacturerData (S)[B
+	public final fun manufacturerData (I)[B
 	public final fun serviceData (Ljava/util/UUID;)[B
 	public fun toString ()Ljava/lang/String;
 }
@@ -134,8 +134,8 @@ public final class com/juul/kable/LazyDescriptor : com/juul/kable/Descriptor {
 }
 
 public final class com/juul/kable/ManufacturerData {
-	public fun <init> (S[B)V
-	public final fun getCode ()S
+	public fun <init> (I[B)V
+	public final fun getCode ()I
 	public final fun getData ()[B
 }
 
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -1,8 +1,13 @@
 public final class com/juul/kable/Advertisement {
-	public fun <init> (ILandroid/bluetooth/BluetoothDevice;)V
+	public fun <init> (Landroid/bluetooth/le/ScanResult;)V
 	public final fun getAddress ()Ljava/lang/String;
+	public final fun getManufacturerData ()Lcom/juul/kable/ManufacturerData;
 	public final fun getName ()Ljava/lang/String;
 	public final fun getRssi ()I
+	public final fun getTxPower ()Ljava/lang/Integer;
+	public final fun getUuids ()Ljava/util/List;
+	public final fun manufacturerData (S)[B
+	public final fun serviceData (Ljava/util/UUID;)[B
 	public fun toString ()Ljava/lang/String;
 }
 
@@ -126,6 +131,12 @@ public final class com/juul/kable/LazyDescriptor : com/juul/kable/Descriptor {
 	public fun getServiceUuid ()Ljava/util/UUID;
 	public fun hashCode ()I
 	public fun toString ()Ljava/lang/String;
+}
+
+public final class com/juul/kable/ManufacturerData {
+	public fun <init> (S[B)V
+	public final fun getCode ()S
+	public final fun getData ()[B
 }
 
 public final class com/juul/kable/NotReadyException : java/io/IOException {
--- a/core/src/androidMain/kotlin/Advertisement.kt
+++ b/core/src/androidMain/kotlin/Advertisement.kt
@@ -1,12 +1,16 @@
 package com.juul.kable
 
 import android.bluetooth.BluetoothDevice
+import android.bluetooth.le.ScanResult
+import android.os.ParcelUuid
 import com.benasher44.uuid.Uuid
 
 public actual class Advertisement(
-    public actual val rssi: Int,
-    internal val bluetoothDevice: BluetoothDevice,
+    internal val scanResult: ScanResult,
 ) {
+
+    internal val bluetoothDevice: BluetoothDevice
+        get() = scanResult.device
 
     public actual val name: String?
         get() = bluetoothDevice.name
@@ -17,17 +21,18 @@ public actual class Advertisement(
     override fun toString(): String =
         "Advertisement(name=$name, rssi=$rssi, bluetoothDevice=$bluetoothDevice)"
 
+    public actual val rssi: Int
+        get() = scanResult.rssi
+
     public actual val txPower: Int?
-        get() = TODO("Not yet implemented")
+        get() = scanResult.scanRecord?.txPowerLevel
 
     public actual val uuids: List<Uuid>
-        get() = TODO("Not yet implemented")
+        get() = scanResult.scanRecord?.serviceUuids?.map { it.uuid } ?: emptyList()
 
-    public actual fun serviceData(uuid: Uuid): ByteArray? {
-        TODO("Not yet implemented")
-    }
+    public actual fun serviceData(uuid: Uuid): ByteArray? =
+        scanResult.scanRecord?.serviceData?.get(ParcelUuid(uuid))
 
-    public actual fun manufacturerData(companyIdentifierCode: Short): ByteArray? {
-        TODO("Not yet implemented")
-    }
+    public actual fun manufacturerData(companyIdentifierCode: Short): ByteArray? =
+        scanResult.scanRecord?.getManufacturerSpecificData(companyIdentifierCode.toInt())
 }
--- a/core/src/androidMain/kotlin/Advertisement.kt
+++ b/core/src/androidMain/kotlin/Advertisement.kt
@@ -35,4 +35,7 @@ public actual class Advertisement(
 
     public actual fun manufacturerData(companyIdentifierCode: Short): ByteArray? =
         scanResult.scanRecord?.getManufacturerSpecificData(companyIdentifierCode.toInt())
+
+    public actual val manufacturerData: ManufacturerData?
+        get() = TODO("Not yet implemented")
 }
--- a/core/src/androidMain/kotlin/Advertisement.kt
+++ b/core/src/androidMain/kotlin/Advertisement.kt
@@ -18,9 +18,6 @@ public actual class Advertisement(
     public val address: String
         get() = bluetoothDevice.address
 
-    override fun toString(): String =
-        "Advertisement(name=$name, rssi=$rssi, bluetoothDevice=$bluetoothDevice)"
-
     public actual val rssi: Int
         get() = scanResult.rssi
 
@@ -43,4 +40,7 @@ public actual class Advertisement(
                 it.valueAt(0)
             )
         }
+
+    override fun toString(): String =
+        "Advertisement(name=$name, bluetoothDevice=$bluetoothDevice, rssi=$rssi, txPower=$txPower)"
 }
--- a/core/src/androidMain/kotlin/Advertisement.kt
+++ b/core/src/androidMain/kotlin/Advertisement.kt
@@ -6,7 +6,7 @@ import android.os.ParcelUuid
 import com.benasher44.uuid.Uuid
 
 public actual class Advertisement(
-    internal val scanResult: ScanResult,
+    private val scanResult: ScanResult,
 ) {
 
     internal val bluetoothDevice: BluetoothDevice
--- a/core/src/androidMain/kotlin/Advertisement.kt
+++ b/core/src/androidMain/kotlin/Advertisement.kt
@@ -30,13 +30,13 @@ public actual class Advertisement(
     public actual fun serviceData(uuid: Uuid): ByteArray? =
         scanResult.scanRecord?.serviceData?.get(ParcelUuid(uuid))
 
-    public actual fun manufacturerData(companyIdentifierCode: Short): ByteArray? =
-        scanResult.scanRecord?.getManufacturerSpecificData(companyIdentifierCode.toInt())
+    public actual fun manufacturerData(companyIdentifierCode: Int): ByteArray? =
+        scanResult.scanRecord?.getManufacturerSpecificData(companyIdentifierCode)
 
     public actual val manufacturerData: ManufacturerData?
         get() = scanResult.scanRecord?.manufacturerSpecificData?.takeIf { it.size() > 0 }?.let {
             ManufacturerData(
-                it.keyAt(0).toShort(),
+                it.keyAt(0),
                 it.valueAt(0)
             )
         }
--- a/core/src/androidMain/kotlin/Advertisement.kt
+++ b/core/src/androidMain/kotlin/Advertisement.kt
@@ -1,6 +1,7 @@
 package com.juul.kable
 
 import android.bluetooth.BluetoothDevice
+import com.benasher44.uuid.Uuid
 
 public actual class Advertisement(
     public actual val rssi: Int,
@@ -15,4 +16,18 @@ public actual class Advertisement(
 
     override fun toString(): String =
         "Advertisement(name=$name, rssi=$rssi, bluetoothDevice=$bluetoothDevice)"
+
+    public actual val txPower: Int?
+        get() = TODO("Not yet implemented")
+
+    public actual val uuids: List<Uuid>
+        get() = TODO("Not yet implemented")
+
+    public actual fun serviceData(uuid: Uuid): ByteArray? {
+        TODO("Not yet implemented")
+    }
+
+    public actual fun manufacturerData(companyIdentifierCode: Short): ByteArray? {
+        TODO("Not yet implemented")
+    }
 }
--- a/core/src/androidMain/kotlin/Advertisement.kt
+++ b/core/src/androidMain/kotlin/Advertisement.kt
@@ -37,5 +37,10 @@ public actual class Advertisement(
         scanResult.scanRecord?.getManufacturerSpecificData(companyIdentifierCode.toInt())
 
     public actual val manufacturerData: ManufacturerData?
-        get() = TODO("Not yet implemented")
+        get() = scanResult.scanRecord?.manufacturerSpecificData?.takeIf { it.size() > 0 }?.let {
+            ManufacturerData(
+                it.keyAt(0).toShort(),
+                it.valueAt(0)
+            )
+        }
 }
--- a/core/src/jsMain/kotlin/Advertisement.kt
+++ b/core/src/jsMain/kotlin/Advertisement.kt
@@ -17,7 +17,7 @@ public actual class Advertisement internal constructor(
         get() = bluetoothDevice.name
 
     public actual val rssi: Int
-        get() = advertisement.rssi ?: 0
+        get() = advertisement.rssi ?: Int.MIN_VALUE
 
     public actual val txPower: Int?
         get() = advertisement.txPower
@@ -38,13 +38,10 @@ public actual class Advertisement internal constructor(
         advertisement.manufacturerData.asDynamic().get(companyIdentifierCode.toString()) as? DataView
 
     public actual val manufacturerData: ManufacturerData?
-        get() = manufacturerDataEntries(advertisement.manufacturerData).firstOrNull()?.let { entry ->
+        get() = advertisement.manufacturerData.entries().iterable().firstOrNull()?.let { entry ->
             ManufacturerData(
                 entry[0] as Short,
                 (entry[1] as DataView).buffer.toByteArray()
             )
         }
 }
-
-private fun manufacturerDataEntries(jsObject: dynamic): Iterable<Array<Any?>> =
-    (jsObject.entries() as JsIterator<Array<Any?>>).iterable()
--- a/core/src/jsMain/kotlin/Advertisement.kt
+++ b/core/src/jsMain/kotlin/Advertisement.kt
@@ -3,12 +3,13 @@ package com.juul.kable
 import com.benasher44.uuid.Uuid
 import com.juul.kable.external.BluetoothAdvertisingEvent
 import com.juul.kable.external.BluetoothDevice
+import com.juul.kable.external.JsIterator
+import com.juul.kable.external.iterable
 import org.khronos.webgl.DataView
 
 public actual class Advertisement internal constructor(
     private val advertisement: BluetoothAdvertisingEvent,
 ) {
-
     internal val bluetoothDevice: BluetoothDevice
         get() = advertisement.device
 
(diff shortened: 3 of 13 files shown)


core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -205,11 +205,13 @@ public class JsPeripheral internal constructor(
...
-            .characteristics
+            .firstOrNull { it.serviceUuid == characteristic.serviceUuid }
+            ?.characteristics
+            ?: throw IOException("Service ${characteristic.serviceUuid} not found")
         return characteristics
-            .first { it.characteristicUuid == characteristic.characteristicUuid }
-            .bluetoothRemoteGATTCharacteristic
+            .firstOrNull { it.characteristicUuid == characteristic.characteristicUuid }
+            ?.bluetoothRemoteGATTCharacteristic
+            ?: throw IOException("Characteristic ${characteristic.characteristicUuid} not found")
```
davidtaylor-juul, line 214: This is unrelated - was added when debugging, can remove if not wanted

twyatt: Ya, please revert. We'll make this change (or similar) when we close #38.

twyatt:
Looking at this implemented, having to provide a manufacturer ID is quite awkward for a library user. Read through the BLE spec a bit and couldn't find much in the way that hints why Android and JavaScript went that route for their APIs.

How ugly is the JavaScript code for getting the bytes (without having to provide the manufacturer ID)?

core/src/jsMain/kotlin/external/BluetoothAdvertisingEvent.kt
```diff
@@ -6,6 +6,11 @@ import org.w3c.dom.events.Event
  * https://webbluetoothcg.github.io/web-bluetooth/#bluetoothadvertisingevent
  */
 internal abstract external class BluetoothAdvertisingEvent : Event {
-    val rssi: Int
     val device: BluetoothDevice
+    val uuids: Array<String>
+    val name: String?
+    val rssi: Int?
+    val txPower: Int?
+    val manufacturerData: Any?
```
twyatt, line 14: The docs say this is map-like. Does that mean that we could just mark it as a `Map` in Kotlin? Since I assume the functions and properties would align with that of a typical map (assuming Kotlin and JavaScript have similar naming for their `Map` properties/functions)?

Wondering if we have access to `keys` and could ask for element at index `0` or use `first`, to be able to ultimately grab the manufacturer ID.

davidtaylor-juul: The docs indicate they conform to `maplike` but I've found this to be only partially true for the current (experimental) implementation of both the service and manufacturer data. Typecasting to a `Map` in Kotlin fails as none of the expected methods are implemented.

The only methods that appear to be exposed are the `get` and a `forEach` iterator.

> having to provide a manufacturer ID is quite awkward for a library user

Perhaps, but the company identifiers are ratified in the BLE specification so are not difficult to find. Both the JS and Android framework designers also appear to agree this is their preferred approach. Apple disagrees but they often do lol. ¯\\_(ツ)_/¯ 

The JS code to obtain the `keys` would use the iterator to traverse the entries, very straightforward in native JS, somewhat convoluted from Kotlin directly.

twyatt: Bummer it's not a straightforward / easy cast to `Map`.
Rather than using `Any?` (and then later having to use `asDynamic`) could we make this `BluetoothManufacturerDataMap` type?

Then define `BluetoothManufacturerDataMap` as something similar to?:

```kotlin
internal interface BluetoothManufacturerDataMap {
    fun get(key: String): ByteArray
    fun forEach(action: (key: String, value: ByteArray) -> Unit): Unit
}
```

davidtaylor-juul: Found a more generic mechanism by hooking into the javascript Object iterator interface and bridging it as a Kotlin `Iterable`. This lets us access any of the javascript methods that return `Iterator` such as `keys()` and `entries()` and typecast them as `Iterable`. This looks to be a less brittle approach and doesn't require any custom types.

See https://github.com/JuulLabs/kable/pull/41/commits/f6995fd5972c9931ba52079734ce8cad614e397e

core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -250,4 +252,4 @@ public class JsPeripheral internal constructor(
     override fun toString(): String = "Peripheral(bluetoothDevice=${bluetoothDevice.string()})"
 }
 
-private fun ArrayBuffer.toByteArray(): ByteArray = Int8Array(this).unsafeCast<ByteArray>()
+internal fun ArrayBuffer.toByteArray(): ByteArray = Int8Array(this).unsafeCast<ByteArray>()
```
davidtaylor-juul, line 255: "Borrowed" this extension for the advertisement data parsing. Should we move it somewhere else?

twyatt: Yes please. Perhaps a new file names `Buffers.kt` or `Bytes.kt`?

core/src/jsMain/kotlin/Advertisement.kt
```diff
@@ -1,12 +1,50 @@
...
+        get() = manufacturerDataEntries(advertisement.manufacturerData).firstOrNull()?.let { entry ->
+            ManufacturerData(
+                entry[0] as Short,
+                (entry[1] as DataView).buffer.toByteArray()
+            )
+        }
 }
+
+private fun manufacturerDataEntries(jsObject: dynamic): Iterable<Array<Any?>> =
+    (jsObject.entries() as JsIterator<Array<Any?>>).iterable()
```
twyatt, line 50: We can remove the "unsafe" cast if we define the structure of the JS advertisement (in `BluetoothAdvertisingEvent.kt`):

```kotlin
internal interface BluetoothManufacturerDataMap {
    fun entries(): JsIterator<Array<Any?>>
}

/**
 * https://webbluetoothcg.github.io/web-bluetooth/#bluetoothadvertisingevent
 */
internal abstract external class BluetoothAdvertisingEvent : Event {
    val device: BluetoothDevice
    val uuids: Array<String>
    val name: String?
    val rssi: Int?
    val txPower: Int?
    val manufacturerData: BluetoothManufacturerDataMap
    val serviceData: Any?
}
```

```suggestion
    public actual val manufacturerData: ManufacturerData?
        get() = advertisement.manufacturerData.entries().iterable().firstOrNull()?.let { entry ->
            ManufacturerData(
                entry[0] as Short,
                (entry[1] as DataView).buffer.toByteArray()
            )
        }
}
```

Maybe we could also define the types of the entries so we don't have to do the casting when we create our `ManufacturerData`).

core/src/jsMain/kotlin/Advertisement.kt
```diff
@@ -1,12 +1,50 @@
...
+    private val advertisement: BluetoothAdvertisingEvent,
 ) {
+    internal val bluetoothDevice: BluetoothDevice
+        get() = advertisement.device
 
     public actual val name: String?
         get() = bluetoothDevice.name
+
+    public actual val rssi: Int
+        get() = advertisement.rssi ?: 0
```
twyatt, line 20: It's bizarre that JS has rssi as nullable. When would it ever not know RSSI for a received advertisement?! 🤷 

Not really liking falling back to `0`, as that is a valid rssi.
Perhaps we should fallback to `Interger.MIN_VALUE`)? Which is well outside the valid range (so could be conditionally checked for).

twyatt (commented):
Re-request review after you've addressed comments and then I'll test on Android and macOS.

core/src/jsMain/kotlin/Advertisement.kt
davidtaylor-juul: Good catch. With a 32bit integer, using `Int.MIN_VALUE` as the RSSI a distance calculation would give a result of the order of 10^10^9 meters, which is essentially infinite. Or NaN if the overflow isn't dealt with 😄

twyatt:
Looks great! I'll test the Android and macOS implementation today (then approve PR).
Aside from that, looks like you need to run `./gradlew apiDump`.

twyatt:
Tested on Android:
_Log snippet from scan_

```
Forerunner 935 (E6:45:6C:C3:47:24)
  code=135
  data=0a 83
  name=Forerunner 935, rssi=-90
  Service UUID 6a4e3e10-667b-11e3-949a-0800200c9a66
    data=null
nxg200-30AEA4F48CE8 (30:AE:A4:F4:8C:EA)
  code=null
  data=null
  name=nxg200-30AEA4F48CE8, rssi=-87
  Service UUID 000000ff-0000-1000-8000-00805f9b34fb
    data=null
  Service UUID 000000ff-0000-1000-8000-00805f9b34fb
    data=null
N0404 (C9:F8:80:E1:2D:E6)
  code=null
  data=null
  name=N0404, rssi=-75
  Service UUID 0000feaf-0000-1000-8000-00805f9b34fb
    data=10 01 00 02 5a 23 09 00 b1 b6 c7 00 00 66 16 64 00
```

twyatt:
macOS test:

```
Forerunner 935 (0BA3628E-1CF1-4E94-8A53-D1F1A4F696B6)
  code=null
  data=null
  name=Forerunner 935, rssi=-82
  Service UUID 6a4e3e10-667b-11e3-949a-0800200c9a66
    data=null

Forerunner 935 (0BA3628E-1CF1-4E94-8A53-D1F1A4F696B6)
  code=null
  data=null
  name=Forerunner 935, rssi=-79

null (D44A5F25-A784-4F00-A557-63BA894EEA21)
  code=null
  data=null
  name=null, rssi=-48
  Service UUID 0000fe9f-0000-1000-8000-00805f9b34fb
    data=00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
```

twyatt:
Possibly just different devices, but never say `code` or `data` populated from the macOS scan.

twyatt:
Build is failing due to unused import, `./gradlew formatKotlin` should fix you up.

core/src/androidMain/kotlin/Advertisement.kt
```diff
@@ -15,4 +20,27 @@ public actual class Advertisement(
 
     override fun toString(): String =
         "Advertisement(name=$name, rssi=$rssi, bluetoothDevice=$bluetoothDevice)"
```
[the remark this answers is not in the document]
twyatt: Would also probably be a good idea to add `toString` overrides (for `Advertisement`) for Apple and JS targets.

davidtaylor-juul:
> Possibly just different devices, but never saw `code` or `data` populated on the macOS scan

Possible, as Advertisment data can differ between packets even for the same device, but could be a bug. Investigating further and will add some tests.

davidtaylor-juul:
> > Possibly just different devices, but never saw `code` or `data` populated on the macOS scan
> 
> Possible, as Advertisment data can differ between packets even for the same device, but could be a bug. Investigating further and will add some tests.

There was a bug, had `slice(2..size)` instead of `slice(2 until size)` which *should* have been throwing exceptions. Perhaps that test app is ignoring them?

Added some unit tests to cover that data parsing logic on macOS.

core/src/macosX64Main/kotlin/Advertisement.kt
```diff
@@ -14,4 +21,39 @@ public actual data class Advertisement(
...
+        get() = (data[CBAdvertisementDataServiceUUIDsKey] as? List<CBUUID>)?.map { it.toUuid() } ?: emptyList()
+
+    public actual fun serviceData(uuid: Uuid): ByteArray? =
+        serviceDataAsNSData(uuid)?.toByteArray()
+
+    public fun serviceDataAsNSData(uuid: Uuid): NSData? =
+        (data[CBAdvertisementDataServiceDataKey] as? Map<CBUUID, NSData>)?.get(uuid.toCBUUID())
+
+    public actual fun manufacturerData(companyIdentifierCode: Int): ByteArray? =
+        manufacturerData?.takeIf { (it.code == companyIdentifierCode) }?.let { it.data }
```
twyatt, line 38: Unnecessary `let`.

```suggestion
        manufacturerData?.takeIf { (it.code == companyIdentifierCode) }?.data
```

core/src/macosX64Main/kotlin/Advertisement.kt
```diff
@@ -14,4 +21,39 @@ public actual data class Advertisement(
...
+
+    public actual val manufacturerData: ManufacturerData?
+        get() = manufacturerDataAsNSData?.toByteArray()?.toManufacturerData()
+
+    public fun manufacturerDataAsNSData(companyIdentifierCode: Int): NSData? =
+        manufacturerData(companyIdentifierCode)?.toNSData()
+
+    public val manufacturerDataAsNSData: NSData?
+        get() = data[CBAdvertisementDataServiceDataKey] as? NSData
```
twyatt, line 47: Wrong key. This is why we were seeing `null` for `code` and `data`.

```suggestion
        get() = data[CBAdvertisementDataManufacturerDataKey] as? NSData
```

Thankfully the tests led me to investigate what data we were actually getting back, and voila! 😄

twyatt (commented):
Amazing work! 💯 

Mostly nitpicks left (which you can optionally address) and [fix macOS key used for manufacturer data](https://github.com/JuulLabs/kable/pull/41#discussion_r549913043). Tested with correct key and it worked as expected (got back `code` and `data`).

core/src/macosX64Test/kotlin/AdvertisementTest.kt
```diff
@@ -0,0 +1,59 @@
...
+    @Test
+    fun manufacturerDataParsingWithFewerThan2BytesIsNull() {
+        val advertisement = fakeAdvertisement(
+            ubyteArrayOf(0x01u)
+        )
+        val data = advertisement.manufacturerData
+        assertNull(data)
+    }
+
+    private fun fakeAdvertisement(manufacturerBytes: UByteArray): Advertisement =
```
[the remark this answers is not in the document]
davidtaylor-juul: 👍 was super messy to specify the data using signed bytes. The lack of unsigned is part of *The Java Hangover* and we should not shy away from it for multi-platform targets IMHO. 😄

core/src/macosX64Main/kotlin/Advertisement.kt
davidtaylor-juul: OMG thanks for spotting this. I remember making sure to use the correct key originally and must have accidentally messed it up at some point 😭 Apple's bag-of-things API for advertisements is so vexing.

davidtaylor-juul: I got it wrong in the tests also, at least my stupidity is consistent! 😑

core/src/macosX64Main/kotlin/Advertisement.kt
davidtaylor-juul: Good catch

twyatt (approved):
💯 🚀

core/api/core.api
```diff
@@ -1,8 +1,13 @@
...
-	public fun <init> (ILandroid/bluetooth/BluetoothDevice;)V
+	public fun <init> (Landroid/bluetooth/le/ScanResult;)V
 	public final fun getAddress ()Ljava/lang/String;
+	public final fun getManufacturerData ()Lcom/juul/kable/ManufacturerData;
 	public final fun getName ()Ljava/lang/String;
 	public final fun getRssi ()I
+	public final fun getTxPower ()Ljava/lang/Integer;
+	public final fun getUuids ()Ljava/util/List;
+	public final fun manufacturerData (I)[B
+	public final fun serviceData (Ljava/util/UUID;)[B
```
Phoenix7351, line 10: For my own curiosity, what are these characters after the function definition? "(Ljava/util/UUID;)[B"

Something to do with transpilation, I assume, but what do they mean?

core/src/androidMain/kotlin/Advertisement.kt
```diff
@@ -1,18 +1,46 @@
...
 
 import android.bluetooth.BluetoothDevice
+import android.bluetooth.le.ScanResult
+import android.os.ParcelUuid
+import com.benasher44.uuid.Uuid
 
 public actual class Advertisement(
-    public actual val rssi: Int,
-    internal val bluetoothDevice: BluetoothDevice,
+    private val scanResult: ScanResult,
```
Phoenix7351, line 9: Nice.

core/src/jsMain/kotlin/Bytes.kt
```diff
@@ -0,0 +1,6 @@
+package com.juul.kable
+
+import org.khronos.webgl.ArrayBuffer
+import org.khronos.webgl.Int8Array
+
+internal fun ArrayBuffer.toByteArray(): ByteArray = Int8Array(this).unsafeCast<ByteArray>()
```
Phoenix7351, line 6: Also nice.

core/src/jsMain/kotlin/external/JsIterator.kt
```diff
@@ -0,0 +1,25 @@
+package com.juul.kable.external
+
+internal external interface JsIterator<T> {
+    fun next(): JsIteratorResult<T>
+}
+
+internal external interface JsIteratorResult<T> {
```
Phoenix7351, line 7: "internal external" Haha. I get it, but funny.

Phoenix7351: Question on this, though: "external" means, "the implementation is defined outside of kotlin," right?

If that's the case, how're we indicating to kotlin where that definition is?

core/src/jsMain/kotlin/external/JsIterator.kt
```diff
@@ -0,0 +1,25 @@
...
+    val done: Boolean
+    val value: T?
+}
+
+internal fun <T> JsIterator<T>.iterable(): Iterable<T> {
```
Phoenix7351, line 12: Sucks that we need to manually define this.

core/src/macosX64Test/kotlin/AdvertisementTest.kt
```diff
@@ -0,0 +1,59 @@
...
+import platform.CoreBluetooth.CBPeripheral
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+import kotlin.test.assertNull
+import kotlin.test.assertTrue
+
+class AdvertisementTest {
+    @Test
+    fun manufacturerDataCanContainData() {
```
Phoenix7351, line 15: For ease of understandability, I like to give my test function names a particular format:

fun [name of item being tested]_[input conditions]_[expected results]

I think in this case, it would be something like:

fun manufacturerData_validAdvertisement_returnsValidData()

Not saying that we need to use that exact format, but doing something akin to that makes it really easy to quickly understand what is being tested, under what conditions, and what the expected results are--a very useful thing to understand should the test ever fail.

core/api/core.api
twyatt: Shorthand notation for Java bytecode.

[Here](http://homepage.divms.uiowa.edu/~slonnegr/wpj/JNI.pdf) is a short reference (for some of the symbols):
Screen Shot 2021-01-04 at 10 47 54 AM

twyatt: > Something to do with transpilation

Nothing to do w/ transpilation actually. Transpilation is for translating Kotlin to JavaScript.
There are Java (JVM) bytecode API dumps (not JavaScript at all).

core/src/jsMain/kotlin/external/JsIterator.kt
twyatt: > If that's the case, how're we indicating to kotlin where that definition is?

We actually don't have to. Since the Kotlin is transpiled to JavaScript, it will be running in the actually JavaScript environment (in the browser in our case). So we're simply promising the compiler that this function signature exists, and for us, it will exist when everything is transpiled. So the transpiled code will call this `interface` and as long as it actually matches the signature that does in fact exist, then all is good.

core/src/jsMain/kotlin/external/JsIterator.kt
twyatt: My hope is the new IR compiler (that JetBrains is working on) will provide this for us. We shall see though.

core/api/core.api
Phoenix7351: Oh, why is this needed? The compiler can't figure this out for itself?

core/src/macosX64Test/kotlin/AdvertisementTest.kt
twyatt: I kind of go back and forth on what the best format for test function names are.

> `[name of item being tested][input conditions][expected results]`

I used to subscribe the this format as well, but found it too restrictive at times.

I'm happy to keep this flexible (and let the developer decide what the best way to convey what the test is doing). As seeing that `manufacturerDataCanContainData` failed, seems reasonable enough to me to know what went wrong.

Happy if it is changed as well though. (I personally am fine with either naming scheme)

Phoenix7351: Well, what I find useful about using the "_", is that you're able to easily parse out all the components of the test (system under test, input conditions, expected results).

Camelcasing this info together will still contain these components, but it's more difficult to determine what they are, and could be interpreted different ways. 

In what way have you found the "_" format restrictive?

core/src/jsMain/kotlin/external/JsIterator.kt
davidtaylor-juul: FYI this protocol is part of the ECMAScript 2015 which is why we can define it as `external` and fully expect it to exist in the target runtime (for conforming browsers).

https://docs.w3cub.com/javascript/iteration_protocols#The_iterator_protocol

core/api/core.api
twyatt: These API dumps (`.api`) are for our reference (not the compiler's). So we can observe how the API has changed. Simplifies the process of adhering to semver for our releases.

core/src/macosX64Test/kotlin/AdvertisementTest.kt
twyatt: > In what way have you found the "_" format restrictive?

Some very small/simplified tests don't adhere to the structure:

1. name of item being tested
2. input conditions
3. expected results

davidtaylor-juul: As these are the first tests in this codebase I'll go ahead and switch to the `fun [name of item being tested]_[input conditions]_[expected results]` naming scheme to set precedent.

And thus defer further argument to our future selves. 🤠

Phoenix7351: > > In what way have you found the "_" format restrictive?
> 
> Some very small/simplified tests don't adhere to the structure:
> 
> 1. name of item being tested
> 2. input conditions
> 3. expected results

That's true. This very example could fit that case. Like, in, "manufacturerData_validAdvertisement_returnsValidData()", "validAdvertisement" might be superfluous (if there is a use case where there could be an invalid advertisement, I'd say it isn't). In that scenario, I think it'd be fine to shorten to `fun manufacturerData_returnsValidData()` as this still clearly denotes what is being tested and the outcome--with the input being understood as "happy path default conditions."

core/api/core.api
Phoenix7351: Gotcha. Thanks!

core/src/macosX64Test/kotlin/AdvertisementTest.kt
```diff
@@ -0,0 +1,61 @@
...
+            )
+        )
+        val data = advertisement.manufacturerData
+        assertNotNull(data)
+        assertEquals(data.code, 0x05c3)
+        assertTrue(data.data.isEmpty())
+    }
+
+    @Test
+    fun manufacturerData_advertisementWithFewerThanTwoBytes_isNull() {
```
Phoenix7351, line 45: https://www.youtube.com/watch?v=pso4SWM75ew

Phoenix7351 (approved):

core/src/macosX64Test/kotlin/AdvertisementTest.kt
twyatt: Good example. Ultimately, I'm pretty flexible on the naming of test functions. 😄

PR #46, opened 2020-12-28
Fix propagation of ConnectionLostException on connect

### LINKED ISSUE #44: Force disconnecting peripheral causes crash by ConnectionLostException
Steps to reproduce:
Start connecting to peripheral by method peripheral.connect and quickly stop Bluetooth server on the peripheral side.

Stack trace:
`com.juul.kable.ConnectionLostException
        at com.juul.kable.PeripheralKt$suspendUntilConnected$2.invokeSuspend(Peripheral.kt:220)
        at com.juul.kable.PeripheralKt$suspendUntilConnected$2.invoke(Unknown Source:10)
        at kotlinx.coroutines.flow.FlowKt__TransformKt$onEach$$inlined$unsafeTransform$1$2.emit(Collect.kt:134)
        at kotlinx.coroutines.flow.StateFlowImpl.collect(StateFlow.kt:348)
        at kotlinx.coroutines.flow.StateFlowImpl$collect$1.invokeSuspend(Unknown Source:12)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)`

twyatt:
Was previously using `launch` internally to establish connection. An early failure (before `join` was called) would cause the failure to propagate to the parent scope (rather than the `connect` caller). Switched to using `async` so that failures are properly propagated on call to `await` (within `connect`).

Fixes #44

### DIFF
--- a/core/src/androidMain/kotlin/Peripheral.kt
+++ b/core/src/androidMain/kotlin/Peripheral.kt
@@ -22,15 +22,16 @@ import kotlinx.atomicfu.updateAndGet
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.CoroutineStart.LAZY
 import kotlinx.coroutines.CoroutineStart.UNDISPATCHED
+import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.async
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.onEach
-import kotlinx.coroutines.launch
 import kotlin.coroutines.CoroutineContext
 
 private val clientCharacteristicConfigUuid = uuidFrom(CLIENT_CHARACTERISTIC_CONFIG_UUID)
@@ -68,7 +69,7 @@ public class AndroidPeripheral internal constructor(
     private val connection: Connection
         inline get() = _connection ?: throw NotReadyException(toString())
 
-    private val connectJob = atomic<Job?>(null)
+    private val connectJob = atomic<Deferred<Unit>?>(null)
 
     private val _ready = MutableStateFlow(false)
     internal suspend fun suspendUntilReady() {
@@ -83,7 +84,7 @@ public class AndroidPeripheral internal constructor(
         ) ?: throw ConnectionRejectedException()
 
     /** Creates a connect [Job] that completes when connection is established, or failure occurs. */
-    private fun createConnectJob(): Job = scope.launch(start = LAZY) {
+    private fun connectAsync() = scope.async(start = LAZY) {
         _ready.value = false
 
         val connection = establishConnection().also { _connection = it }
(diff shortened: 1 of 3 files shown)


core/src/androidMain/kotlin/Peripheral.kt
```diff
@@ -111,7 +112,7 @@ public class AndroidPeripheral internal constructor(
 
     public override suspend fun connect() {
         check(job.isNotCancelled) { "Cannot connect, scope is cancelled for $this" }
-        connectJob.updateAndGet { it ?: createConnectJob() }!!.join()
+        connectJob.updateAndGet { it ?: connectAsync() }!!.await()
```
QuantumRand, line 115: What happens if `await()` is called when the `Deferred` is already awaiting?

twyatt: Multiple callers will all suspend until there is a result. Similar behavior as `join`, it's just expecting a result now. We aren't actually providing a result via `async`, but wanted the behavior of holding onto any failures for us until we've had a chance to call `await`.

QuantumRand: Does it matter that the `connectJob` is never cleared? Wouldn't disconnecting and then reconnecting just return the result of the first `connectAsync()` without actually doing anything?

twyatt: It's a bit of a dance, so admittedly hard to follow. The dance was necessary as multiple code paths (e.g. `disconnect` invoked, or remote peripheral dropped connection, etc) needed to cause the clearing of the `connectJob`.

On Android, handled via the `Connection`'s `invokeOnClose`:

https://github.com/JuulLabs/kable/blob/fe64627bd19e43f602d7e574c0f148e2977c24df/core/src/androidMain/kotlin/Peripheral.kt#L78-L83

macOS: https://github.com/JuulLabs/kable/blob/fe64627bd19e43f602d7e574c0f148e2977c24df/core/src/macosX64Main/kotlin/Peripheral.kt#L113-L118

JS: https://github.com/JuulLabs/kable/blob/fe64627bd19e43f602d7e574c0f148e2977c24df/core/src/jsMain/kotlin/Peripheral.kt#L231-L237

QuantumRand: Ah, got it. I was looking for it in the disconnect-related calls. 👍

QuantumRand (approved):

davidtaylor-juul (approved):

twyatt:
Note to self: double check that multiple callers will all suspend when calling `async`'s `await` (https://github.com/JuulLabs/kable/pull/46#discussion_r549821905) — made comment from memory, just want to make sure it's accurate before merging.

twyatt:
Double checked `await` behavior behaves as desired: https://gist.github.com/twyatt/cb05bef5756082b31d71f69e8afdc427

PR #48, opened 2020-12-29
Prevent exceptions being thrown when offering to closed Channels

### LINKED ISSUE #47: AbortFlowException crash on repeated connect/disconnect calls
> @twyatt  yes I could not reproduce this crash but I have found another crash when does connect/disconnect a lot of times in raw.
> `kotlinx.coroutines.flow.internal.AbortFlowException: Flow was aborted, no more elements needed`
> Unfortunately that there are not other lines of the stack trace. 
> It happens just after call method connect.
> 
> `kotlinx.coroutines.flow.internal.AbortFlowException: Flow was aborted, no more elements needed
> 2020-12-29 12:39:45.027 6461-6511/com.xx D/BluetoothGatt: connect() - device: 43:95:9A:31:C7:6C, auto: false
> 2020-12-29 12:39:45.028 6461-6511/com.xx D/BluetoothAdapter: isSecureModeEnabled
> 2020-12-29 12:39:45.029 6461-6511/com.xx D/BluetoothGatt: registerApp()
> 2020-12-29 12:39:45.030 6461-6511/com.xx D/BluetoothGatt: registerApp() - UUID=1228b816-0f28-48f8-a49d-ef4c752b9027
> 2020-12-29 12:39:45.034 6461-6476/com.xx D/BluetoothGatt: onClientRegistered() - status=0 clientIf=10`

_Originally posted by @kostya29-strikersoft in https://github.com/JuulLabs/kable/issues/44#issuecomment-752030738_

twyatt:
Suppresses possible exceptions thrown when attempting to send to closed `Channel`s.

Fixes #47

### DIFF
--- a/core/src/commonMain/kotlin/Channels.kt
+++ b/core/src/commonMain/kotlin/Channels.kt
@@ -1,0 +1,8 @@
+package com.juul.kable
+
+import kotlinx.coroutines.channels.SendChannel
+
+// Workaround until https://github.com/Kotlin/kotlinx.coroutines/issues/974 is closed.
+internal fun <E> SendChannel<E>.offerCatching(element: E): Boolean {
+    return runCatching { offer(element) }.getOrDefault(false)
+}
--- a/core/src/androidMain/kotlin/gatt/Callback.kt
+++ b/core/src/androidMain/kotlin/gatt/Callback.kt
@@ -27,6 +27,7 @@ import com.juul.kable.gatt.Response.OnDescriptorWrite
 import com.juul.kable.gatt.Response.OnMtuChanged
 import com.juul.kable.gatt.Response.OnReadRemoteRssi
 import com.juul.kable.gatt.Response.OnServicesDiscovered
+import com.juul.kable.offerCatching
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.channels.Channel.Factory.CONFLATED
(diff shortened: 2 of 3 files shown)


core/src/commonMain/kotlin/Channels.kt
```diff
@@ -0,0 +1,8 @@
+package com.juul.kable
+
+import kotlinx.coroutines.channels.SendChannel
+
+// Workaround until https://github.com/Kotlin/kotlinx.coroutines/issues/974 is closed.
```
Phoenix7351, line 5: Should we mark this with TODO to make sure we know to check up on coroutines' issue?

twyatt: In this particular case, I don't feel `TODO` is necessary, since the workaround will exhibit desired behavior (i.e. there isn't a bug remaining after this change — there will just be a way to do it in the future that requires less code, after issue 974 is closed) and it would be fine if we never changed it.

**tl;dr** I don't think there's a need for an explicit marker of work to be done.

Phoenix7351 (approved):

sdonn3 (approved):

QuantumRand (approved):

davidtaylor-juul (approved):

twyatt:
Before merging, going to investigate https://github.com/JuulLabs/kable/issues/47#issuecomment-752360748.

twyatt:
Superseded by #109.

PR #49, opened 2020-12-31
Add ability to configure writing to config descriptor for observations

### LINKED ISSUE #42: Add notification setup mode for devices that do not contain client characteristic config descriptor
There are some BLE devices which does not have client characteristic config descriptor. It's a violation of BLE specification and the bug of the firmware of peripheral, but there is no possibility to change it for mobile developers. For a workaround, it's good to have the possibility to subscribe on notification by only setCharacteristicNotification(characteristic, boolean) call without writing to the descriptor.
Library RXAndroidBle has this possibility (NotificationSetupMode.Compact) and earlier I have used it but unfortunately, it's not friendly for kotlin implementation.

### LINKED ISSUE #43: Android - observer.collect throws NoSuchElementException
I'm trying to read some data of a UART Service that provides a "write without response" and a "notify" characteristic. I can write to it fine using cable but Kable but get the following error when using `observer.collect { ... }` to subscribe to notifications. 

`java.util.NoSuchElementException: Collection contains no element matching the predicate.`

The characteristic works fine to read using other apps and sends out data about once a second. As probably will be apparent from my sample code below I'm fairly new to Kotlin but as far as I understand after reading up and digging through the source code I can't see why this error is produced. 

I looked at issue [38](https://github.com/JuulLabs/kable/issues/38) but since I go through the steps to get the characteristic I get this error even if it is discovered. Or did i misunderstand the cause of that issue?

The characteristic and observer objects looks fine to me, what I can notice is that the 

Sample code
```kotlin       
bluetoothScope.launch {
            val peripheral: Peripheral? = getPeripheral() // Get Peripheral using Scanner()
            if (peripheral != null) {
                peripheral.connect()
                val characteristic = peripheral.services
                        ?.first { service -> service.serviceUuid == UUID.fromString(UART_SERVICE)}
                        ?.characteristics?.first { characteristic ->
                            characteristic.characteristicUuid == UUID.fromString(RX_CHARACTERISTIC_UUID)
                        }
                if (characteristic != null ) {
                    val observer = peripheral.observe(characteristic)
                                try {
                                    observer.collect { data ->
                                        Log.d(TAG, data)
                                    }
                                } catch (e: NoSuchElementException) {
                                    Log.d(TAG, "Error: $e")
                                }
                }
                peripheral.state?.collect { state ->
                    when (state) {
                        State.Connected -> {
                            Log.d(TAG, "Connected")
                        }
                        else -> Log.d(TAG, "Not Connected")
                    }
                }
            }

        }
    }
```

Scan result
```
DiscoveredService(serviceUuid=6e400001-b5a3-f393-e0a9-e50e24dcca9e, 
characteristics=[
    DiscoveredCharacteristic(
        serviceUuid=6e400001-b5a3-f393-e0a9-e50e24dcca9e,
        characteristicUuid=6e400003-b5a3-f393-e0a9-e50e24dcca9e, 
        descriptors=[]), 
    DiscoveredCharacteristic(
        serviceUuid=6e400001-b5a3-f393-e0a9-e50e24dcca9e, 
        characteristicUuid=6e400002-b5a3-f393-e0a9-e50e24dcca9e, 
        descriptors=[])]
)]
```

twyatt:
Adds ability to configure if the config descriptor for a characteristic is written to when starting/stopping an observation.

> ~I do not have a peripheral that supports notifications on a characteristic that is missing the config descriptor, so I am unable to testes this, nor can I confirm behavior for Apple and JavaScript platforms (which do not have this as configurable).~

**UPDATE:** Tested via Android test app (see https://github.com/JuulLabs/kable/issues/42#issuecomment-757428891 for details).

> ~I suspect that Apple and JavaScript only write to config descriptor if it exists, otherwise many peripherals that violate this requirement (presence of config descriptor) would not work with characteristic notifications.~

**UPDATE:** This assumption was found to be incorrect.

> ~In trying to make behavior consistent across platforms, I'm running with the assumption that Apple and JavaScript both fail gracefully, therefor making the default on Android as `Auto` (which only writes to config descriptor if present — i.e. its presence is not required).~

**UPDATE:** It was found that Apple and JavaScript exhibit behavior that matches `Always`, PR has been updated to use `Always` as default.

Closes #42
Closes #43

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -153,8 +153,10 @@ public final class com/juul/kable/Peripheral$DefaultImpls {
 }
 
 public final class com/juul/kable/PeripheralKt {
-	public static final fun peripheral (Lkotlinx/coroutines/CoroutineScope;Landroid/bluetooth/BluetoothDevice;)Lcom/juul/kable/Peripheral;
+	public static final fun peripheral (Lkotlinx/coroutines/CoroutineScope;Landroid/bluetooth/BluetoothDevice;Lcom/juul/kable/WriteNotificationDescriptor;)Lcom/juul/kable/Peripheral;
 	public static final fun peripheral (Lkotlinx/coroutines/CoroutineScope;Lcom/juul/kable/Advertisement;)Lcom/juul/kable/Peripheral;
+	public static final fun peripheral (Lkotlinx/coroutines/CoroutineScope;Lcom/juul/kable/Advertisement;Lcom/juul/kable/WriteNotificationDescriptor;)Lcom/juul/kable/Peripheral;
+	public static synthetic fun peripheral$default (Lkotlinx/coroutines/CoroutineScope;Landroid/bluetooth/BluetoothDevice;Lcom/juul/kable/WriteNotificationDescriptor;ILjava/lang/Object;)Lcom/juul/kable/Peripheral;
 }
 
 public final class com/juul/kable/ScanFailedException : java/lang/IllegalStateException {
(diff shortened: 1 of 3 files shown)


Phoenix7351 (approved):

sdonn3 (approved):

QuantumRand (approved):

twyatt:
Going to wait to merge this PR.

Ordered a NRF52840-DONGLE so that I can sniff the BLE packets to make sure the various platforms are behaving consistently.

davidtaylor-juul (approved):

Issue #70, opened 2021-02-27
Using Kable in an Android Project


Computr0n:
I have an android gradle module and I am following the platform specific instructions in the readme that say to add:

```groovy
dependencies {
    implementation("com.juul.kable:core-android:0.2.0")
}
```

Gradle successfully syncs, but I don't see the library show up in the project window under external libraries and I cannot seem to use any kable classes in my project. 

I'm not sure whats missing, I have never used a multiplatform library from an android project. 

If I set the version to 0.1.0, android studio will recommend the next available highest version. (0.2.0 in my case although I see 0.3.0 was published yesterday). Because of this, I don't think its gradle failing to sync silently. not sure where to look next.

twyatt:
From what you've described, it _should_ work.

I find that sometimes Android Studio doesn't give the clearest failure as to what went wrong w/ Gradle sync, or it silently fails.

Can you try from the command line and let me know what kind of failure you're getting when you try to use:

```groovy
dependencies {
    implementation("com.juul.kable:core-android:0.3.0")
}
```

Example Gradle command line invocation:

```
./gradlew clean assemble
```

twyatt:
Looking at Maven, it looks like the `-metadata` artifact failed to publish w/ `0.3.0`. Although multiplatform seems to still work (tested by bumping dependency to `0.3.0` in [SensorTag sample app](https://github.com/JuulLabs/sensortag) and it compiled). I don't know a lot about the Android Studio dependency version suggestions, but it's possible that the missing metadata artifact for `0.3.0` is why it isn't suggestion the version bump to that version. 🤷 

In either case, I believe `0.3.0` should still work via the process you've described, so do let me know what the result of the Gradle command line invocation is. If we still can't figure it out, then I'll investigate the publication issue and/or publish another version.

Computr0n:
OK, I tried what you said on a freshly reset stable copy of the latest stable build of
```
Android Studio 4.1.2
Build #AI-201.8743.12.41.7042882, built on December 19, 2020
Runtime version: 1.8.0_242-release-1644-b3-6915495 x86_64
```

using a brand new project with only this line added to the app gradle modules dependency block:
```
    implementation("com.juul.kable:core-android:0.X.0")
```

Here is the output after running 
```shell
./gradlew clean assemble
```

with version 0.2.0:
```
~/AndroidStudioProjects/MyApplication  ./gradlew clean assemble

FAILURE: Build failed with an exception.

* What went wrong:
[...]

PR #72, opened 2021-02-28
Simplify platform specific instructions

Computr0n:
https://github.com/JuulLabs/kable/issues/70

PR #73, opened 2021-02-28
Update Gradle example for platform specific configuration

twyatt:
Per https://github.com/JuulLabs/kable/issues/70#issuecomment-787509116, the undecorated Maven artifact should be used even for platform specific configurations.

Rendered markdown [here](https://github.com/JuulLabs/kable/tree/twyatt/readme#platform-specific).


twyatt:
Closing in favor of simpler change in #72.

PR #74, opened 2021-03-04
Configure observation based on characteristic properties

### LINKED ISSUE #54: Add support for enable indication
There are devices which requests enabling indication (BluetoothGattDescriptor.ENABLE_INDICATION_VALUE) instead of notification (BluetoothGattDescriptor.ENABLE_NOTIFICATION_VALUE) in order to receive (observe in kable) values from them. 
Maybe kable-api should allow such thing, as I do not see any kind of workaround at the moment.
      
This could be checked against characteristic property like so:
characteristic.getProperties() and PROPERTY_NOTIFY / PROPERTY_INDICATION   > 0

twyatt:
# 📜  Background

When Android receives and **indication** it automatically (transparent to us) responds with an ACK. It is up to the library to inform peripheral which form of observation is desired (via Client Characteristic Configuration descriptor; a.k.a. CCCD).

## Indication

```
.--------.             .--------.
| Client |             | Server |
'--------'             '--------'
     |                      |
     |      Indication      |
     | <------------------- |
     |                      |
     |         Ack          |
     | -------------------> |
     |                      |
```

## Notification

```
.--------.             .--------.
| Client |             | Server |
'--------'             '--------'
     |                      |
     |     Notification     |
     | <------------------- |
     |                      |
```

# 🚧  Changes

Prior to this PR, `observe` always blindly enabled **notification**. With this PR, the characteristic properties are examined to determine if `observe` should be **notification** based and/or **indication** based.

If Client Characteristic Configuration descriptor (CCCD) is absent then neither indicate nor notify bit is written and it is assumed that characteristic emits unconditionally (this matches the [behavior of JavaScript](https://webbluetoothcg.github.io/web-bluetooth/#dom-bluetoothremotegattcharacteristic-startnotifications)).

Closes #54


core/src/androidMain/kotlin/Peripheral.kt
```diff
@@ -33,29 +36,42 @@ import kotlinx.coroutines.flow.asStateFlow
...
 
 public actual fun CoroutineScope.peripheral(
     advertisement: Advertisement,
 ): Peripheral = peripheral(advertisement.bluetoothDevice)
 
+public fun CoroutineScope.peripheral(
+    bluetoothDevice: BluetoothDevice,
+): Peripheral = AndroidPeripheral(coroutineContext, bluetoothDevice)
+
+@Deprecated(
```
sdonn3, line 52: Are these already being used somewhere downstream that they need to be deprecated instead of just removed?

sdonn3 (approved):

core/src/androidMain/kotlin/Peripheral.kt
twyatt: They were available in `0.3.0` release, so just giving users of that version a notice/path to the new usage.

PR #98, opened 2021-05-14
Prevent concurrent I/O operations on JavaScript

### LINKED ISSUE #97: Prevent concurrent I/O operations on JavaScript
Internally our team as begun to see `DOMException: GATT operation already in progress.` errors.

I believe for the current JavaScript implementation I made the incorrect assumption that Web Bluetooth would handle the "queueing" of multiple I/O operations. This appears to be the wrong assumption and similar to Android and Apple platforms, we need to put all I/O operations within a lock (to prevent concurrent I/O requests).

Current locking implementations in Android and Apple that prevent concurrent I/O operations:

## Android

https://github.com/JuulLabs/kable/blob/151e54d255bf5595c67023927084d083e6180706/core/src/androidMain/kotlin/Connection.kt#L58

## Apple

https://github.com/JuulLabs/kable/blob/151e54d255bf5595c67023927084d083e6180706/core/src/appleMain/kotlin/Connection.kt#L15

twyatt:
Closes #97

### DIFF
--- a/core/src/jsMain/kotlin/Observers.kt
+++ b/core/src/jsMain/kotlin/Observers.kt
@@ -73,14 +73,14 @@ internal class Observers(
         if (--count < 1) {
             bluetoothRemoteGATTCharacteristic.apply {
                 /* Throws `DOMException` if connection is closed:
-                     *
-                     * DOMException: Failed to execute 'stopNotifications' on 'BluetoothRemoteGATTCharacteristic':
-                     * Characteristic with UUID [...] is no longer valid. Remember to retrieve the characteristic
-                     * again after reconnecting.
-                     *
-                     * Wrapped in `runCatching` to silently ignore failure, as notification will already be
-                     * invalidated due to the connection being closed.
-                     */
+                 *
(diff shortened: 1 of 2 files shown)


QuantumRand (approved):
Is this issue unit testable? It sounds like the intetntion is to update this code once more when `KT-37279` is fixed.

sdonn3 (approved):

twyatt:
> Is this issue unit testable?

Yes, but not easily. Would need to mock a lot of the underlying bluetooth objects, of which the infrastructure for testing that on the JavaScript platform has not been setup. I hope to one day make the project more unit test friendly, but I think that is still a ways out. 😢 

> It sounds like the intetntion is to update this code once more when `KT-37279` is fixed.

The intention is the update the code once we move to IR compiler (and drop support for the LEGACY compiler). I suspect KT-37279 may never be fixed since it only effects the LEGACY compiler.

davidtaylor-juul (approved):

cedrickcooke (approved):

PR #107, opened 2021-06-06
Bump CI dependencies and use in-memory ascii-armored key for signing

twyatt:
Using built in GPG mechanisms ([supported by Gradle](https://docs.gradle.org/current/userguide/signing_plugin.html#using_in_memory_ascii_armored_openpgp_subkeys)) for storing keys as plain text (as GitHub secret), rather than the previously used method of base64'ing the binary key ring.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,42 +2,46 @@ name: CI
 on:
   push:
 
+env:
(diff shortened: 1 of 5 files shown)


sdonn3 (approved):

twyatt:
Pulling in additional reviewers as @QuantumRand is busy at the moment.

cedrickcooke (approved):

PR #108, opened 2021-06-06
Support constructing `Peripheral`s using builder lambda

twyatt:
Usage details can be found in the [README Peripherals section](https://github.com/JuulLabs/kable/tree/twyatt/builder#peripheral).

Worth mentioning (with more detail beyond the `README`) is the `onServicesDiscovered` option (in the builder lambda):
It allows for actions to be performed "after service discovery".

Without this feature, performing an action upon connection required wrapping Kable's `Peripheral` so that operations could be performed after the `connect` (additionally, ensuring those operations were idempotent and thread-safe was difficult and error prone).

Kable already incorporates the necessary locks/safe-guards so this makes it possible for consumers to leverage that functionality.

The `DiscoveredServicesPeripheral` was created to provide a reduced functionality `Peripheral` which is available in the `onServiceDiscovered` lambda. It excludes the ability `connect` or `disconnect`, whereas `connect` from the lambda would create a deadlock. So, this limited functionality `Peripheral` prevents a footgun situation.


davidtaylor-juul (approved):
🚀 very nice

sdonn3 (approved):
Love it

QuantumRand (approved):

PR #109, opened 2021-06-07
Use `trySend` instead of deprecated `offer`

twyatt:


sdonn3 (approved):

core/src/androidMain/kotlin/gatt/Callback.kt
```diff
@@ -174,3 +181,9 @@ private fun Int.toStatus(): State.Disconnected.Status? = when (this) {
     GATT_CONN_CANCEL -> Cancelled
     else -> Unknown(this)
 }
+
+private fun <E> SendChannel<E>.trySendOrLog(element: E) {
+    trySend(element).getOrElse {
+        Log.w(TAG, "Callback was unable to deliver $element")
```
davidtaylor-juul, line 187: Is there any use in also printing the exception here?

core/src/jsMain/kotlin/Scanner.kt
```diff
@@ -31,9 +32,7 @@ public class JsScanner internal constructor(
 
         val scan = bluetooth.requestLEScan(options.toDynamic()).await()
         val listener: (Event) -> Unit = {
-            runCatching {
-                offer(Advertisement(it as BluetoothAdvertisingEvent))
-            }.onFailure {
+            trySend(Advertisement(it as BluetoothAdvertisingEvent)).getOrElse {
                 console.warn("Unable to deliver advertisement event due to failure in flow or premature closing.")
```
davidtaylor-juul, line 36: Ditto, do we want to print the exception too?

davidtaylor-juul (approved):
LGTM

QuantumRand (approved):

core/src/androidMain/kotlin/gatt/Callback.kt
twyatt: Ya, you're probably right. I was thinking this condition shouldn't occur. I think it would only occur if the OS delivered us a response _after_ a disconnected state, which would be very bizarre for the OS to do.

Being that we don't expect it, if it is seen, it would be good a stacktrace to see where the message is coming from along with the message.

core/src/jsMain/kotlin/Scanner.kt
twyatt: What is the standard way to log an exception in JavaScript?

davidtaylor-juul: For production it's best not to log at all and rely on `throw`, but this is probably fine.

PR #110, opened 2021-06-08
Propagate early characteristic changes

### LINKED ISSUE #105: Early characteristic changes can be missed
As was seen in https://github.com/JuulLabs/kable/issues/99#issuecomment-842941308, an unprovoked (prior to writing to the config descriptor to enable Notify/Indicate) characteristic may trigger a "characteristic change" event.

Code used in test branch that should be brought in to main code:

<details>
<summary><code>Observers.kt</code></summary>

```kotlin
    fun acquire(characteristic: Characteristic) = flow {
        try {
            characteristicChanges
                .onSubscription {
                    peripheral.suspendUntilReady()

                    if (observers.incrementAndGet(characteristic) == 1) {
                        peripheral.startObservation(characteristic)
                    }
                }
                .collect {
                    if (it.characteristic.characteristicUuid == characteristic.characteristicUuid &&
                        it.characteristic.serviceUuid == characteristic.serviceUuid
                    ) emit(it.data)
                }
        } finally {
            if (observers.decrementAndGet(characteristic) < 1) {
                try {
                    peripheral.stopObservation(characteristic)
                } catch (e: NotReadyException) {
                    // Silently ignore as it is assumed that failure is due to connection drop, in which case Android
                    // will clear the notifications.
                    Log.d(TAG, "Stop notification failure ignored.")
                }
            }
        }
    }
```
</details>

### LINKED ISSUE #112: peripheral.observe(readCharacteristic) doesn't provide a form to know when is safe to write
Hello there!

I'm trying to migrate a project to kable. The ble device (and others I worked with...) uses some opcode communication where you write to a writeCharacteristic and all the data is received on the readCharacteristic subscribed to notify.
After notify is subscribed I can enqueue calls to get a first response as soon as posible.

Enter kable, the peripheral.observe(readCharacteristic) provides a flow with the data, but cant signal when it is actually able to receive data. So i can't just enqueue a write to the descriptor as doing it "on the next line" because the first response will be missed as the observe isn't really ready. I can workaround that doing a delay(2000) for example between the .observe and the write.

Maybe i've missed something but can't seem to find a method or param that handles that.
Could it be possible to set the notify/observe on the new (0.6) connect params? or maybe provide a param to the observe to set an initial write?

Keep the good work, awesome library!

twyatt:
Closes #105
Closes #112

Also fixes JavaScript observations to wait until connection is ready. Only users of experimental advertisement support in Chrome would've seen this bug, as using `requestDevice` obtains devices in a connected state, i.e. already ready.

### DIFF
--- a/core/src/androidMain/kotlin/Observers.kt
+++ b/core/src/androidMain/kotlin/Observers.kt
@@ -41,12 +41,14 @@ internal class Observers(
     private val lock = Mutex()
 
     fun acquire(
-        characteristic: Characteristic
+        characteristic: Characteristic,
+        onObservationStarted: ObservationStartedAction,
     ) = characteristicChanges
         .onSubscription {
             peripheral.suspendUntilReady()
             if (observers.incrementAndGet(characteristic) == 1) {
                 peripheral.startObservation(characteristic)
+                onObservationStarted()
             }
         }
         .filter {
--- a/core/src/androidMain/kotlin/Observers.kt
+++ b/core/src/androidMain/kotlin/Observers.kt
@@ -3,6 +3,7 @@ package com.juul.kable
 import android.util.Log
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.catch
 import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.onCompletion
@@ -77,24 +78,24 @@ internal class Observers(
 private class Observations {
 
     private val lock = Mutex()
-    private val collection = HashMap<Characteristic, MutableList<OnSubscriptionAction>>()
+    private val observations = mutableMapOf<Characteristic, MutableList<OnSubscriptionAction>>()
 
     suspend inline fun forEach(
         action: (Characteristic, List<OnSubscriptionAction>) -> Unit
     ) = lock.withLock {
-        collection.forEach { (characteristic, onSubscriptionActions) ->
+        observations.forEach { (characteristic, onSubscriptionActions) ->
             action(characteristic, onSubscriptionActions)
         }
     }
 
     suspend fun add(
         characteristic: Characteristic,
-        onSubscription: OnSubscriptionAction
+        onSubscription: OnSubscriptionAction,
     ): Int = lock.withLock {
-        val actions = collection[characteristic]
+        val actions = observations[characteristic]
         if (actions == null) {
             val newActions = mutableListOf(onSubscription)
-            collection[characteristic] = newActions
+            observations[characteristic] = newActions
             1
         } else {
             actions += onSubscription
@@ -104,14 +105,19 @@ private class Observations {
 
     suspend fun remove(
         characteristic: Characteristic,
-        onSubscription: OnSubscriptionAction
+        onSubscription: OnSubscriptionAction,
     ): Int = lock.withLock {
-        val actions = collection[characteristic]
-        if (actions == null) {
-            0
-        } else {
-            actions -= onSubscription
-            actions.count()
+        val actions = observations[characteristic]
+        when {
+            actions == null -> 0
+            actions.count() == 1 -> {
+                observations -= characteristic
+                0
+            }
+            else -> {
+                actions -= onSubscription
+                actions.count()
+            }
         }
     }
 }
--- a/core/src/androidMain/kotlin/Observers.kt
+++ b/core/src/androidMain/kotlin/Observers.kt
@@ -3,8 +3,10 @@ package com.juul.kable
 import android.util.Log
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableSharedFlow
-import kotlinx.coroutines.flow.collect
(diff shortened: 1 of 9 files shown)


core/src/androidMain/kotlin/Observers.kt
```diff
@@ -38,20 +40,21 @@ internal class Observers(
...
-            peripheral.startObservation(characteristic)
-        }
-
-        try {
-            characteristicChanges.collect {
-                if (it.characteristic.characteristicUuid == characteristic.characteristicUuid &&
-                    it.characteristic.serviceUuid == characteristic.serviceUuid
-                ) emit(it.data)
+    fun acquire(
+        characteristic: Characteristic
```
sdonn3, line 44: nit: seems like a single parameter could just go on one line?

sdonn3 (approved):
looks good!

cedrickcooke (approved):

twyatt:
@sdonn3 @cedrickcooke re-requested review as I've added a additional changes since last reviewed:

- refactored JavaScript observation handling to resemble the other platforms
- implemented `observe`'s `onSubscription` lambda to be invoked on re-connection
    - now supports invoking all `onSubscription` lambdas when multiple `observe` Flows are created for the same `characteristic`

core/src/commonMain/kotlin/Peripheral.kt
```diff
@@ -114,8 +115,27 @@ public interface Peripheral {
...
+     * The [onSubscription] action is useful in situations where an initial operation is needed when starting an
+     * observation (such as writing a configuration to the peripheral and expecting the response to come back in the
+     * form of a characteristic change). The [onSubscription] is invoked for every new subscriber; if it is desirable to
+     * only invoke the [onSubscription] once per connection (for the specified [characteristic]) then you can either
+     * use the [shareIn][kotlinx.coroutines.flow.shareIn] [Flow] operator on the returned [Flow], or call [observe]
+     * again with the same [characteristic] and without specifying an [onSubscription] action.
+     *
+     * If multiple [observations][observe] are created for the same [characteristic] but with different [onSubscription]
+     * actions, then the [onSubscription] actions will be executed in the order in which the returned [Flow]s are
+     * subscribed to.
```
cedrickcooke, line 135: ❤️

core/src/jsMain/kotlin/Observers.kt
```diff
@@ -1,129 +1,89 @@
...
+            peripheral.startObservation(characteristic)
+            onSubscriptionActions.forEach { it() }
         }
     }
+}
 
-    fun invalidate() {
-        observers.forEach { (_, observation) ->
-            observation.listener = null
+private class Observations {
```
cedrickcooke, line 52: It's a little sad that this `Observations` class can't live in common code. I can see that they're different but it feels _super_ similar between the three implementations.

cedrickcooke (approved):

core/src/jsMain/kotlin/Observers.kt
twyatt: I agree it's a little sad. Maybe one day I can figure out a way to make it generic and reusable, but for now, just getting it working. 🤷

twyatt:
Going to do some internal testing against the JavaScript target, then merge.

core/src/androidMain/kotlin/Observers.kt
```diff
@@ -34,25 +36,26 @@ internal class Observers(
...
-                if (it.characteristic.characteristicUuid == characteristic.characteristicUuid &&
-                    it.characteristic.serviceUuid == characteristic.serviceUuid
-                ) emit(it.data)
+    fun acquire(
+        characteristic: Characteristic,
+        onSubscription: OnSubscriptionAction,
+    ) = characteristicChanges
+        .onSubscription {
+            peripheral.suspendUntilReady()
+            if (observations.add(characteristic, onSubscription) == 1) {
```
sdonn3, line 47: Should this be '>= 1' to mirror the '< 1' below?

sdonn3 (approved):
Totally appreciate how easy this is to read through

twyatt:
lol, @sdonn3 either I'm going crazy, or you left a comment then deleted it.

Here was the response I was going to do, though I'm guessing you don't need it now. 🤷 

----

Technically, the `< 1` below could be a `0` (but `< 1` was a little bit of a safety buffer), whereas this can only be a `== 1`.

I'll explain why:

In `onSubscription` we "register", aka "add", a subscription to the `Map` of characteristics to subscriptions. The return is the number of subscribers we have _after_ "registering" (or aka "adding"). The underlying system only needs to setup BLE notifications/indications once per characteristic (and not more than once). So, we essentially only want to enable them when we go from `0` to `1` subscribers and we'll just leave the notifications/indications enabled as more subscribers jump on.

On the reverse side (in `onCompletion`) we need to keep track as `Flow` collectors unsubscribe, all the way down until we have no more subscribers, and can therefore stop the BLE notifications/indications for that characteristic. Since we're subtracting within the `observations.remove` I thought it were safer if I messed up the math (and we somehow ended up with `-1`) that checking for `< 1` would be good. But, assuming the maths checks out 😉  then `== 0` would've sufficed. Essentially it would've represented when we are transitioning from `1` to `0` subscribers.

PR #113, opened 2021-06-13
Add `iosArm32` support

Jahor:
Add an extra target platform

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -16,7 +16,7 @@ kotlin {
         publishAllLibraryVariants()
     }
     iosX64()
-	iosArm32()
+    iosArm32()
     iosArm64()
     macosX64()
 
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -16,6 +16,7 @@ kotlin {
         publishAllLibraryVariants()
     }
     iosX64()
+	iosArm32()
     iosArm64()
     macosX64()
 
@@ -52,6 +53,13 @@ kotlin {
             dependencies {
                 api(coroutines("core", version = "1.5.0-native-mt!!"))
                 implementation(stately("isolate-iosx64"))
+            }
+        }
+
+        val iosArm32Main by getting {
+            dependencies {
+                api(coroutines("core", version = "1.5.0-native-mt!!"))
+                implementation(stately("isolate-iosarm32"))
             }
         }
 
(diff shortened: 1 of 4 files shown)


twyatt:
Thanks so much for the contribution! I need to check with our legal team, I think we'll need a CLA (or similar) in place to accept external contributions. Shouldn't be a problem (apologies that it might slow this first external contribution a bit). I'll follow up soon.

twyatt:
Heard back from legal, we don't need a CLA. 🎉 
I'll look over this PR today or tomorrow.
Thanks again for the contribution and patience!

core/src/appleMain/kotlin/NSData.kt
```diff
@@ -19,6 +21,6 @@ internal fun NSData.toByteArray(): ByteArray = ByteArray(length.toInt()).apply {
 internal fun ByteArray.toNSData(): NSData = memScoped {
     NSData.create(
         bytes = allocArrayOf(this@toNSData),
-        length = size.toULong()
+        length = size.convert<NSUInteger>()
```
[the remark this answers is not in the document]
davidtaylor-juul: Line 14 is Kotlin stdlib API not cinterop though.

davidtaylor-juul (approved):

core/build.gradle.kts
```diff
@@ -47,6 +48,13 @@ kotlin {
             }
         }
 
+        val iosArm32Main by getting {
+            dependencies {
+                api(coroutines("core", version = "1.5.0-native-mt!!"))
```
twyatt, line 53: Per #129, Kable no longer depends on `-native-mt` Coroutines and expects library consumers to properly pull in the `-native-mt` Coroutines for targets that need it.

```suggestion
```

core/src/appleMain/kotlin/NSData.kt
```diff
@@ -2,10 +2,12 @@ package com.juul.kable
 
 import kotlinx.cinterop.addressOf
 import kotlinx.cinterop.allocArrayOf
+import kotlinx.cinterop.convert
 import kotlinx.cinterop.memScoped
 import kotlinx.cinterop.usePinned
 import platform.Foundation.NSData
 import platform.Foundation.create
+import platform.darwin.NSUInteger
```
twyatt, line 10:
```suggestion
```

twyatt:
Thanks so much for the contribution @Jahor!

twyatt:
Was released in 0.7.1.

PR #125, opened 2021-06-24
Support requesting connection priority on Android

### LINKED ISSUE #36: Provide ability to request connection priority
[Android BLE API supports requesting various connection priorities](https://developer.android.com/reference/android/bluetooth/BluetoothGatt#requestConnectionPriority(int)). Kable should expose this functionality for platforms that support it.

Implemented in Able via https://github.com/JuulLabs/able/pull/95.

twyatt:
Closes #36

Unlike nearly every other BLE request on Android, this one does **not** (or at least it isn't documented to) invoke a function on [`BluetoothGattCallback`](https://developer.android.com/reference/android/bluetooth/BluetoothGattCallback) to indicate when the process is complete (and whether successful or not). It seems you just have to blindly hope it worked. 🤷 

Per [Android ble, requestConnectionPriority not working](https://stackoverflow.com/a/47491885), it is recommended that `requestConnectionPriority` not be called during service discovery. Since Kable's `connect` suspends until connected _and_ services are discovered, it should be safe to call `requestConnectionPriority` anytime after `connect` and not worry about the issue described in the StackOverflow post.


cedrickcooke:
I went to check the docs to see what the boolean this was returning meant, and, um... wow.

Screen Shot 2021-06-24 at 9 25 57 AM

cedrickcooke (approved):

davidtaylor-juul (approved):

Issue #93, opened 2021-04-22
Services with the same UUIDs


Shakenbeer:
Thanks a lot for you work. The library is awesome.

I have BLE device with this strange firmware. There are two services with the same UUIDs and with identical sets of characteristics. The only difference is in one of descriptors. So, there are two services S1 and S2, they both have characteristic C, and value of a descriptor D differs. All uuids are the same.

S1.uuid == S2.uuid
S1.C.uuid == S2.C.uuid
S1.C.D.uuid == S2.C.D.uuid
S1.C.D.value != S1.C.D.value.

This way I should distinct those 2 services. 

So after services are discovered I do a loop over them and for every service S1 and S2 I read that descriptor. 
Unfortunately I always get the same value, as in both cases I read only descriptor of the service S1. 
`readDescriptor` function searches for descriptor through all services and always return the first one found.
Is there a way to read identical descriptors (and characteristics) of two different services?

twyatt:
Android provides [`instanceId`](https://developer.android.com/reference/android/bluetooth/BluetoothGattCharacteristic#getInstanceId()) for characteristics. I think that should allow multiple characteristics with the same UUID to be identified as distinct.

I'll have to investigate of Core Bluetooth (e.g. iOS) and JavaScript offer a similar property (to keep behavior parity between the platforms).

twyatt:
@Shakenbeer it looks like [BluetoothGattService](https://developer.android.com/reference/android/bluetooth/BluetoothGattService) and [BluetoothGattCharacteristic](https://developer.android.com/reference/android/bluetooth/BluetoothGattCharacteristic) both have an `instanceId`, which allow some form of API in Kable to allow you to interact with distinct items with the same UUID, but I don't see a similar property for [BluetoothGattDescriptor](https://developer.android.com/reference/android/bluetooth/BluetoothGattDescriptor).

You don't have multiple descriptors (with the same UUID) on the same characteristic, do you?

Shakenbeer:
@twyatt There is `mInstance` in BluetoothGattDescriptor, but the method `getInstanceId` is hidden:
```
     * @hide
     */
    public int getInstanceId() {
        return mInstance;
    }
```
I did a quick fix for myself. Now I can read/write/observe characteristics and read/write descriptors for services with the same UUID. `Peripheral.kt`:
```
public suspend fun read(
        descriptor: Descriptor,
        serviceInstance: Int = 0
    ): ByteArray
```
[...]

PR #128, opened 2021-06-29
Finding characteristic by UUID + required property

solvek:
This allows to find an appropriate characteristic when there are duplicating two characteristics by UUID (but with different property).
So for writing we use the charactertistic with `WRITE` or `WRITE_NO_RESPONSE` property, for observing we use only property with `NOTIFY` or `INDICATE`.

BTW, what is the easiest way to test this locally for me? How can I use my modified library in my local project?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,13 +34,22 @@ jobs:
       - run: ./gradlew assemble
       - run: ./gradlew check
 
-      - if: runner.os == 'macOS'
-        run: >
-          ./gradlew
-          --no-parallel
-          -PsigningInMemoryKey="${{ secrets.SIGNING_KEY }}"
-          -PsigningInMemoryKeyPassword="${{ secrets.SIGNING_PASSWORD }}"
-          publishToMavenLocal
+      - run: |
+          set -o xtrace
+          if [ ! -z "${{ secrets.SIGNING_KEY }}" ]; then
+            ./gradlew \
+            --no-parallel \
+            -PVERSION_NAME="unspecified" \
+            -PsigningInMemoryKey="${{ secrets.SIGNING_KEY }}" \
+            -PsigningInMemoryKeyPassword="${{ secrets.SIGNING_PASSWORD }}" \
+            publishToMavenLocal
+          else
+            ./gradlew \
+            --no-parallel \
+            -PVERSION_NAME="unspecified-SNAPSHOT" \
+            publishToMavenLocal
+          fi
+        if: ${{ runner.os == 'macOS' && github.repository_owner == 'JuulLabs' }}
 
       - run: |
           rm -f ~/.gradle/caches/modules-2/modules-2.lock
--- a/core/src/androidMain/kotlin/BluetoothDevice.kt
+++ b/core/src/androidMain/kotlin/BluetoothDevice.kt
@@ -2,7 +2,12 @@ package com.juul.kable
 
 import android.annotation.TargetApi
 import android.bluetooth.BluetoothDevice
-import android.bluetooth.BluetoothDevice.*
+import android.bluetooth.BluetoothDevice.PHY_LE_1M_MASK
+import android.bluetooth.BluetoothDevice.PHY_LE_2M_MASK
+import android.bluetooth.BluetoothDevice.PHY_LE_CODED_MASK
+import android.bluetooth.BluetoothDevice.TRANSPORT_AUTO
+import android.bluetooth.BluetoothDevice.TRANSPORT_BREDR
+import android.bluetooth.BluetoothDevice.TRANSPORT_LE
 import android.content.Context
 import android.os.Build
 import android.os.Handler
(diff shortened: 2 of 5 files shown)


twyatt:
Thanks for the PR!
I'm pretty swamped at the moment, so I'll take a look at this PR when I have a free moment.

> BTW, what is the easiest way to test this locally for me? How can I use my modified library in my local project?

```
cd kable/
./gradlew publishToMavenLocal
```

It will default to installing as version `main-SNAPSHOT`. To specify an alternate version, use the `VERSION_NAME` property:

```
./gradlew publishToMavenLocal -PVERSION_NAME=0.0.1-example-SNAPSHOT
```

Versions should be suffixed with `-SNAPSHOT` to disable publication signing.

Then, in the app that is using Kable, add `mavenLocal()` to list of repositories, for example:

```kotlin
allprojects {
    repositories {
        mavenLocal()
        google()
        mavenCentral()
    }
}
```

https://github.com/JuulLabs/sensortag/blob/main/build.gradle.kts#L19-L21

And where you use the library, to use your local version, for example:

```kotlin
kotlin {
    android()

    sourceSets {
        val commonMain by getting {
            dependencies {
                implementation("com.juul.kable:core:main-SNAPSHOT")
            }
        }
    }
}
```

solvek:
Thanks a lot.
Tested this locally and fixed a small bug.

twyatt:
I really like the idea of only performing operations on characteristics that identify as having the corresponding capabilities (i.e. only searching for characteristic that have properties that indicate they can be observed for starting observations) _but_ I worry that all to often peripherals are improperly implemented and might not be setup with the correct properties. 

I'm thinking this functionality should should be configurable; the hard part thinking of what the terminology would be for the configuration for how to search for characteristics.

What you have in this PR (which is great) I would refer to as **strict** search mode, but not quite sure what I would call the configuration parameter. Seems like a good fit for the [`peripheral` builder lambda](https://github.com/JuulLabs/kable#peripheral) though.

twyatt:
@solvek thanks again for this PR! Apologies I haven't given it the attention it deserves. I haven't forgotten about it and I'll try to find time soon to get it reviewed and/or merged.

twyatt:
Apologies it has taken so long to follow up on this.

Took some time and was going to update this PR with some minor implementation updates but then ran into the following unknowns that I'm not sure the best way this should be approached.

Imagine the following scenario; a user has a peripheral with the following characteristic (and corresponding properties):

| Characteristic | Properties |
|-|-|
| `A` | Indicate |
| `B` | Write |
| `C` | Notify |

Let's assume they all have the same UUID (definitely a poorly structured service/characteristic tree, but lets go with this extreme as an example).

For `Peripheral.observe`, which characteristic should we observe? If we do an iterative search then `A` would technically be a good match, but `C` is also a good candidate (and for most users Notify is usually what would be preferred).

I'd imagine it'd be good to prefer characteristic properties (for `observe`) in the following order:

- Notify
- Indicate
- _first match_

Whereas we search all characteristics with matching UUID, if we find Notify then our search is done, if not then search for Indicate, if not found then first match. Perhaps something along the lines of?:

```kotlin
private fun List<PlatformService>.findObserveCharacteristic(
    serviceUuid: Uuid,
    characteristicUuid: Uuid,
): PlatformCharacteristic {
    val characteristics = first(serviceUuid).characteristics

    var indicateCharacteristic: PlatformCharacteristic? = null
    var firstMatch: PlatformCharacteristic? = null
    for (characteristic in characteristics) {
        if (characteristic.characteristicUuid == characteristicUuid) {
            val properties = characteristic.bluetoothGattCharacteristic.properties
            if (properties and PROPERTY_NOTIFY != 0) return characteristic
            if (indicateCharacteristic == null && properties and PROPERTY_INDICATE != 0) indicateCharacteristic = characteristic
            if (firstMatch == null) firstMatch = characteristic
        }
    }
    if (indicateCharacteristic != null) return indicateCharacteristic
    if (firstMatch != null) return firstMatch
    throw NoSuchElementException("No compatible characteristic found for observation matching service $serviceUuid, characteristic $characteristicUuid")
}
```

solvek:
I see the point.
Because for both Indicate and Notify we use the same method there can be a conflict.
But in reality I do not think there can be such case. Most probably the same UUID will be used for Read/Notify or Write/Indicate.
So, your idea with priority should be working fine.

twyatt:
> I see the point.
> Because for both Indicate and Notify we use the same method there can be a conflict.
> But in reality I do not think there can be such case. Most probably the same UUID will be used for Read/Notify or Write/Indicate.
> So, your idea with priority should be working fine.

Due to the complexity of the new characteristic search mechanism, I'd like to put it behind a configuration flag. I'll try to find some time soon to update the PR and get it merged.

solvek:
Hello
Is ths PR planned for 0.11 version?

twyatt:
> Hello Is ths PR planned for 0.11 version?

Apologies, I've neglected this PR for way too long.
I should've been way more transparent about continually prioritizing other features/bugs.

I was honestly struggling to decide on the appropriate API for making this behavior configurable. I believe the implementation of this PR is more correct than the current implementation, but I worry that many library users may rely on the current behavior.

All that being said, that you for pinging me about this, I have been swamped at work with other projects (that depend on other priorities for this library) but this PR certainly deserves the last bit of effort to get it integrated. I will do my best to find some time soon.

Again, sorry for the very long delay.

twyatt:
Update on this PR: I've begun work on a branch that performs the same functionality of this PR while also addressing #93, and supports multiplatform.

Once I have the related PR up, I'll close this PR.

twyatt:
Superseded by #238.

PR #131, opened 2021-07-04
Add logging support

### LINKED ISSUE #130: Provide mechanism for debug logging
To aid in debugging bluetooth issues, Kable should provide a way to configure logging.

The following are possible log levels that should be configurable per `Peripheral`:

| Log level | Description |
|-|-|
| `Warnings` | Disables debug logging and only logs warnings when unexpected failures occur. |
| `Events` | Logs warnings and all events. |
| `Data` | Logs warnings, all events and hex representation of I/O data. |

twyatt:
Closes #130


core/src/commonMain/kotlin/HexString.kt
```diff
@@ -0,0 +1,23 @@
+package com.juul.kable
+
+/**
+ * Copied from Tuulbox.
+ * https://github.com/JuulLabs/tuulbox/blob/fde4eb74d2aeb37b6aaac2002bccc553adc02c5a/encoding/src/commonMain/kotlin/HexString.kt
```
cedrickcooke, line 5: Why not just include the dependency? `encoding` is a small artifact, and Tuulbox 4.3.0 added support for all of the targets we use.

core/src/commonMain/kotlin/logs/Logging.kt
```diff
@@ -0,0 +1,51 @@
...
+         *   detail1: value1
+         *   detail2: value2
+         *   ...
+         * ```
+         */
+        Multiline,
+    }
+
+    public fun interface DataProcessor {
+        public fun process(data: ByteArray): String
```
cedrickcooke, line 44: Are there ever going to be `ByteArray` details that aren't just the `data` key? If there's potential to have them under multiple keys in the future, I'd love to see a `key` field here even if unused for now. It's a breaking change to retrofit in the future, but only a `_, ` to ignore an extra field.

cedrickcooke (approved):

core/src/commonMain/kotlin/HexString.kt
twyatt: Good call out. At the time I copied this, I think I was blocked by missing some targets in `encoding` that Kable needed. Perhaps that was resolved with https://github.com/JuulLabs/tuulbox/pull/68. I'll give it another try.

core/src/commonMain/kotlin/logs/Logging.kt
twyatt: I was thinking a similar thing when I wrote this, that it would be good to know move about the `LogMessage` in the data processor. Knowing this like what characteristic is being written to or read from might drive how the `ByteArray` is processed.

I thought about providing `process` function with the `LogMessage` as well, but it is an internal class (due to its mutability — I wouldn't want to expose it here).

What you're asking specifically is concerning if we later have multiple fields that we want to log that hold data, right?
If, for example, when writing a characteristic it had both the payload it is going to write and potentially some other binary data associated with the write operation?

In which case, would the signature of this function change to?:

```suggestion
        public fun process(key: String, data: ByteArray): String
```

...and then it would be called twice for that write operation? Once for the payload (e.g. `key` of `data`) and again for the additional (we'll call it `extra`) data?

log-engine-tuulbox/src/commonMain/kotlin/logs/TuulboxLogEngine.kt
```diff
@@ -0,0 +1,30 @@
+package com.juul.kable.logs
+
+import com.juul.tuulbox.logging.HideFromStackTraceTag
+import com.juul.tuulbox.logging.Log
+
+public object TuulboxLogEngine : LogEngine, HideFromStackTraceTag {
```
cedrickcooke, line 6: This is my bad, but it occurs to me now that it would be nice to have an `object Kable : Key<Boolean>` so we can set `metadata[Kable] = true` on these log calls.

Not a breaking change to add this later, so if you would rather convert to issue feel free.

core/src/commonMain/kotlin/logs/Logging.kt
cedrickcooke: > What you're asking specifically is concerning if we later have multiple fields that we want to log that hold data, right? If, for example, when writing a characteristic it had both the payload it is going to write and potentially some other binary data associated with the write operation?

Yes, that's what I was asking about. But your example, exposing the characteristic, is _way more interesting_. A future-proof way to leave room to add fields is to do something like this:

```suggestion
        public fun process(metadata: Metadata, data: ByteArray): String
        public class Metadata internal constructor()
```

Adding fields to a class is a minor/non-breaking change, and we don't have to commit to anything extra right now.

PR #137, opened 2021-07-09
Throw exception when Android's `BluetoothAdapter` is not `On`

### LINKED ISSUE #136: `connect` should not attempt to connect if BLE is off
#133 added support for explicitly closing BLE connections when BLE is turned off on the phone, but testing (with a Pixel 3a running Android 11) the functionality on the [SensorTag sample](https://github.com/JuulLabs/sensortag) resulted in the auto-reconnect reconnecting upon BLE being turned off.

Admittedly, it is super strange that Android allows a connection to be re-established while BLE is off, but Kable should probably handle this and prevent `connect` from reconnecting while BLE is off.

cedrickcooke:
Fixes #136.

### DIFF
--- a/core/src/androidMain/kotlin/Peripheral.kt
+++ b/core/src/androidMain/kotlin/Peripheral.kt
@@ -201,21 +201,7 @@ public class AndroidPeripheral internal constructor(
 
     public override suspend fun connect() {
         check(job.isNotCancelled) { "Cannot connect, scope is cancelled for $this" }
-
-        // Explicitly check the adapter state before connecting in order to respect system settings.
-        // Android doesn't actually turn bluetooth off when the setting is disabled, so without this
-        // check we're able to reconnect the device illegally.
-        val adapterState = BluetoothAdapter.getDefaultAdapter().state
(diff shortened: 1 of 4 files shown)


core/src/androidMain/kotlin/Peripheral.kt
```diff
@@ -197,6 +201,21 @@ public class AndroidPeripheral internal constructor(
...
+        val adapterState = BluetoothAdapter.getDefaultAdapter().state
+        if (adapterState != STATE_ON) {
+            val stateName = when (adapterState) {
+                STATE_OFF -> "Off"
+                STATE_TURNING_OFF -> "TurningOff"
+                STATE_TURNING_ON -> "TurningOn"
+                else -> "Unknown"
+            }
+            throw BluetoothDisabledException("Cannot connect to device while bluetooth adapter state is $stateName.")
+        }
```
twyatt, line 217: nit: Can we pull this out into a separate function (e.g. `checkBluetoothState`)?

twyatt (approved):

sdonn3 (approved):

PR #141, opened 2021-07-13
Specify Kable in `common` source set of example Gradle configuration

twyatt:
Found a Gradle configuration that works for both application and library Gradle configurations.

Now using (tested) this configuration in https://github.com/JuulLabs/sensortag/pull/10.

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,8 +295,10 @@ kotlin {
 
         val macosX64Main by getting {
             dependencies {
+                // Need to specify the Coroutines artifact specific for the target platform (`-macosx64`):
                 implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core-macosx64:${coroutinesVersion}-native-mt") {
                     version {
+                        // `strictly` needed to make sure Gradle uses `-native-mt` version.
                         strictly("${coroutinesVersion}-native-mt")
                     }
                 }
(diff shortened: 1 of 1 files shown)


sdonn3 (approved):

cedrickcooke (approved):

README.md
```diff
@@ -282,43 +282,40 @@ kotlin {
...
+                api("org.jetbrains.kotlinx:kotlinx-coroutines-core:${coroutinesVersion}")
                 implementation("com.juul.kable:core:${kableVersion}")
             }
         }
 
-        val androidMain by getting {
+        val macosX64Main by getting {
             dependencies {
-                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:${coroutinesVersion}")
+                api("org.jetbrains.kotlinx:kotlinx-coroutines-core-macosx64:${coroutinesVersion}-native-mt") {
```
QuantumRand, line 292: What's the need behind exposing this library from the iOS/Mac Native modules?

QuantumRand (approved):
Why

README.md
twyatt: `api` is needed if any of Kable (that exposes Coroutines classes) is used on the consuming side.

For example, if someone is writing an app that uses Kable and is being written in Kotlin and targeting iOS then they'll need Coroutines on their classpath (e.g. for `Flow` availability). I imagine that most people will define Coroutines in their app's `build.gradle.kts`, but defining as `api` ensures that it will work for them even if they don't define Coroutines in their app.

twyatt: Oh, I think I understand what you were getting at.
The `commonMain` already exposes the core APIs for Coroutines, so other specialized dependencies don't need to be `api`.
I'll update that. Good catch.

PR #146, opened 2021-07-20
Fix `requestPeripheral` when using Kotlin/JS IR compiler

twyatt:
Was seeing:

```
TypeError: Failed to execute 'requestDevice' on 'Bluetooth': Either 'filters' should be present or 'acceptAllDevices' should be true, but not both.
```

Suspected to be caused by an incompatibility with using an anonymous `object`. This PR instead uses `kotlin-wrappers` library provided `jsObject`.

### DIFF
--- a/core/src/jsMain/kotlin/Bluetooth.kt
+++ b/core/src/jsMain/kotlin/Bluetooth.kt
@@ -1,5 +1,8 @@
 package com.juul.kable
 
+import com.juul.kable.Options.Filter.Name
+import com.juul.kable.Options.Filter.NamePrefix
+import com.juul.kable.Options.Filter.Services
 import com.juul.kable.external.Bluetooth
 import kotlinext.js.jsObject
 import kotlinx.coroutines.CoroutineScope
@@ -15,6 +18,25 @@ public fun CoroutineScope.requestPeripheral(
     .requestDevice(options.toDynamic())
     .then { device -> peripheral(device, builderAction) }
 
+/**
+ * Converts [Options] to JavaScript friendly object.
+ *
+ * According to the `requestDevice`
+ * [example](https://developer.mozilla.org/en-US/docs/Web/API/Bluetooth/requestDevice#example), the form of the
+ * JavaScript object should be similar to:
+ * ```
+ * let options = {
+ *   filters: [
+ *     {services: ['heart_rate']},
+ *     {services: [0x1802, 0x1803]},
+ *     {services: ['c48e6067-5295-48d3-8d5c-0395f61792b1']},
+ *     {name: 'ExampleName'},
+ *     {namePrefix: 'Prefix'}
+ *   ],
+ *   optionalServices: ['battery_service']
+ * }
+ * ```
+ */
 private fun Options.toDynamic(): dynamic = if (filters == null) {
     jsObject {
         this.acceptAllDevices = true
@@ -23,6 +45,13 @@ private fun Options.toDynamic(): dynamic = if (filters == null) {
 } else {
     jsObject {
         this.optionalServices = optionalServices
-        this.filters = filters
+        this.filters = filters.map { it.toDynamic() }.toTypedArray()
     }
 }
+
+private fun Options.Filter.toDynamic(): dynamic =
+    when (this) {
+        is Name -> jsObject { this.name = name }
+        is NamePrefix -> jsObject { this.namePrefix = namePrefix }
+        is Services -> jsObject { this.services = services }
+    }
--- a/core/src/jsMain/kotlin/Bluetooth.kt
+++ b/core/src/jsMain/kotlin/Bluetooth.kt
@@ -1,6 +1,7 @@
 package com.juul.kable
 
 import com.juul.kable.external.Bluetooth
+import kotlinext.js.jsObject
 import kotlinx.coroutines.CoroutineScope
 import kotlin.js.Promise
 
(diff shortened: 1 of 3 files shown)


core/src/jsMain/kotlin/Bluetooth.kt
```diff
@@ -15,13 +16,13 @@ public fun CoroutineScope.requestPeripheral(
...
     }
 } else {
-    object {
-        val optionalServices = this@toDynamic.optionalServices
-        val filters = this@toDynamic.filters
+    jsObject {
+        this.optionalServices = optionalServices
+        this.filters = filters
     }
 }
```
davidtaylor-juul, line 28: The filters need to be converted to POJOs containing native arrays and/or types. I'm wondering how this even worked prior to the compiler upgrade?

```suggestion
        this.filters = filters.map { it.toDynamic() }.toTypedArray()
    }
}

private fun Options.Filter.toDynamic(): dynamic =
    when (this) {
        is Options.Filter.Name -> jsObject { this.name = name }
        is Options.Filter.NamePrefix -> jsObject { this.namePrefix = namePrefix }
        is Options.Filter.Services -> jsObject { this.services = services }
    }
```

davidtaylor-juul: > I'm wondering how this even worked

Just realized it probably didn't but nobody noticed unless they are relying on a specific filter...

twyatt: Are you wondering how the arrays worked?
I _believe_ it worked using the built in interop: https://kotlinlang.org/docs/js-to-kotlin-interop.html#kotlin-types-in-javascript
So Kotlin Arrays are automatically mapped to JS Arrays.

davidtaylor-juul: For reference, the object shape we are going for it like this example (from https://developer.mozilla.org/en-US/docs/Web/API/Bluetooth/requestDevice):
```javascript
let options = {
  filters: [
    {services: ['heart_rate']},
    {services: [0x1802, 0x1803]},
    {services: ['c48e6067-5295-48d3-8d5c-0395f61792b1']},
    {name: 'ExampleName'},
    {namePrefix: 'Prefix'}
  ],
  optionalServices: ['battery_service']
}
```

twyatt: Ohh, I see. The POJO mapping is undocumented and is surprising that worked before. I guess that legacy compiler did some magic that we were relying on that it didn't even promise it would do? 🤷

twyatt: Since Arrays are mapped automatically, is `this.filters = filters.map { it.toDynamic() }.toTypedArray()` needed?

twyatt: > For reference, the object shape we are going for it like this example (from https://developer.mozilla.org/en-US/docs/Web/API/Bluetooth/requestDevice):
> 
> ```js
> let options = {
>   filters: [
>     {services: ['heart_rate']},
>     {services: [0x1802, 0x1803]},
>     {services: ['c48e6067-5295-48d3-8d5c-0395f61792b1']},
>     {name: 'ExampleName'},
>     {namePrefix: 'Prefix'}
>   ],
>   optionalServices: ['battery_service']
> }
> ```

Thanks for this. Should've started back here, been too long since I've look at this. I'll get this fixed up. 👍

davidtaylor-juul: The problem here is not obvious, and I haven't investigated whether this worked correctly on the legacy compiler but some testing shows this to fail now with the IR compiler unless we map to JS compatible objects. A unit test would perhaps be handy here.

`optionalServices ` works because it is `Array<String>`, but `filters` needs to be an `Array<Object>` where each `Object` has an index operator with a key mapping to one of `services`, `name` or `namePrefix`.

For `services`, the value may be `Array<String>`, and for the others the value must be `String`.

It's possible the legacy compiler happened to convert `class Services(val services: Array<String>)` to something resembling `{ services: [string] }`, even if it was a Kotlin class and not a native `Object`. Given the shape of the source code is the same it certainly seems like it should work. I was curious so dumped the raw object from the browser and it actually produces this shape:

```
Services
  _services: Array(1)
    0: "0000fde4-0000-1000-8000-00805f9b34fb"
    length: 1
    __proto__: Array(0)
  __proto__: Filter
    constructor: ƒ Services(services)
    equals: ƒ (other)
    hashCode: ƒ ()
    toString: ƒ ()
    __proto__: Object
```

So it would appear the IR compiler (rightly) fully preserves the class hierarchy, and keeps any instance variables private (via underbar prefix e.g. `_services`). My guess is there needs to be `@JsExport` annotations and appropriate `public` scoping such that the desired shape is exposed explicitly.
 
As for `Array.map` it emits `ArrayList` (I think?), so `toTypedArray()` is always needed to get back to a JS compatible `Array.

davidtaylor-juul: Out of curiosity I hacked in `@JsExport` and that is indeed the difference. After doing so the object looks like this:
```
Services
  _services: ["0000ffff-0000-1000-8000-000000000000"]
  services: (...)
  __proto__: Filter
    component1: ƒ ()
    constructor: ƒ Services(services)
    copy: ƒ (services)
    copy$default_2taqey_k$: ƒ (services, $mask0, $handler)
    copy_ortqkv_k$: ƒ (services)
    equals: ƒ (other)
    hashCode: ƒ ()
    toString: ƒ ()
    _get_services__0_k$: ƒ ()
    services: (...)
    get services: ƒ ()
    __proto__: Object
```

Now it has an actual `services` property! This would work directly without all that `jsObject` noise.

*unresolving this thread temporarily to highlight this additional comment

davidtaylor-juul: Side note: manual conversion via `jsObject` produces much tighter JS code. The class hierarchy and `data class` extensions, and the custom getters adds a lot of weight.

davidtaylor-juul (approved):

QuantumRand (approved):

sdonn3 (approved):

PR #149, opened 2021-08-07
Fix web listener bug

davidtaylor-juul:
For Web targets, system would hang after the first read. This seems to be the culprit.

### DIFF
--- a/core/src/jsMain/kotlin/Peripheral.kt
+++ b/core/src/jsMain/kotlin/Peripheral.kt
@@ -265,8 +265,7 @@ public class JsPeripheral internal constructor(
     }
 
     internal suspend fun stopObservation(characteristic: Characteristic) {
-        val listener = observationListeners[characteristic] ?: return
-        observationListeners.remove(characteristic)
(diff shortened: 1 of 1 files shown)


cedrickcooke (approved):

sdonn3 (approved):

core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -266,6 +266,7 @@ public class JsPeripheral internal constructor(
 
     internal suspend fun stopObservation(characteristic: Characteristic) {
         val listener = observationListeners[characteristic] ?: return
+        observationListeners.remove(characteristic)
```
twyatt, line 269:
```suggestion
        val listener = observationListeners.remove(characteristic) ?: return
```

twyatt (commented):
Wow, thanks so much for tracking this down.
🤦  such a silly mistake on my part.

twyatt (approved):

twyatt:
Released in 0.7.2

PR #151, opened 2021-08-20
Add configurable `LogEngine` for less opinionated logging

cedrickcooke:
I got a little carried away while trying to write a response in JuulLabs/tuulbox#71. I think this is a nice solution for providing logging-by-default in Kable while also providing a close to one-line swap-out from the default logging to our configurable Tuulbox logging.

While I was implementing this, I removed the rest of the `android.util.Log` calls and did a little bit of package/file shuffling.


log-engine-tuulbox/src/androidMain/AndroidManifest.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest
+    package="com.juul.kable"
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    />
```
twyatt, line 6:
```suggestion
<manifest package="com.juul.kable" />
```

log-engine-tuulbox/build.gradle.kts
```diff
@@ -0,0 +1,63 @@
+plugins {
+    // Android plugin must be before multiplatform plugin until https://youtrack.jetbrains.com/issue/KT-34038 is fixed.
+    id("com.android.library")
+    kotlin("multiplatform")
+    id("kotlinx-atomicfu")
```
twyatt, line 5:
```suggestion
```

log-engine-tuulbox/build.gradle.kts
```diff
@@ -0,0 +1,63 @@
...
+            languageSettings.enableLanguageFeature("InlineClasses")
+        }
+    }
+}
+
+atomicfu {
+    transformJvm = true
+    transformJs = false
+}
+
```
twyatt, line 41:
```suggestion
```

log-engine-tuulbox/build.gradle.kts
```diff
@@ -0,0 +1,63 @@
...
+        val commonMain by getting {
+            dependencies {
+                api(project(":core"))
+                implementation(tuulbox("logging"))
+            }
+        }
+
+        all {
+            languageSettings.enableLanguageFeature("InlineClasses")
+        }
```
twyatt, line 33: This feature isn't used in this module.

```suggestion
```

log-engine-tuulbox/build.gradle.kts
```diff
@@ -0,0 +1,63 @@
...
+android {
+    compileSdkVersion(30)
+
+    defaultConfig {
+        minSdkVersion(21)
+    }
+
+    buildFeatures {
+        buildConfig = false
+    }
```
twyatt, line 51: nit: I believe this can be configured via `gradle.properties`? If so, would be better to do it there and drop this configuration in the other modules as well.

core/src/androidMain/kotlin/gatt/Callback.kt
```diff
@@ -286,8 +284,10 @@ private val Int.connectionStateString: String
...
     }
 
-private fun <E> SendChannel<E>.trySendOrLog(element: E) {
-    trySend(element).getOrElse { cause ->
-        Log.w(TAG, "Callback was unable to deliver $element", cause)
+private fun <E> Logger.trySendOrLog(channel: SendChannel<E>, element: E) {
+    channel.trySend(element).getOrElse { cause ->
+        warn(cause) {
+            message = "Callback was unable to deliver $element"
+        }
```
twyatt, line 291: I personally would rather this remain an extension function on `SendChannel` and be moved as a class function (not top-level) so that call sites would look like:

```kotlin
onResponse.trySendOrLog(event)
```

twyatt (approved):
Awesome work! Love it!

PR #153, opened 2021-08-27
Add Connecting sub-states; delay Connected state until ready

### LINKED ISSUE #75: IllegalStateException: Services have not been discovered for Peripheral
Hello guys.

I get the IllegalStateException when trying read a characteristic inside peripheral state collect.

```
peripheral.state.collect { state ->
    if (state == State.Connected) {
        // delay(200); // this line fix the exception
        peripheral.read(characteristic)
    }
}
```

This behavioural is normal because the services are discovered after peripheral connection is established. The commented line fixes the exception because the delay permits the peripheral to discover the services before the reading is performed.

Maybe we can change this behaviour. My suggestion is: update the peripheral state to CONNECTED after the services discovery is performed or creating a new peripheral state SERVICES_DISCOVERED.

What do you think?

cedrickcooke:
Closes #75.


sdonn3 (approved):
All seems reasonable to me

core/src/commonMain/kotlin/Peripheral.kt
```diff
@@ -138,3 +140,19 @@ public interface Peripheral {
         onSubscription: OnSubscriptionAction = {},
     ): Flow<ByteArray>
 }
+
+/** Remember the bluetooth connected is not the same thing as Kable's [State.Connected]. */
```
twyatt, line 144: I think it will be easier to read if we're more explicit.

```suggestion
/**
 * Suspends until [Peripheral] receiver is connected at the bluetooth layer, but [Peripheral] not yet ready for general
 * use (i.e. services have not been discovered and observations have not been wired up).
 *
 * @see suspendUntilConnected for implementation that suspends until ready (Kable's [State.Connected]).
 */
```

core/src/commonMain/kotlin/Peripheral.kt
```diff
@@ -138,3 +140,19 @@ public interface Peripheral {
...
+internal suspend fun Peripheral.suspendUntilConnectingServices() {
+    state.first { it == State.Connecting.Services }
+}
+
+/** Remember the bluetooth connected is not the same thing as Kable's [State.Connected]. */
```
twyatt, line 149:
```suggestion
/**
 * Suspends until [Peripheral] receiver is connected at the bluetooth layer, but [Peripheral] not yet ready for general
 * use (i.e. services have not been discovered and observations have not been wired up).
 *
 * @see suspendUntilConnected for implementation that suspends until ready (Kable's [State.Connected]).
 */
```

core/src/commonMain/kotlin/Peripheral.kt
```diff
@@ -138,3 +140,19 @@ public interface Peripheral {
...
+internal suspend fun Peripheral.suspendUntilConnectingServicesOrThrow() {
+    state
+        .onEach { if (it is State.Disconnected) throw ConnectionLostException() }
+        .first { it == State.Connecting.Services }
+}
+
+internal suspend fun Peripheral.suspendUntilConnected() {
+    state.first { it == State.Connected }
+}
```
twyatt, line 158:
```suggestion
/**
 * Suspends until [Peripheral] receiver is ready for general user (i.e. bluetooth layer is connected, services have been
 * discovered and observations have been wired up).
 */
internal suspend fun Peripheral.suspendUntilConnected() {
    state.first { it == State.Connected }
}
```

core/src/commonMain/kotlin/State.kt
```diff
@@ -2,7 +2,11 @@ package com.juul.kable
 
 public sealed class State {
 
-    public object Connecting : State()
+    public sealed class Connecting : State() {
+        public object Bluetooth : Connecting()
+        public object Services : Connecting()
+        public object Observers : Connecting()
```
twyatt, line 8: I'm torn if `Observes` or `Observations` would be better?

We are indicating that we're wiring up the observations for flows created via calls to `observe(..)` function. 🤷

core/src/commonMain/kotlin/Peripheral.kt
```diff
@@ -138,3 +140,19 @@ public interface Peripheral {
         onSubscription: OnSubscriptionAction = {},
     ): Flow<ByteArray>
 }
+
+/** Remember the bluetooth connected is not the same thing as Kable's [State.Connected]. */
+internal suspend fun Peripheral.suspendUntilConnectingServices() {
+    state.first { it == State.Connecting.Services }
+}
```
twyatt, line 147: Perhaps we should change this to?:

```suggestion
internal suspend fun Peripheral.suspendUntil(state: State) {
    state.first { it == state }
}
```

It barely does anything and almost feels we should just use `state.first` where we were calling this function, but I think the syntax sugar does make the code easier to read at the call-site.

But if we pass in the state it makes the code a bit more self documenting and we can get rid of a lot (if not all) of the KDoc suggestions I made, for example:

```kotlin
peripheral.suspendUntil(Connecting.Services)
```

Command clicking `Connecting.Services` can take you to the KDoc for it which would explain when that state is emitted (basically a re-write of the KDoc I suggested above.

core/src/commonMain/kotlin/State.kt
sdonn3: Observations seems best to me? Seems like we're waiting until observations are set up, not the observers

sdonn3: Went with observes in this scenario.

twyatt:
If you feel inclined the update the `Peripheral.state` KDoc:

```kotlin
    /**
     * Provides a conflated [Flow] of the [Peripheral]'s [State].
     *
     * After [connect] is called, the [state] will typically transition through the following [states][State]:
     *
     * ```
     *           connect()
     *               :
     *               v
     *   .----------------------.
     *   | Connecting.Bluetooth |
     *   '----------------------'
     *               |
     *               v
     *    .---------------------.
     *    | Connecting.Services |
     *    '---------------------'
     *               |
     *               v
     *    .--------------------.       .-----------.
     *    | Connecting.Observe | ----> | Connected |
     *    '--------------------'       '-----------'
     *                                       :
     *                                disconnect() or
     *                                connection drop
     *                                       :
     *                                       v
     *                               .---------------.       .--------------.
     *                               | Disconnecting | ----> | Disconnected |
     *                               '---------------'       '--------------'
     * ```
     */
    public val state: Flow<State>
```

twyatt:
Here is an updated `artwork/connection-states.png` (so that we show the new states in the project README):

connection-states

core/src/commonMain/kotlin/Peripheral.kt
```diff
@@ -138,3 +150,29 @@ public interface Peripheral {
...
+
+/**
+ * Suspends until [Peripheral] receiver arrives at the [State] specified.
+ *
+ * @see State for a description of the potential states.
+ * @throws ConnectionLostException if peripheral state arrives at [State.Disconnected].
+ */
+internal suspend inline fun <reified T : State> Peripheral.suspendUntilOrThrow() {
+    require(T::class != State.Disconnected::class) {
+        "Peripheral.suspendUntilThrow() throws on State.Disconnected, not intended for use with that State."
```
cedrickcooke, line 173: Nit:
```suggestion
        "Peripheral.suspendUntilOrThrow() throws on State.Disconnected, not intended for use with that State."
```

twyatt (approved):

PR #154, opened 2021-08-29
Android observe should suspend until Connecting.Observes OR Connected

cedrickcooke:


core/src/commonMain/kotlin/State.kt
```diff
@@ -108,3 +110,34 @@ public sealed class State {
...
+ */
+internal inline fun <reified T: State> State.isAtLeast(): Boolean {
+    val currentState = when (this) {
+        is State.Disconnected -> 0
+        State.Disconnecting -> 1
+        State.Connecting.Bluetooth -> 2
+        State.Connecting.Services -> 3
+        State.Connecting.Observes -> 4
+        State.Connected -> 5
+        else -> error("Unreachable.")
```
QuantumRand, line 129: With `State` being a sealed class, isn't this exhaustive without the `else`?

core/src/commonMain/kotlin/State.kt
```diff
@@ -108,3 +110,34 @@ public sealed class State {
...
+    }
+    val targetState = when (T::class) {
+        State.Disconnected::class -> 0
+        State.Disconnecting::class -> 1
+        State.Connecting.Bluetooth::class -> 2
+        State.Connecting.Services::class -> 3
+        State.Connecting.Observes::class -> 4
+        State.Connected::class -> 5
+        else -> error("Unreachable.")
```
QuantumRand, line 138: Same with here.

twyatt (approved):
Tested working with the SensorTag sample app.
Thanks for the quick fix!

davidtaylor-juul (approved):

core/src/commonMain/kotlin/State.kt
cedrickcooke: I believe so, but it refused to compile (and I'm not sure why).

core/src/commonMain/kotlin/State.kt
cedrickcooke: I believe so, but it refused to compile (and this time it makes sense, because as far as Kotlin is concerned it can be _any_ `KClass`).

core/src/commonMain/kotlin/State.kt
QuantumRand: I guess the compiler still can't handle nested sealed classes. Human still smarter than computer...I'll chalk that up as a win

sdonn3 (approved):
👍

core/src/commonMain/kotlin/State.kt
cedrickcooke: Nevermind. The IDE is highlighting is at an error but it actually does compile. Thanks for making me double check.

QuantumRand: Human smarter than IDE...still win?

PR #155, opened 2021-08-29
Fix characteristic observations on Apple and JavaScript targets

twyatt:
Applies same fix as #154 to Apple and JavaScript targets.


davidtaylor-juul (approved):
LGTM

cedrickcooke (approved):

PR #159, opened 2021-09-03
Android ScanSettings

mtrewartha:
Per [this discussion](https://github.com/JuulLabs/kable/discussions/158), this PR enables users to create a `Scanner` with Android-specific `ScanSettings` if the current default ones don't suffice. It simply adds a non-`actual` `var scanSettings` to the builder that's available when calling from an Android source set only. It defaults to the existing default. This means that users who don't care about customizing this can still invoke the DSL from a common source set, but those who want more control can invoke from an Android source set and get the control they need.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ The [`Scanner`] may be configured via the following DSL (shown are defaults, whe
 ```kotlin
 val scanner = Scanner {
     services = null
+    scanSettings = ScanSettings.Builder().build() // Android only, see details below
     logging {
         engine = SystemLogEngine
         level = Warnings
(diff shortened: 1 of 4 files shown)


twyatt:
@mtrewartha looks great! Thanks so much!

Some minor tweaks I plan to make before merging:

1. Documentation formatting changes
2. Add `@ObsoleteKableApi` annotation and mark the `scanSettings` property with the annotation

Long term, I hope to provide all the `ScanSettings` options as native `Scanner` DSL builder options rather than having to use Android's directly. Although it is very low priority and will probably be a while before that changes, so this implementation is a great solution in the mean time. The annotation will simply warn consumers that the API is planned to change.

mtrewartha:
@twyatt sounds great! I definitely agree on the Scanner DSL front. I'm arguably less familiar with iOS, macOS, and JS BLE scanning capabilities. I poked around to see if those platforms had anything that made sense to include in some multiplatform "settings" notion that could be added, I just didn't see anything and that's why I opted for the Android-specific one for now. That annotation sounds good in the interim 👍🏻

twyatt:
@sdonn3 when you have a chance to review, please see updated changes via https://github.com/JuulLabs/kable/compare/main...twyatt/android-scan-settings.

twyatt (approved):
Will merge https://github.com/JuulLabs/kable/compare/main...twyatt/android-scan-settings manually once approved.

sdonn3 (approved):
lgtm!

twyatt:
Merged manually. Thanks for the contribution @mtrewartha!

This will be available in the next release. I had planned to get some additional changes into the next release, but let me know if you have a more immediate need for this and I can cut a release sooner.

mtrewartha:
Great, thanks @twyatt! We do have an immediate need for this change, but we can just use my fork for now, then swap back to the new release whenever it's ready.

Issue #157, opened 2021-09-01
`write` suspends indefinitely on subsequent calls when `WithoutResponse` is used


renatosmf:
I’m attempting to send a write command sequentially. This approach works over android Bluetooth API, but in iOS I’m receiving the notification that the characteristics change only for the first command.

For the iOS write command, I'm using the write without response. In another way, If I set to write with the response, I'm receiving the error for the write permission.

twyatt:
Thanks for the report!

Can you provide logs using the [Kable logging](https://github.com/JuulLabs/kable#logging) mechanism? Either log `level` either `Data` or `Events`, please.

renatosmf:
> Thanks for the report!
> 
> Can you provide logs using the [Kable logging](https://github.com/JuulLabs/kable#logging) mechanism? With log `level` either `Data` or `Events`, please.

@twyatt Do you want Scanner log level or Perfipheral?

twyatt:
> I’m attempting to send a write command sequentially

If I understand correctly, you have a characteristic you're writing to and responses come back on a dedicated characteristic (via characteristic change notifications), right?

Looking at a snippet of your logs, is this where it is problematic?

```
2021-09-01 18:23:44.530598-0300 iosApp[23165:8562722] D/Kable: 7BCADA26-F146-A594-B32A-1DE58DBCE7AF write
  service: 0000ffe0-0000-1000-8000-[CENSURED]
  characteristic: 0000ffe1-0000-1000-8000-[CENSURED]
  writeType: WithoutResponse
  data: 54 53 3A 39 39 31 38 39 39 32 30 2D 33 39 30 [CENSURED]
2021-09-01 18:23:44.531612-0300 iosApp[23165:8562722] D/Kable: 7BCADA26-F146-A594-B32A-1DE58DBCE7AF write
  service: 0000ffe0-0000-1000-8000-[CENSURED]
  characteristic: 0000ffe1-0000-1000-8000-[CENSURED]
  writeType: WithoutResponse
  data: 54 53 3A 39 39 31 38 39 39 32 30 2D 33 39 30 [CENSURED]
2021-09-01 18:23:44.857803-0300 iosApp[23165:8564787] D/Kable: 7BCADA26-F146-A594-B32A-1DE58DBCE7AF didUpdateValueForCharacteristic
  service: FFE0
  characteristic: FFE1
  data: 54 3A 4F 4B 3A 56 34 2E [CENSURED]
[...]

PR #162, opened 2021-09-22
fix issue #157 iOS semaphore handler

renatosmf:
Fix semaphore configuration, changing the minimum value of simultaneous thread to 2

### DIFF
--- a/core/src/appleMain/kotlin/Connection.kt
+++ b/core/src/appleMain/kotlin/Connection.kt
@@ -8,7 +8,7 @@ internal class Connection(
 ) {
 
     // Using Semaphore as Mutex never relinquished lock when multiple concurrent `withLock`s are executed.
-    val semaphore = Semaphore(1)
+    val semaphore = Semaphore(2)
 
     suspend inline fun <T> execute(
         action: () -> Unit,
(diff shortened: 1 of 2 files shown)


twyatt:
I really appreciate looking into the issue you're having and providing a PR to fix it, though I would like to get a little more information before we move forward with this fix.

It isn't clear to me why this would solve any BLE I/O issues. Android I/O goes through a `Mutex` (which doesn't allow parallelism) so why is this needed on Apple platforms?

On Android you can definitely only have a single I/O operation occurring at a time, so the `Mutex` is critical. I do remember reading that Core Bluetooth can support more than one operation at a time, but I don't know the details. Though I'm hesitant to make a change like this without understanding why it is needed on Apple platform and not others?

You had previously described that you were trying to sequentially perform write-for-response style operations, why would you need parallelism if the intent was to perform sequential operations?

Thank you!

renatosmf:
@twyatt I rushed into the possible solution to the problem. You were right about changing the value of the semaphore, it would simply be going to work with multithreading.
What I identified when performing more in-depth tests is that the semaphore is not being released for the next task in the queue. I can't say why it doesn't return to the available state and running next to the queue, but the fact that the problem is running at this location in the context of iOS

PR #166, opened 2021-09-29
Make Android connection failures more robust

cedrickcooke:
Fixes two separate issues, but they're close enough I think it's okay to couple them.

1. `establishConnection` can throw. Before this PR that exception wouldn't make it to the `logger`, now it will.
2. In the `catch` block of `connectAsync` call `disconnect` instead of `closeConnection`. This should guarantee that we've reached the `Disconnected` state before we actually throw the exception into the connection attempt.

### DIFF
--- a/core/src/androidMain/kotlin/Peripheral.kt
+++ b/core/src/androidMain/kotlin/Peripheral.kt
@@ -193,14 +193,14 @@ public class AndroidPeripheral internal constructor(
 
     /** Creates a connect [Job] that completes when connection is established, or failure occurs. */
     private fun connectAsync() = scope.async(start = LAZY) {
-        val connection = establishConnection().also { _connection = it }
-
-        connection
-            .characteristicChanges
-            .onEach(observers.characteristicChanges::emit)
-            .launchIn(scope, start = UNDISPATCHED)
-
         try {
+            val connection = establishConnection().also { _connection = it }
+
+            connection
+                .characteristicChanges
+                .onEach(observers.characteristicChanges::emit)
+                .launchIn(scope, start = UNDISPATCHED)
+
             suspendUntilOrThrow<State.Connecting.Services>()
             discoverServices()
             onServicesDiscovered(ServicesDiscoveredPeripheral(this@AndroidPeripheral))
@@ -208,7 +208,7 @@ public class AndroidPeripheral internal constructor(
             logger.verbose { message = "Configuring characteristic observations" }
             observers.rewire()
         } catch (t: Throwable) {
-            closeConnection()
+            disconnect()
             logger.error(t) { message = "Failed to connect" }
             throw t
         }
--- a/core/src/androidMain/kotlin/Peripheral.kt
+++ b/core/src/androidMain/kotlin/Peripheral.kt
@@ -42,8 +42,8 @@ import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.onEach
-import kotlin.coroutines.CoroutineContext
 import kotlinx.coroutines.flow.update
+import kotlin.coroutines.CoroutineContext
 
 private val clientCharacteristicConfigUuid = uuidFrom(CLIENT_CHARACTERISTIC_CONFIG_UUID)
 
--- a/core/src/androidMain/kotlin/Peripheral.kt
+++ b/core/src/androidMain/kotlin/Peripheral.kt
@@ -221,6 +221,7 @@ public class AndroidPeripheral internal constructor(
     private fun closeConnection() {
         _connection?.close()
         _connection = null
+        // Avoid trampling existing `Disconnected` state (and its properties) by only updating if not already `Disconnected`.
         _state.update { previous -> previous as? State.Disconnected ?: State.Disconnected() }
     }
 
--- a/core/src/androidMain/kotlin/Peripheral.kt
+++ b/core/src/androidMain/kotlin/Peripheral.kt
@@ -208,7 +208,7 @@ public class AndroidPeripheral internal constructor(
             logger.verbose { message = "Configuring characteristic observations" }
             observers.rewire()
         } catch (t: Throwable) {
-            disconnect()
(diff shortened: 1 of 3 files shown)


core/src/androidMain/kotlin/Peripheral.kt
```diff
@@ -193,22 +193,22 @@ public class AndroidPeripheral internal constructor(
...
+            val connection = establishConnection().also { _connection = it }
 
-        connection
-            .characteristicChanges
-            .onEach(observers.characteristicChanges::emit)
-            .launchIn(scope, start = UNDISPATCHED)
+            connection
+                .characteristicChanges
+                .onEach(observers.characteristicChanges::emit)
+                .launchIn(scope, start = UNDISPATCHED)
```
twyatt, line 202: Is it less about moving the `connection...launchIn` portion inside the `try` and more about moving the `establishConnection`?

I just want to understand the rationale. Line 199-202 won't throw an exception, at least not that I can see (it would be propagated to the `scope` anyways).

But I can see the `establishConnection` might throw, in which case getting the log statement could be helpful.

core/src/androidMain/kotlin/Peripheral.kt
```diff
@@ -193,22 +193,22 @@ public class AndroidPeripheral internal constructor(
...
-        try {
             suspendUntilOrThrow<State.Connecting.Services>()
             discoverServices()
             onServicesDiscovered(ServicesDiscoveredPeripheral(this@AndroidPeripheral))
             _state.value = State.Connecting.Observes
             logger.verbose { message = "Configuring characteristic observations" }
             observers.rewire()
         } catch (t: Throwable) {
-            closeConnection()
+            disconnect()
```
twyatt, line 211: I'm probably missing something, but what is gained with this change?
If we throw in `establishConnection` then `_connection` will still be `null` and `disconnect()` don't do anything more than `closeConnection()` did?

core/src/androidMain/kotlin/Peripheral.kt
cedrickcooke: > Is it less about moving the connection...launchIn portion inside the try and more about moving the establishConnection?

Yup. The launch just got caught up in moving it. I could put a separate try-catch around the `establishConnection` if you'd prefer.

core/src/androidMain/kotlin/Peripheral.kt
cedrickcooke: This is unrelated to the first change.

My thought (and I could be wrong here) is that the callback system could cause a brief delay between calling `bluetoothGatt.disconnect` and actually updating to the `State.Disconnected`. Using `disconnect` instead of `closeConnection` will suspend until we have updated to that state, so we can't return from `connect` while still in the `Connecting` state.

If my initial assumption is wrong I'd be happy to remove this change.

cedrickcooke: Another option is that I could keep this as `closeConnection` and instead we could add something like `if (_state.value !is State.Disconnected) { _state.value = State.Disconnected() }` to that function. That would have the added effect of defending us in the case that the callback is never invoked.

twyatt: > My thought (and I could be wrong here) is that the callback system could cause a brief delay between calling `bluetoothGatt.disconnect` and actually updating to the `State.Disconnected`.

Ohh, ya, I could definitely imagine a delay occurring there.

> Using `disconnect` instead of `closeConnection` will suspend until we have updated to that state, so we can't return from `connect` while still in the `Connecting` state.

Gotchya, now I'm starting to follow what the goal is.

> Another option is that I could keep this as `closeConnection` and instead we could add something like `if (_state.value !is State.Disconnected) { _state.value = State.Disconnected() }` to that function.

Is the `if` check necessary, would there be harm in always just setting `_state.value = State.Disconnected` in the `closeConnection()` call?

core/src/androidMain/kotlin/Peripheral.kt
twyatt: > Yup. The launch just got caught up in moving it. I could put a separate try-catch around the `establishConnection` if you'd prefer.

Naw, no need. Was just inquiring to understand.

I think it is fine how you have it now.

core/src/androidMain/kotlin/Peripheral.kt
twyatt: Looking at it more closely, I think we do always want to set `_state.value = State.Disconnected` if an exception occurs.
Otherwise if connection is rejected then we'd get stuck in `Connecting` state.

https://github.com/JuulLabs/kable/blob/6b4ca38fd30842c75fdf261e6f60602df2202501/core/src/androidMain/kotlin/Peripheral.kt#L191

...so, the combination of moving the `try` up and always setting to `Disconnected` on exception should protect against that.

With `_state` being `StateFlow` there isn't any harm in setting the same state multiple times.

cedrickcooke: I have the `if` condition in there because `State.Disconnected` has a nullable `status` field I don't want to clobber.

twyatt: Oh, good point/catch.

cedrickcooke: Updated to use `MutableStateFlow.update` to avoid race conditions.

core/src/androidMain/kotlin/Peripheral.kt
```diff
@@ -219,10 +220,8 @@ public class AndroidPeripheral internal constructor(
 
     private fun closeConnection() {
         _connection?.close()
-        if (_state.value !is State.Disconnected) {
-            _state.value = State.Disconnected()
-        }
         _connection = null
+        _state.update { previous -> previous as? State.Disconnected ?: State.Disconnected() }
```
twyatt, line 224: This is clever enough that probably warrants a code comment.

```suggestion
        // Set state to `Disconnected` only if not already `Disconnected`.
        _state.update { previous -> previous as? State.Disconnected ?: State.Disconnected() }
```

twyatt (approved):

sdonn3 (approved):

Issue #164, opened 2021-09-24
NullPointerException when interrupting scanning by turning off bluetooth


CTMBNara:
Steps to reproduce:
- Start scanning
- Turn off bluetooth via android UI
- NullPointerException will be thrown

```
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: some.project.android, PID: 7248
    java.lang.NullPointerException: Attempt to invoke virtual method 'void android.bluetooth.le.BluetoothLeScanner.stopScan(android.bluetooth.le.ScanCallback)' on a null object reference
        at com.juul.kable.AndroidScanner$advertisements$1$2.invoke(Scanner.kt:71)
        at com.juul.kable.AndroidScanner$advertisements$1$2.invoke(Scanner.kt:70)
        at kotlinx.coroutines.channels.ProduceKt.awaitClose(Produce.kt:58)
        at kotlinx.coroutines.channels.ProduceKt$awaitClose$1.invokeSuspend(Unknown Source:13)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:104)
        at androidx.compose.ui.platform.AndroidUiDispatcher.performTrampolineDispatch(AndroidUiDispatcher.android.kt:81)
        at androidx.compose.ui.platform.AndroidUiDispatcher.access$performTrampolineDispatch(AndroidUiDispatcher.android.kt:41)
        at androidx.compose.ui.platform.AndroidUiDispatcher$dispatchCallback$1.run(AndroidUiDispatcher.android.kt:57)
        at android.os.Handler.handleCallback(Handler.java:938)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:223)
        at android.app.ActivityThread.main(ActivityThread.java:7656)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)
```

sdonn3:
As we are no longer attempting to retrieve the scanner in scenarios when BLE is disabled, https://github.com/JuulLabs/kable/pull/187 should solve this issue

PR #187, opened 2021-11-01
Improve Advertising Robustness

sdonn3:
Previously, we were crashing at times when a user toggled Bluetooth during a scan. This should allow that to be cancelled gracefully.

https://jira.juullabs.com/browse/ESC-1869
https://jira.juullabs.com/browse/ESC-1858

### DIFF
--- a/core/src/androidMain/kotlin/Scanner.kt
+++ b/core/src/androidMain/kotlin/Scanner.kt
@@ -9,12 +9,15 @@ import android.os.ParcelUuid
 import com.benasher44.uuid.Uuid
 import com.juul.kable.logs.Logger
 import com.juul.kable.logs.Logging
+import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.channels.awaitClose
 import kotlinx.coroutines.channels.onFailure
 import kotlinx.coroutines.channels.trySendBlocking
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.callbackFlow
+import kotlinx.coroutines.flow.flowOn
+import kotlinx.coroutines.withContext
 
 public class ScanFailedException internal constructor(
     public val errorCode: Int,
@@ -32,7 +35,7 @@ public class AndroidScanner internal constructor(
         ?: error("Bluetooth not supported")
 
     public override val advertisements: Flow<Advertisement> = callbackFlow {
-        check(bluetoothAdapter.isEnabled) { "Bluetooth is disabled" }
+        val scanner = checkNotNull(bluetoothAdapter.bluetoothLeScanner) { "Bluetooth disabled." }
 
         val callback = object : ScanCallback() {
             override fun onScanResult(callbackType: Int, result: ScanResult) {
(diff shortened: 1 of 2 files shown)


core/src/androidMain/kotlin/Scanner.kt
```diff
@@ -53,22 +56,34 @@ public class AndroidScanner internal constructor(
...
-            bluetoothAdapter.bluetoothLeScanner.stopScan(callback)
+            logger.info {
+                message = when (filterServices) {
+                    null -> "Stopping scan with no service filter."
+                    else -> "Stopping scan for services ${filterServices.joinToString()}."
+                }
+            }
+            try {
+                scanner.stopScan(callback)
+            } catch (e: IllegalStateException) {
```
twyatt, line 84: Can we add a comment as to why we need to catch this exception here.
I seem to remember it being due to some Android weirdness, so it would be good to capture that and explain it here a little.

twyatt (approved):

core/src/androidMain/kotlin/Scanner.kt
cedrickcooke: tl;dr is that we can't check BLE state, only full Bluetooth state, so we need to try to stop the scan blindly. If it fails, it _probably_ just means that BLE is off.

core/src/androidMain/kotlin/Scanner.kt
```diff
@@ -53,22 +56,34 @@ public class AndroidScanner internal constructor(
...
+                logger.warn { message = "Failed to stop scan." }
```
cedrickcooke, line 85: Does Kable's logging allow plugging in the exception?

cedrickcooke (approved):

core/src/androidMain/kotlin/Scanner.kt
twyatt: I'm pretty sure it does.

twyatt:
Does this close #164?

sdonn3:
I believe it does

twyatt:
#164 has a stacktrace w/ `NullPointerException` though, while this changeset catches `IllegalStateException`.
Is it a different issue, or are we catching the wrong exception?

core/src/androidMain/kotlin/Scanner.kt
twyatt: Why can't we check BLE state here?

sdonn3:
@twyatt That's due to some Android BLE weirdness. They don't expose some internal BLE states, so we have to make guesses and attempt to shut down scanning, which may fail with `IllegalStateException` - That being said we shouldn't see the `NullPointerException` anymore

core/src/androidMain/kotlin/Scanner.kt
cedrickcooke: We can't check it anywhere. `BluetoothAdapter` internally keeps track of state, but filters it out before exposing it to apps.

From the Android sources:

```java
    /**
     * Get the current state of the local Bluetooth adapter.
     * <p>Possible return values are
     * {@link #STATE_OFF},
     * {@link #STATE_TURNING_ON},
     * {@link #STATE_ON},
     * {@link #STATE_TURNING_OFF}.
     *
     * @return current state of Bluetooth adapter
     */
    @RequiresPermission(Manifest.permission.BLUETOOTH)
    @AdapterState
    public int getState() {
        int state = getStateInternal();

        // Consider all internal states as OFF
        if (state == BluetoothAdapter.STATE_BLE_ON || state == BluetoothAdapter.STATE_BLE_TURNING_ON
                || state == BluetoothAdapter.STATE_BLE_TURNING_OFF) {
            if (VDBG) {
                Log.d(TAG, "Consider " + BluetoothAdapter.nameForState(state) + " state as OFF");
            }
            state = BluetoothAdapter.STATE_OFF;
        }
        if (VDBG) {
            Log.d(TAG, "" + hashCode() + ": getState(). Returning " + BluetoothAdapter.nameForState(
                    state));
        }
        return state;
    }
```

twyatt:
> @twyatt That's due to some Android BLE weirdness. They don't expose some internal BLE states (BLE-only and its intermediate states), so we have to make guesses and attempt to shut down scanning, which, if we guess incorrectly, may fail with `IllegalStateException`

Sounds good, I was kind of hoping a bit of this clarity ended up in the code comment for the `try`-`catch`, but I should've blocked the PR if I really wanted that done; comment is good enough for now.

> That being said we shouldn't see the `NullPointerException` anymore

Then, can you close out #164 with a comment explaining how this PR resolves that issue? Thank you!

Issue #183, opened 2021-10-26
Manufacturer Data Scan Filtering on Android


mtrewartha:
In our app, we make use of Android's `ScanFilter` to filter advertisements based on manufacturer data like so:

```kotlin
ScanFilter.Builder().setManufacturerData(..., ...).build()
```

but Kable's current `Scanner` abstraction only supports the idea of filtering based on service UUIDs. It would be great for Android devs to have support for this in the `Scanner { ... }` DSL, but iOS doesn't support the notion of filtering on manufacturer data at all. To keep things clean and platform-agnostic in the `Scanner` DSL, we could employ a similar approach to how `ScanSettings` are supported for Android only. It would be super simple and I'm happy to create a PR for it.  Thoughts?

PR #191, opened 2021-11-17
Add Android-only manufacturer data scan filters

mtrewartha:
This PR address #183.


core/src/androidMain/kotlin/Scanner.kt
```diff
@@ -60,22 +61,36 @@ public class AndroidScanner internal constructor(
...
+                        it.manufacturerData,
+                        it.manufacturerDataMask
+                    )
+                    .build()
+            }
+        val serviceScanFilters = filterServices
             ?.map { ScanFilter.Builder().setServiceUuid(ParcelUuid(it)).build() }
-            ?.toList()
+        val scanFilters = (manufacturerDataScanFilters ?: emptyList())
+            .plus(serviceScanFilters ?: emptyList())
```
twyatt, line 77: nitpick

```suggestion
        val scanFilters = manufacturerDataScanFilters.orEmpty() + serviceScanFilters.orEmpty()
```

core/src/androidMain/kotlin/ManufacturerDataFilter.kt
```diff
@@ -0,0 +1,7 @@
+package com.juul.kable
+
+public class ManufacturerDataFilter(
+    public val manufacturerId: Int,
+    public val manufacturerData: ByteArray,
+    public val manufacturerDataMask: ByteArray?
```
twyatt, line 6: Being that this is in the `ManufacturerDataFilter`, having the properties prefixed with `manufacturer` feels a bit redundant to me.

```suggestion
    public val id: Int,
    public val data: ByteArray,
    public val dataMask: ByteArray?
```

twyatt (commented):
Did a quick review pass (I'll review again soon).

In the mean time, it seems it would make sense to create a `sealed class` of various Android filter types. For example:

```kotlin
sealed class Filter {
    data class Service(uuid: Uuid) : Filter()

    data class ManufacturerData(
        public val id: Int,
        public val data: ByteArray,
        public val dataMask: ByteArray?,
    ) : Filter()
}
```

We can start with the above types, but it makes it easy to add additional types in the future, as the [`ScanFilter` documentation](https://developer.android.com/reference/android/bluetooth/le/ScanFilter) describes support for:

Current filtering on the following fields are supported:
- Service UUIDs which identify the bluetooth gatt services running on the device.
- Name of remote Bluetooth LE device.
- Mac address of the remote device.
- Service data which is the data associated with a service.
- Manufacturer specific data which is the data associated with a particular manufacturer.

We could then drop the `filterServices` parameter of `AndroidScanner` and replace it with a `filters: List<Filter>?` parameter.

I believe it would both simplify the call-site and the internal mapping of the various filter types (to the corresponding `ScanFilter.build` operations).

Thoughts?

core/src/androidMain/kotlin/ManufacturerDataFilter.kt
mtrewartha: Agreed.

core/src/androidMain/kotlin/Scanner.kt
mtrewartha: Ah, I didn't know that `orEmpty()` function existed! Agreed, I like it here.

mtrewartha:
Oh yeah, I like that idea. I like the extensibility and I agree, it cleans some stuff up nicely. I'll knock it out.

mtrewartha:
@twyatt There's one thing that I keep coming back to... In the `Scanner` DSL created by `ScannerBuilder`, `services` are not really named as if they're a filter. However, service UUIDs as filters conceptually make sense to me and of course, Android and JS treat them as such. On Apple, they're simply passed to `CBCentralManager`. If we're willing to use service UUIDs as filters in kable, then we can harmonize the DSL for Android, Apple, and JS platforms by lifting your `Filter` class up into the common code. Maybe this is what you originally had in mind with your comment before, or maybe you were just thinking of using the `Filter` in the Android source set. Either way, I took some time to explore the idea and like how it turned out. I'll commit and push just so you can see it all laid out in code. I'm happy to chat and rework if need be.

twyatt:
> Maybe this is what you originally had in mind with your comment before

I wasn't actually thinking of it like that, but looking at it now, wish I had thought of it; I really like it! 😄 

I'll review this PR soon (hopefully in the next few days), the overall design/direction looks great though.

mtrewartha:
Sounds good! No rush here- we're using a fork with these changes for now.

core/src/androidMain/kotlin/Filter.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import com.benasher44.uuid.Uuid
+
+public actual sealed class Filter {
+
+    public class ManufacturerData(
+        public val id: Int,
+        public val data: ByteArray,
+        public val dataMask: ByteArray?
+    ) : Filter()
```
twyatt, line 11: Documentation copied from Android docs with slight modifications.

```suggestion
    /**
     * https://developer.android.com/reference/android/bluetooth/le/ScanFilter.Builder#setManufacturerData(int,%20byte[],%20byte[])
     */
    public class ManufacturerData(
        /** A negative [id] is considered as invalid id. */
        public val id: Int,

        public val data: ByteArray,

        /**
         * For any bit in the mask, set it the 1 if it needs to match the one in [data], otherwise set it to 0.
         * [dataMask] must have the same length as [data].
         */
        public val dataMask: ByteArray?
    ) : Filter()
```

core/src/androidMain/kotlin/ScannerBuilder.kt
```diff
@@ -1,12 +1,11 @@
 package com.juul.kable
 
 import android.bluetooth.le.ScanSettings
-import com.benasher44.uuid.Uuid
 import com.juul.kable.logs.Logging
 import com.juul.kable.logs.LoggingBuilder
 
 public actual class ScannerBuilder {
-    public actual var services: List<Uuid>? = null
+    public actual var filters: List<Filter>? = null
```
twyatt, line 8: We can give people a little bit of time to migrate to the new `filters` property with a deprecation cycle.

```suggestion
    @Deprecated(message = "Replaced by filters property")
    public var services: List<Uuid>?
        set(value) {
            filters = value?.map { Filter.Service(it) }
        }
        get() = filters?.filterIsInstance<Filter.Service>()?.map { it.uuid }

    public actual var filters: List<Filter>? = null
```

core/src/appleMain/kotlin/ScannerBuilder.kt
```diff
@@ -1,11 +1,10 @@
 package com.juul.kable
 
-import com.benasher44.uuid.Uuid
 import com.juul.kable.logs.Logging
 import com.juul.kable.logs.LoggingBuilder
 
 public actual class ScannerBuilder {
-    public actual var services: List<Uuid>? = null
+    public actual var filters: List<Filter>? = null
```
twyatt, line 7: Same as above (a deprecation would be good for the old property).

```suggestion
    @Deprecated(message = "Replaced by filters property")
    public var services: List<Uuid>?
        set(value) {
            filters = value?.map { Filter.Service(it) }
        }
        get() = filters?.filterIsInstance<Filter.Service>()?.map { it.uuid }

    public actual var filters: List<Filter>? = null
```

core/src/jsMain/kotlin/ScannerBuilder.kt
```diff
@@ -1,20 +1,21 @@
 package com.juul.kable
 
-import com.benasher44.uuid.Uuid
 import com.juul.kable.logs.Logging
 import com.juul.kable.logs.LoggingBuilder
 
 public actual class ScannerBuilder {
-    public actual var services: List<Uuid>? = null
+    public actual var filters: List<Filter>? = null
```
twyatt, line 7: Lastly, here too.

```suggestion
    @Deprecated(message = "Replaced by filters property")
    public var services: List<Uuid>?
        set(value) {
            filters = value?.map { Filter.Service(it) }
        }
        get() = filters?.filterIsInstance<Filter.Service>()?.map { it.uuid }

    public actual var filters: List<Filter>? = null
```

core/src/jsMain/kotlin/ScannerBuilder.kt
```diff
@@ -1,20 +1,21 @@
...
     private var logging: Logging = Logging()
 
     public actual fun logging(init: LoggingBuilder) {
         logging = Logging().apply(init)
     }
 
     internal actual fun build(): Scanner {
-        val filters = services
+        val filters = filters
             ?.map { it.toString() }
```
twyatt, line 16: Need to drop the string conversion here now that we have `Filter` objects and the next chain operator will do a filter on the type (and we don't want it to already all be strings).

```suggestion
```

core/src/jsMain/kotlin/ScannerBuilder.kt
```diff
@@ -1,20 +1,21 @@
...
+    public actual var filters: List<Filter>? = null
     private var logging: Logging = Logging()
 
     public actual fun logging(init: LoggingBuilder) {
         logging = Logging().apply(init)
     }
 
     internal actual fun build(): Scanner {
-        val filters = services
+        val filters = filters
```
twyatt, line 15: Let's rename the local variable to `scannerFilters` to remove confusion.

```suggestion
        val scannerFilters = filters
```

_You'll also need to edit line 23 below._

twyatt (commented):
Awesome work on this, just some minor tweaks / deprecations to add, then it is ready. 💯

twyatt:
Oh, could you also update the documentation in the project's `README.md` re: scanning.

I _think_ this is the only example that needs to be updated:

Screen Shot 2021-11-24 at 12 44 10 AM

mtrewartha:
Getting things up-to-date here today and hoping to polish up the final changes for you guys...

core/src/jsMain/kotlin/ScannerBuilder.kt
mtrewartha: Oh, good catch! That would've been an annoying silent failure.

core/src/jsMain/kotlin/ScannerBuilder.kt
mtrewartha: Good call, it shadows the instance `val`.

mtrewartha:
I think this is ready for another quick look whenever you get a chance, @twyatt and @QuantumRand. All good suggestions, thanks for the solid reviews 👍🏻 Do either of you know what's up with the macOS CI failure?

@twyatt I updated the docs and think they're reasonable, but obviously feel free to tweak however you see fit. I did notice a weird line break rendering issue, but I think the markdown itself is fine and there's something going on with GitHub's rendering (single line breaks seem to be rendering as line breaks, not as spaces like they normally would).

twyatt:
Thanks @mtrewartha! I'll take a look soon.

The MacOS CI failure is due to a typo on my part (missing trailing `\` at the end of line 48 of `ci.yml` - you can either add the missing character or I'll put a PR up soon to fix that).

mtrewartha:
Nice, that fixed it 👍🏻

core/src/androidMain/kotlin/Filter.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import com.benasher44.uuid.Uuid
+
+public actual sealed class Filter {
+
+    /**
+     * https://developer.android.com/reference/android/bluetooth/le/ScanFilter.Builder#setManufacturerData(int,%20byte[],%20byte[])
+     */
+    public class ManufacturerData(
+        /** A negative [id] is considered as invalid id. */
```
davidtaylor-juul, line 11: Nit
```suggestion
        /** A negative [id] is considered an invalid id. */
```

core/src/androidMain/kotlin/Filter.kt
```diff
@@ -0,0 +1,26 @@
...
+        public val id: Int,
+
+        public val data: ByteArray,
+
+        /**
+         * For any bit in the mask, set it the 1 if it needs to match the one in [data], otherwise set it to 0.
```
davidtaylor-juul, line 17: Nit:
```suggestion
         * For any bit in the mask, set it to 1 if it needs to match the corresponding bit in [data], otherwise set it to 0.
```

davidtaylor-juul (approved):
Terrific work, thanks @mtrewartha 🚀

mtrewartha:
Sure thing! Appreciate the quick feedback!

twyatt:
Thanks again @mtrewartha for the contribution. It will be included in the next release.

PR #193, opened 2021-11-22
Optimize observation handling

### LINKED ISSUE #190: Should not stop notifications/indications if not previously started
For `Peripheral.observe`, when subscribers goes from `1` to `0`, then underlying notifications/indications are stopped. Although this will also occur even if starting notifications/indications previously failed, or if connection drops. In both cases, stopping notifications/indications should not be attempted (the underlying BLE system will automatically clear notifications/indications).

Because these I/O operations are occurring when they shouldn't (aren't needed), the following exception is being thrown from the `observe` flow:

```
Caused by com.juul.kable.GattRequestRejectedException
       at com.juul.kable.PeripheralKt.setCharacteristicNotification(PeripheralKt.java:471)
       at com.juul.kable.PeripheralKt.access$setCharacteristicNotification(PeripheralKt.java:1)
       at com.juul.kable.AndroidPeripheral.stopObservation$core_release(AndroidPeripheral.java:391)
       at com.juul.kable.AndroidPeripheral$stopObservation$1.invokeSuspend(AndroidPeripheral.java:12)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(BaseContinuationImpl.java:33)
       at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.java:104)
       at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.java:571)
       at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.java:750)
       at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.java:678)
       at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.java:665)
```

twyatt:
Observation handling has been reworked/optimized and moved to common code.

- `Peripheral.state` promoted to `StateFlow` (from `Flow`)
- ~Replaced usage of [`stately-isolate`](https://github.com/touchlab/Stately#stately-isolate) with Tuulbox's [synchronized map](https://github.com/JuulLabs/tuulbox#synchronizedmap)~
  - We still need stately until we #237
- `ConnectionLostException` and `NotReadyException` were changed to extend `NotConnectedException` (which extends `IOException`) — they previously directly extended `IOException`
- Connection related failures during observation spin up/down are now suppressed (closes #190).


core/src/commonMain/kotlin/Observers.kt
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    suspend fun onConnected() {
+        observations.entries.forEach { (characteristic, observation) ->
+            try {
+                observation.onConnected()
+            } catch (cancellation: CancellationException) {
+                throw cancellation
+            } catch (e: Exception) {
+                characteristicChanges.emit(ObservationEvent.Error(characteristic, e))
```
davidtaylor-juul, line 70: It's unclear here what the behavior is regarding the `CancellationException`, if this is as intended can we please have a comment explaining what is going on?

davidtaylor-juul (approved):
Awesome refactor, great work.

core/src/commonMain/kotlin/Observers.kt
```diff
@@ -0,0 +1,80 @@
...
+ * @param peripheral to perform notification actions against to enable/disable the observations.
+ */
+internal class Observers<T>(
+    private val peripheral: Peripheral,
+    private val logging: Logging,
+    extraBufferCapacity: Int = 0,
+) {
+
+    val characteristicChanges = MutableSharedFlow<ObservationEvent<T>>(extraBufferCapacity = extraBufferCapacity)
+    private val observations = synchronizedMapOf<Characteristic, Observation>()
```
cedrickcooke, line 43: Just checking that this works on Native. `SynchronizedMap` ensures that there isn't concurrent access, but unlike `stately-isolate` this doesn't enforce that all reads/writes take place on the same thread. Is Kable already using the new native memory model?

cedrickcooke (approved):

core/src/commonMain/kotlin/Observers.kt
twyatt: It is not using the new memory model, I _think_ I tested MacOS w/ the change, but I'm not totally confident.
Thanks for the callout, I'll test before merging.

twyatt:
Writing unit tests to validate functionality, then will merge.

core/src/commonMain/kotlin/Observers.kt
twyatt: Tested via SensorTag sample using this branch. It worked on macOS! Looking through the code I'm actually a bit surprised it does work. I expected the `Observation` objects (within the `Observers.observations` `SynchronizedMap`) to be frozen — but it seems that an observation worked correctly. 🤷

core/src/commonMain/kotlin/Observers.kt
twyatt: Is the following code comment sufficient / what you had in mind?

https://github.com/JuulLabs/kable/blob/b8fd68a8266f384e859cd8bbbdb14f65619b4ee7/core/src/commonMain/kotlin/Observers.kt#L63-L74

twyatt: There is a whole discussion around being able to catch exceptions except for `CancellationException`s: https://github.com/Kotlin/kotlinx.coroutines/issues/1814

Basically, it is unfortunate that we can't use `runCatching` to make this read much more clearly.

Other libraries have introduced their own variants of `runCatching` (e.g. [`runSuspendCatching`](https://github.com/michaelbull/kotlin-result/pull/67)) to make code more readable for this kind of control flow.

core/src/commonMain/kotlin/Observation.kt
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    suspend fun onCompletion(action: OnSubscriptionAction) = mutex.withLock {
+        subscribers -= action
+        disableObservationIfNeeded()
+    }
+
+    suspend fun onConnected() = mutex.withLock {
+        // `force = true` is used in case a disconnected state (to reset the observation state) was missed.
+        enableObservationIfNeeded(force = true)
```
davidtaylor-juul, line 53: Interesting, am curious what case requires this. Is there a specific unit test that replicates the asymmetric observations enable/disable that fails if `force == false`?

core/src/commonTest/kotlin/ObservationTest.kt
```diff
@@ -0,0 +1,407 @@
...
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+
+private fun generateCharacteristic() = characteristicOf(
+    service = uuid4().toString(),
+    characteristic = uuid4().toString(),
+)
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class ObservationTest {
```
davidtaylor-juul, line 27: 🚀

davidtaylor-juul (approved):

core/src/commonMain/kotlin/Observers.kt
davidtaylor-juul: The bit that is ambiguous (to me at least!) is whether a `CancellationException` is thrown or emitted by this flow.

core/src/commonMain/kotlin/Observation.kt
twyatt: I think the function name is a bit misleading, and isn't helpful to understand why.

I'll try to explain the reasoning and maybe you'll have an idea of better naming or better approach:

Essentially, we need to start observations (via platform specific calls) when all of the following conditions are met:
- we are connected
- subscribers has gone from `0` to `1`

Typically this is triggered in `enableObservationIfNeeded` via:

https://github.com/JuulLabs/kable/blob/b8fd68a8266f384e859cd8bbbdb14f65619b4ee7/core/src/commonMain/kotlin/Observation.kt#L67

| Condition | Logical check |
|-|-|
| we are connected | `isConnected` |
| subscribers has gone from `0` to `1` | `!isObservationEnabled.value && hasSubscribers` |

But...and this is where `force` comes in: If we lose connection, then on reconnect, we will **always** want to start observation if we `hasSubscribers` (regardless if we've already started previously, i.e. even if `isObservationEnabled.value` is true). This is because when the connection drops, all platforms clear the observations for us. There isn't a way to know that has happened other than monitoring for disconnected events, hence the `onConnectionLost` function — but that isn't 100% reliable as it is coming from a conflated channel. So, the approach is to say that if `onConnected` is called then we are certain this is a _new_ connection, so it is safe to always start up the observations again (i.e. `force`).

It might be slightly clearer to re-write the `if` statement as?:

```
if ((force || !isObservationEnabled.value) && isConnected && hasSubscribers) {
```

Showing that `force` is overriding the `!isObservationEnabled.value` check.

twyatt: In hindsight, maybe checking that `subscribers.size == 1` would be better than `!isObservationEnabled.value && hasSubscribers`?

For some reason, the `isObservationEnabled` boolean felt safer? But, looking at it now, it just seems to be adding unnecessary complexity.

twyatt: > Interesting, am curious what case requires this. Is there a specific unit test that replicates the asymmetric observations enable/disable that fails if `force == false`?

To answer your initial comment: any test that starts observations on connect (when we already have subscribers) is essentially utilizing `force = true`, and would fail if `force = false`.

core/src/commonMain/kotlin/Observers.kt
twyatt: Oh, my understanding is that any suspension point can potentially throw `CancellationException` if the Coroutine it is running in has been cancelled. But to be honest, I'm not 100% confident in that. Like, `ensureActive` function must exist for a reason, so maybe cancellation would have to be explicitly checked for to throw that exception.

Since cancellation is cooperative (and I'm not calling `ensureActive` anywhere) then `CancellationException` may never be thrown here.

I'll look into exactly when `CancellationException` is thrown and follow up.

davidtaylor-juul: Looking at this again in-situ rather than in the PR diff it makes sense. I think I had a bit of a brain fade there when I first read it but looking at it again now it's fine. I gotta reduce the blood volume in my caffeine system...

twyatt: > I'll look into exactly when `CancellationException` is thrown and follow up.

<details>
<summary>suspend function cancellation (or lack thereof)</summary>

```kotlin
var i = 0

fun main() = runBlocking<Unit> {
    val job = launch(Dispatchers.IO) {
        try {
            while (true) test()
        } catch (cancellation: CancellationException) {
            println("caught cancellation")
            throw cancellation
        }
    }

    delay(5_000L)
    println("cancel")
    job.cancel()
    println("done")
}

private suspend fun test() {
    println(i++)
    Thread.sleep(1_000L)
}
```

Output:

```
0
1
2
3
4
cancel
done
5
6
7
8
9
..
```
</details>

<details>
<summary>withContext cancellation (works)</summary>

```kotlin
var i = 0

fun main() = runBlocking<Unit> {
    val job = launch(Dispatchers.IO) {
        try {
            while (true) test()
        } catch (cancellation: CancellationException) {
            println("caught cancellation")
            throw cancellation
        }
    }

    delay(5_000L)
    println("cancel")
    job.cancel()
    println("done")
}

private suspend fun test() = withContext(Dispatchers.IO) {
    println(i++)
    Thread.sleep(1_000L)
}
```

Output:

```kotlin
0
1
2
3
4
cancel
done
caught cancellation

Process finished with exit code 0
```
</details>

In summary: `suspend` functions on their own do **NOT** automatically check for cancellation, _but_ `withContext` (and some other Coroutines functions, e.g. `ensureActive`) **do** check for cancellation (and throw `CancellationException` if Coroutine context is cancelled).

Deep down in the callstack of `observation.onConnected()`, `withContext` is used, so catching (and re-throwing `CancellationException`) does make sense here.

core/src/commonMain/kotlin/Observers.kt
twyatt: > Tested via SensorTag sample using this branch.
> 
> It worked on macOS!

Doh, I spoke too soon. 😢 

<details>
<summary>Crash</summary>

```
Uncaught Kotlin exception: kotlin.native.concurrent.InvalidMutabilityException: mutation attempt of frozen kotlin.Array@ed80c308
    at 0   sensortag.kexe                      0x000000010b8901c1 kfun:kotlin.Throwable#<init>(kotlin.String?){} + 97 (/Users/teamcity1/teamcity_work/6326934d18cfe24e/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Throwable.kt:24:56)
    at 1   sensortag.kexe                      0x000000010b889e2d kfun:kotlin.Exception#<init>(kotlin.String?){} + 93 (/Users/teamcity1/teamcity_work/6326934d18cfe24e/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Exceptions.kt:23:58)
    at 2   sensortag.kexe                      0x000000010b889fcd kfun:kotlin.RuntimeException#<init>(kotlin.String?){} + 93 (/Users/teamcity1/teamcity_work/6326934d18cfe24e/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/Exceptions.kt:34:58)
    at 3   sensortag.kexe                      0x000000010b8bac5d kfun:kotlin.native.concurrent.InvalidMutabilityException#<init>(kotlin.String){} + 93 (/Users/teamcity1/teamcity_work/6326934d18cfe24e/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/native/concurrent/Freezing.kt:24:85)
    at 4   sensortag.kexe                      0x000000010b8bc2bf ThrowInvalidMutabilityException + 431 (/Users/teamcity1/teamcity_work/6326934d18cfe24e/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/native/concurrent/Internal.kt:109:11)
    at 5   sensortag.kexe                      0x000000010ba06a69 Kotlin_Array_copyImpl + 153
    at 6   sensortag.kexe                      0x000000010b882135 kfun:kotlin.collections#copyInto__at__kotlin.Array<out|0:0>(kotlin.Array<0:0>;kotlin.Int;kotlin.Int;kotlin.Int){0§<kotlin.Any?>}kotlin.Array<0:0> + 277 (/Users/teamcity1/teamcity_work/6326934d18cfe24e/kotlin/kotlin-native/runtime/src/main/kotlin/generated/_ArraysNative.kt:1023:2)
    at 7   sensortag.kexe                      0x000000010b89d608 kfun:kotlin.collections.ArrayList.insertAtInternal#internal + 536 (/Users/teamcity1/teamcity_work/6326934d18cfe24e/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/collections/ArrayList.kt:224:15)
    at 8   sensortag.kexe                      0x000000010b89d934 kfun:kotlin.collections.ArrayList.addAtInternal#internal + 644 (/Users/teamcity1/teamcity_work/6326934d18cfe24e/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/collections/ArrayList.kt:234:13)
    at 9   sensortag.kexe                      0x000000010b89aa18 kfun:kotlin.collections.ArrayList#add(1:0){}kotlin.Boolean + 152 (/Users/teamcity1/teamcity_work/6326934d18cfe24e/kotlin/kotlin-native/runtime/src/main/kotlin/kotlin/collections/ArrayList.kt:81:16)
    at 10  sensortag.kexe                      0x000000010bb62c35 kfun:com.juul.kable.Observation.$onSubscriptionCOROUTINE$58#invokeSuspend(kotlin.Result<kotlin.Any?>){}kotlin.Any? + 2261 (/Users/travis/Projects/kable/core/src/commonMain/kotlin/Observation.kt:37:20)
    at 11  sensortag.kexe                      0x000000010bb63d61 kfun:com.juul.kable.Observation#onSubscription(kotlin.coroutines.SuspendFunction0<kotlin.Unit>){} + 353 (/Users/travis/Projects/kable/core/src/commonMain/kotlin/Observation.kt:36:13)
    at 12  sensortag.kexe                      0x000000010bb6c3ba kfun:com.juul.kable.Observers.$acquire$lambda-0COROUTINE$68.invokeSuspend#internal + 554 (/Users/travis/Projects/kable/core/src/commonMain/kotlin/Observers.kt:57:43)
    at 13  sensortag.kexe                      0x000000010bb6c8d6 kfun:com.juul.kable.Observers.$acquire$lambda-0COROUTINE$68.invoke#internal + 310 (/Users/travis/Projects/kable/core/src/commonMain/kotlin/Observers.kt:57:29)
    at 14  sensortag.kexe                      0x000000010baa0521 kfun:kotlinx.coroutines.flow.SubscribedFlowCollector.$onSubscriptionCOROUTINE$151#invokeSuspend(kotlin.Result<kotlin.Any?>){}kotlin.Any? + 1329 (/opt/buildAgent/work/44ec6e850d5c63f0/kotlinx-coroutines-core/common/src/flow/operators/Share.kt:419:27)
    at 15  sensortag.kexe                      0x000000010baa0ae4 kfun:kotlinx.coroutines.flow.SubscribedFlowCollector#onSubscription(){} + 276 (/opt/buildAgent/work/44ec6e850d5c63f0/kotlinx-coroutines-core/common/src/flow/operators/Share.kt:416:13)
    at 16  sensortag.kexe                      0x000000010ba7fb60 kfun:kotlinx.coroutines.flow.SharedFlowImpl.$collectCOROUTINE$74#invokeSuspend(kotlin.Result<kotlin.Any?>){}kotlin.Any? + 3760 (/opt/buildAgent/work/44ec6e850d5c63f0/kotlinx-coroutines-core/common/src/flow/SharedFlow.kt:361:65)
    at 17  sensortag.kexe                      0x000000010ba7fea1 kfun:kotlinx.coroutines.flow.SharedFlowImpl#collect(kotlinx.coroutines.flow.FlowCollector<1:0>){}kotlin.Nothing + 353 (/opt/buildAgent/work/44ec6e850d5c63f0/kotlinx-coroutines-core/common/src/flow/SharedFlow.kt:358:22)
    at 18  sensortag.kexe                      0x000000010ba9f7f1 kfun:kotlinx.coroutines.flow.SubscribedSharedFlow.$collectCOROUTINE$150.invokeSuspend#internal + 785 (/opt/buildAgent/work/44ec6e850d5c63f0/kotlinx-coroutines-core/common/src/flow/operators/Share.kt:409:20)
    at 19  sensortag.kexe                      0x000000010ba9fb51 kfun:kotlinx.coroutines.flow.SubscribedSharedFlow.collect#internal + 353 (/opt/buildAgent/work/44ec6e850d5c63f0/kotlinx-coroutines-core/common/src/flow/operators/Share.kt:408:22)
    at 20  sensortag.kexe                      0x000000010bb6d939 kfun:com.juul.kable.Observers.<no name provided>_1.$collect_2COROUTINE$63.invokeSuspend#internal + 809 (/Users/travis/Projects/kable/core/src/commonMain/kotlin/Observers.kt:76:819)
    at 21  sensortag.kexe                      0x000000010bb6dc31 kfun:com.juul.kable.Observers.<no name provided>_1.collect_2#internal + 353 (/Users/travis/Projects/kable/core/src/commonMain/kotlin/Observers.kt:76:819)
    at 22  sensortag.kexe                      0x000000010bb6eba7 kfun:com.juul.kable.Observers.<no name provided>_1.$collect_2COROUTINE$65.invokeSuspend#internal + 791 (/Users/travis/Projects/kable/core/src/commonMain/kotlin/Observers.kt:76:819)
    at 23  sensortag.kexe                      0x000000010bb6ee91 kfun:com.juul.kable.Observers.<no name provided>_1.collect_2#internal.64 + 353 (/Users/travis/Projects/kable/core/src/commonMain/kotlin/Observers.kt:76:819)
    at 24  sensortag.kexe                      0x000000010ba9b468 kfun:kotlinx.coroutines.flow.<no name provided>_1.$collect_2COROUTINE$115.invokeSuspend#internal + 1064 (/opt/buildAgent/work/44ec6e850d5c63f0/kotlinx-coroutines-core/common/src/flow/operators/Emitters.kt:114:49)
    at 25  sensortag.kexe                      0x000000010ba9bda1 kfun:kotlinx.coroutines.flow.<no name provided>_1.collect_2#internal.128 + 353 (/opt/buildAgent/work/44ec6e850d5c63f0/kotlinx-coroutines-core/common/src/flow/operators/Emitters.kt:114:49)
    at 26  sensortag.kexe                      0x000000010bb34d47 kfun:com.juul.kable.ApplePeripheral.<no name provided>_1.$collect_2COROUTINE$21.invokeSuspend#internal + 791 (/Users/travis/Projects/kable/core/src/appleMain/kotlin/Peripheral.kt:82:36)
    at 27  sensortag.kexe                      0x000000010bb35031 kfun:com.juul.kable.ApplePeripheral.<no name provided>_1.collect_2#internal.35 + 353 (/Users/travis/Projects/kable/core/src/appleMain/kotlin/Peripheral.kt:82:36)
    at 28  sensortag.kexe                      0x000000010b86cd57 kfun:com.juul.sensortag.SensorTag.<no name provided>_1.$collect_2COROUTINE$0.invokeSuspend#internal + 791 (/Users/travis/Projects/sensortag/app/src/commonMain/kotlin/SensorTag.kt:82:1198)
```
</details>

I'll investigate and fix before merging.

core/src/commonMain/kotlin/Observation.kt
twyatt: Significantly simplified the code. Removed the need for `force`, and figured out that having special handling on the disconnected event was unnecessary.

twyatt:
Currently investigating a crash due to object freezing on macOS. Once resolved, will merge.

core/src/commonMain/kotlin/Observers.kt
twyatt: Fixed crash by reverting back to using Stately for the subscription actions list:

https://github.com/JuulLabs/kable/blob/15bbce4e28df8688ccf80d5f44ebaa167409facc/core/src/commonMain/kotlin/Observation.kt#L28

I'm still not sure if `synchronizedMapOf` is safe for `observations` collection. I'm not sure how to test it considering you really only get a single thread in the Native Coroutines world, and that single thread is what I'm testing the Flow subscription from (which in turn accesses this map).

core/src/commonMain/kotlin/Observation.kt
```diff
@@ -0,0 +1,96 @@
...
+        get() = subscribers.isNotEmpty()
+
+    suspend fun onSubscription(action: OnSubscriptionAction) = mutex.withLock {
+        subscribers += action
+        val shouldStartObservation = !didStartObservation && hasSubscribers && isConnected
+        if (shouldStartObservation) {
+            suppressConnectionExceptions {
+                startObservation()
+                action()
+            }
```
davidtaylor-juul, line 48: Nice 😀 🚀

core/src/commonMain/kotlin/Observers.kt
twyatt: Reverted back to using Stately for observation collection handling; as it was a safer approach.
Will revisit and clean up later when we use the new memory model (#237).

core/src/commonMain/kotlin/Observers.kt
twyatt: Great article on the subject: https://maxkim.eu/things-every-kotlin-developer-should-know-about-coroutines-part-5-cancellation

PR #202, opened 2021-12-04
Remove default publication version name

twyatt:
Using `main-SNAPSHOT` as a default version name makes sense for library users who install to Maven local from the `main` branch; but for other branches it is just confusing. Better to remove the default entirely and let Maven plugin decide default version name (`unspecified` is used when not set).


sdonn3 (approved):

davidtaylor-juul (approved):
👍

PR #209, opened 2021-12-16
Add Apple Silicon support

lammertw:
Since Kotlin 1.5.30 added support for Apple Silicon it's now possible to build KMM projects on M1 without Rosetta. However it does require that all dependencies also support it explicitly.

This change should add Apple Silicon support for Kable as well. Unfortunately I haven't been able to test it locally since I'm not sure how to get the whole build working locally.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -36,7 +36,7 @@ allprojects {
     }
 }
 
-// Workaround for https://youtrack.jetbrains.com/issue/KT-49109 that allows building on Apple Silicon.
+// Workaround for https://youtrack.jetbrains.com/issue/KT-49109 that allows building on Apple Silicon
 // This should no longer be needed with Kotlin version 1.6.20
 rootProject.plugins.withType<org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootPlugin> {
     rootProject.the<org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootExtension>().nodeVersion = "16.0.0"
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -35,3 +35,7 @@ allprojects {
         }
     }
 }
+
+rootProject.plugins.withType<org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootPlugin> {
+    rootProject.the<org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootExtension>().nodeVersion = "16.0.0"
+}
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -36,7 +36,7 @@ allprojects {
     }
 }
 
-// Workaround for https://youtrack.jetbrains.com/issue/KT-49109 that allows building on Apple Silicon
+// Workaround for https://youtrack.jetbrains.com/issue/KT-49109 that allows building on Apple Silicon.
 // This should no longer be needed with Kotlin version 1.6.20
 rootProject.plugins.withType<org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootPlugin> {
     rootProject.the<org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootExtension>().nodeVersion = "16.0.0"
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -36,6 +36,8 @@ allprojects {
     }
 }
 
+// Workaround for https://youtrack.jetbrains.com/issue/KT-49109 that allows building on Apple Silicon
+// This should no longer be needed with Kotlin version 1.6.20
 rootProject.plugins.withType<org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootPlugin> {
     rootProject.the<org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootExtension>().nodeVersion = "16.0.0"
 }
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -27,6 +27,7 @@ kotlin {
     iosX64()
     iosArm32()
     iosArm64()
+    iosSimulatorArm64()
     macosX64()
 
     sourceSets {
@@ -103,6 +104,13 @@ kotlin {
         }
 
         val iosArm64Test by getting {
+            dependsOn(appleTest)
+        }
+
+        val iosSimulatorArm64Main by getting {
+            dependsOn(appleMain)
+        }
+        val iosSimulatorArm64Test by getting {
             dependsOn(appleTest)
         }
 
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -25,6 +25,7 @@ kotlin {
     }
     js().browser()
     iosX64()
+    macosArm64()
     iosArm32()
     iosArm64()
     iosSimulatorArm64()
@@ -80,6 +81,14 @@ kotlin {
         }
 
         val macosX64Test by getting {
+            dependsOn(appleTest)
+        }
+
+        val macosArm64Main by getting {
+            dependsOn(appleMain)
+        }
+
+        val macosArm64Test by getting {
             dependsOn(appleTest)
         }
 
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -4,7 +4,7 @@ android-min = "21"
 atomicfu = "0.17.1"
 coroutines = "1.6.0"
 kotlin = "1.6.10"
-tuulbox = "6.0.1"
+tuulbox = "6.1.0"
 
 [libraries]
 androidx-startup = { module = "androidx.startup:startup-runtime", version = "1.1.1" }


twyatt:
@QuantumRand you have access to an M1 machine, right?
Would you be able to test that this change works for you locally?

```
./gradlew --no-parallel -PRELEASE_SIGNING_ENABLED=false publishToMavenLocal
```

lammertw:
I do indeed have an M1. However it fails with the following (also without my change):
```
FAILURE: Build failed with an exception.

* What went wrong:
Could not determine the dependencies of task ':kotlinNodeJsSetup'.
> Could not resolve all files for configuration ':detachedConfiguration2'.
   > Could not find org.nodejs:node:14.17.0.
     Searched in the following locations:
       - https://dl.google.com/dl/android/maven2/org/nodejs/node/14.17.0/node-14.17.0.pom
       - https://repo.maven.apache.org/maven2/org/nodejs/node/14.17.0/node-14.17.0.pom
       - https://nodejs.org/dist/v14.17.0/node-v14.17.0-darwin-arm64.tar.gz
     Required by:
         project :

```

twyatt:
@lammertw I'm not sure of the cause of the dependency issue, but things are moving so quickly w/ Kotlin multiplatform, usually the first thing I try is bumping the Kotlin version to see if it has already been fixed.

Can you try changing Kotlin to `1.6.10` in:

https://github.com/JuulLabs/kable/blob/f48497cf5323dc609ae352a55875b5dc0ba60f0e/gradle/libs.versions.toml#L7

Let me know if that gets your any further. If not, then I'll look into it some more.

lammertw:
Unfortunately that did not change the result

twyatt:
It appears that `arm64` support for node wasn't added until [16.0.0](https://nodejs.org/dist/v16.0.0/) ([15.14.0](https://nodejs.org/dist/v15.14.0/) does not have `arm64`).
I'm not sure how to have the Kotlin multiplatform setup use a newer version of node.
Although this is only for the Javascript target of Kable.

I'm guessing that the arm64 targets can be compiled on a non-M1 machine, so publication of Kable for the arm64 simulator target might not be blocked at all — and it would just mean that until Kotlin upgrades its Node version used, that M1 users won't be able to publish locally, but should be able to compile all but the Javascript target, as needed.

I've published a `SNAPSHOT` of this PR (from a non-M1), @lammertw can you give it a try and see if it works for you?

```kotlin
repositories {
}

dependencies {
    implementation("com.juul.kable:core:0.10.4-issue-209-1-SNAPSHOT")
}
```

davidtaylor-juul:
> I'm not sure how to have the Kotlin multiplatform setup use a newer version of node.

It's a long shot, but I wonder if it could be enforced by including the Gradle plugin with a specified version (via `npmVersion` parameter), then somehow make sure the relevant tasks use the installed plugin version of nodejs instead of the globally installed version?

See https://plugins.gradle.org/plugin/com.github.node-gradle.node

QuantumRand:
@lammertw as a workaround until the default node version is bumped to 16.0.0, you can overwrite the version by adding the following to your root `build.gradle.kts` file:

```
rootProject.plugins.withType<org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootPlugin> {
    rootProject.the<org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootExtension>().nodeVersion = "16.0.0"
}
```

core/build.gradle.kts
```diff
@@ -27,6 +27,7 @@ kotlin {
     iosX64()
     iosArm32()
     iosArm64()
+    iosSimulatorArm64()
```
QuantumRand, line 30: I haven't tried targeting Apple silicon yet, but we probably want both `iosSimulatorArm64`, which is the Apple silicon equivalent of `iosX64`, and `macosArm64`, which is the Apple silicon target equivalent of `macosX64`.
```suggestion
    iosSimulatorArm64()
    macosArm64()
```

QuantumRand (changes requested):

lammertw:
@twyatt I've just tried with the 0.10.4-issue-209-1-SNAPSHOT version. It seems to work even though I had to change my Kotlin version from 1.5.30 to 1.6.x. `iosSimulatorArm64` should be supported by 1.5.30 as well but I'm assuming that's unrelated and the whole Kable library now required 1.6.x? Also I wasn't able to fully confirm yet since I had some compilation errors related to changes in creating a Scanner.

twyatt:
> It seems to work even though I had to change my Kotlin version from 1.5.30 to 1.6.x.

Thanks for giving that a try!

Regarding requiring Kotlin 1.6.x, I think Kotlin versions are usually compatible between minor versions, but we have noticed some incompatibilities internally with mixing libraries compiled with older or newer Kotlin than what you're using on the consuming side. We want to accommodate consumers who are using the newer versions of Kotlin (to get the latest bug fixes) so moving forward we will try to keep dependencies (Kotlin included) up-to-date in Kable.

Unfortunately, when Kotlin incpompatibilies do arise, you'll likely have to update to the same version of Kotlin that Kable is using. Hopefully in the future that won't be as much of an issue as some of the Kotlin multiplatform support stabilizes. 🤞 

> Also I wasn't able to fully confirm yet since I had some compilation errors related to changes in creating a Scanner.

When feasible, we try to slowly deprecate APIs, and I had thought that we had done that with the `Scanner`, but it is possible it was an oversight and we didn't properly deprecate the old API you were using. Apologies for that. Let me know if you need assistance in migrating to the new `Scanner` API changes.

firmwaves:
Hi guys, any update on this? When running the sample sensortag app with `./gradlew runDebugExecutableMacosX64` on an m1 mac, I end up getting a skipped task on the running task:

```
> Task :app:compileKotlinMacosX64 UP-TO-DATE
> Task :app:linkDebugExecutableMacosX64 UP-TO-DATE
> Task :app:runDebugExecutableMacosX64 SKIPPED
```

I am not sure if this is the task/plugin failing to see that even though the platform doesn't match, it can still be run with rosetta. Even `arch -x86_64 ./gradlew runDebugExecutableMacosX64` is not working. Maybe I can create a kotlin bug about that, but just curious about this ticket

firmwaves:
Ah turns out I can just directly run the executable located in `/build/bin/macosX64/debugExecutable/...`. That will hold me over for now until this ticket moves forward. Thanks guys for looking at this

twyatt:
> any update on this?

@firmwaves we're getting M1 Macs at work, but they're weeks backordered from Apple. Once we get them then I can hopefully test and get this PR merged.

burnhamd:
@twyatt were you able to test with M1 Macs?

lammertw:
I've updated this PR:
- brought it up to date with `main`
- Added `macosArm64` as suggested by @QuantumRand 
- Force it to use node 16 as mentioned by @QuantumRand

I've also created a PR to add support to your tuulbox library: https://github.com/JuulLabs/tuulbox/pull/154

Using `./gradlew --no-parallel -PRELEASE_SIGNING_ENABLED=false publishToMavenLocal` I was able to build/publish tuulbox for M1 and then using that as dependency I was able to build/publish Kable for M1.

twyatt:
> @twyatt were you able to test with M1 Macs?

Not yet, still backordered. I'm hoping we'll get them in the next few weeks.

twyatt (approved):
@lammertw took in your upstream changes to Tuulbox into [6.1.0](https://github.com/JuulLabs/tuulbox/releases/tag/6.1.0).

Just bump Tuulbox dependency to `6.1.0` in this PR and it looks ready to go to me.

https://github.com/JuulLabs/kable/blob/a68a3a8c15e96ef2c24e26b6c15d626030841c5b/gradle/libs.versions.toml#L7

lammertw:
Thanks @twyatt. I've changed the dependency to 6.1.0.

twyatt (approved):

build.gradle.kts
```diff
@@ -35,3 +35,7 @@ allprojects {
         }
     }
 }
+
+rootProject.plugins.withType<org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootPlugin> {
+    rootProject.the<org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootExtension>().nodeVersion = "16.0.0"
+}
```
QuantumRand, line 41: I would like to see a comment explaining this workaround and its resolution plan.

```suggestion

// Workaround for https://youtrack.jetbrains.com/issue/KT-49109 that allows building on Apple Silicon
// This should no longer be needed with Kotlin version 1.6.20
rootProject.plugins.withType<org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootPlugin> {
    rootProject.the<org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootExtension>().nodeVersion = "16.0.0"
}
```

QuantumRand (approved):
Looks good! I was able to test these changes on M1 hardware using https://github.com/JuulLabs/sensortag 

It only took some fiddling with the native Coroutines dependencies and everything worked smoothly after that.

Thanks!

build.gradle.kts
```diff
@@ -35,3 +35,9 @@ allprojects {
         }
     }
 }
+
+// Workaround for https://youtrack.jetbrains.com/issue/KT-49109 that allows building on Apple Silicon
```
twyatt, line 39: Change suggestion in an attempt to trigger GitHub checks to run.

```suggestion
// Workaround for https://youtrack.jetbrains.com/issue/KT-49109 that allows building on Apple Silicon.
```

twyatt:
Apparently committing suggested changes to your branch is not triggering the GitHub checks to run. 😢 
@lammertw can you make a minor comment change (e.g. revert e8592df — which just added a trailing period to the "workaround" comment) in order to trigger the GitHub checks. Thank you!

lammertw:
@twyatt sure, I just reverted that commit.

twyatt:
Bizarre, I have no idea why the GitHub actions/checks are not running.

When I get into the office, I'll build locally to make sure things are working, then just merge manually.

twyatt:
> @twyatt GitHub was having issues today. So might be related to that.

Ya, must've been that, as the "approve run" button showed up and the checks are running now. 👍

PR #235, opened 2022-01-04
Simplify service filter creation in common code

### LINKED ISSUE #230: Fix scanner filter creation in common code
With #191, the `Scanner` implementation was commonized across platforms and added some platform specific features. Unfortunately, as reported in https://github.com/JuulLabs/kable/pull/209#issuecomment-998975092, there isn't a friendly deprecation for consumers that created a `Scanner` with a list of service UUIDs.

The new scan filter classes provide a nice implementation for Android consumers, but in common code it is difficult to create a service UUID filter.

twyatt:
#191 helped to commonize the scanner settings implementation, but as reported in https://github.com/JuulLabs/kable/pull/209#issuecomment-998975092, the updates in #191 don't provide a simple migration for the legacy API. The ability to construct common filters was also difficult from common code.

This PR aims to simplify filter creation (when using from common code) and provide deprecated functions for the legacy API.

Closes #230

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ Scan results can be filtered by manufacturer data using the same ID, data, and d
 ```kotlin
 val scanner = Scanner {
     filters = listOf(
-        Filter.ManufacturerData(manufacturerID = 1, data = byteArrayOf(), dataMask = byteArrayOf())
+        Filter.ManufacturerData(id = 1, data = byteArrayOf(), dataMask = byteArrayOf())
     )
 }
 ``` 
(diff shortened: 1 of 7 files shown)


twyatt:
@mtrewartha do these API changes still match your intended design?

mtrewartha:
@twyatt Oh yeah, looks I forgot about the migration part of that other PR. The aim of this PR makes sense, but it seems I can't reference the `Filter.Companion.ManufacturerData` function from my Android source set when I use this build... strange. Studio is definitely pulling in the correct version of the library and looking at your code, I'm not sure why this would happen. Are you able to reproduce?

twyatt:
Thanks for trying it, I had just done quick scratch files in the Kable project to check if it compiled. I'll do some tests pulling it into a project.

I'm not really a fan of the companion object approach I took, I couldn't think of a better way at the time, but if it proves troublesome then I'll try to come up with a better way.

I'll keep you posted.

twyatt:
@mtrewartha I had forgotten about `expect constructor`s. 🤦 

This made the change set very small against your original implementation while enabling creation of `Filter.Service` objects from common code. 💯 

If you have time, and want to confirm this still works as you expected (although I don't expect it to break anything for your usage — i.e. when using the API from Android):

```kotlin
repositories {
}

dependencies {
    implementation("com.juul.kable:core:0.10.4-issue-183-2-SNAPSHOT")
}
```

mtrewartha:
@twyatt Ah, that'll do it! Works great for me now 👍🏻

davidtaylor-juul (approved):
LGTM

cedrickcooke (approved):

PR #238, opened 2022-01-06
Make `services` direct references and match properties for I/O

### LINKED ISSUE #93: Services with the same UUIDs

### LINKED ISSUE #127: Characteristics with duplicating UUIDs are not supported
I have a device which has two different characteristics with the same UUID (8b6b2dbe-cffa-4f90-b004-b411af90d4c4):
image

One of them should be used for writing data and another for notifications.

Seems kable uses standard android method for getting characteristic and it returns the first one by UUID:
`service.getCharacteristic(BLE_DATA_STREAM_CHARACTERISTIC_DATA_AVAILABILITY)`

For my device I wrote a pure app without libraries and I loop the service characteristic and choose proper one by checking its properties:
`    private fun BluetoothGattService.getCharaWithProperty(uuid: UUID, property: Int) =
        characteristics.first { it.uuid == uuid && it.properties and property != 0 }`

This pure app works fine but if I use kable my device is not working properly.

Is there any way to choose service/characteristic from discovered but not just by UUIDs?
Is there a workaround to my problem?

twyatt:
Inspired by #128.

Previously, all I/O operations would lazily search for the first match when looking up characteristics or descriptors. This was problematic if duplicate UUIDs existed. Discovered GATT profile items (services, characteristics and descriptors) as provided via `Peripheral.services` now hold direct references to underlying platform types. This means they no longer trigger lookups for I/O operations and specific GATT profile items can be manually sought after and used.

When using `characteristicOf` (which is still a lazy lookup), the properties of the characteristic are now taken into account when searching for the underlying platform object (this will allow support for duplicate UUIDs assuming they differ by the appropriate properties). For example, when performing a read, only a characteristic with a matching UUID **and** `read` property will be considered a match.

This PR implements a "more correct" behavior for characteristic lookups (when using `characteristicOf`) but is a potentially breaking change (at runtime) for some users. If a library consumer has a peripheral with characteristics with incorrect (or missing) properties, then they may have been relying on `characteristicOf` to match the first one found with the same UUID, but this PR may make it so that when lazily looking up the characteristic, it is no longer found if the peripheral's characteristic is not setup with the appropriate properties. In this case, these users should either update the firmware on the peripheral to have correct properties, or manually traverse the `Peripheral.services` in order to perform I/O on the characteristic they were using before.

Updated documentation can be read [here](https://github.com/JuulLabs/kable/tree/twyatt/profile#io).

Closes #127
Closes #93


davidtaylor-juul (approved):
🚀

cedrickcooke (approved):

Issue #237, opened 2022-01-06
Have project use the new memory model


twyatt:
https://blog.jetbrains.com/kotlin/2021/08/try-the-new-kotlin-native-memory-manager-development-preview/

Evaluate if:
- any `atomics` can be replaces with raw types
- usages of Stately can be replaced by Tuulbox's `SynchronizedMap`
  - Stately spins up a thread on JVM target for handling isolated operations, we're also not disposing of these when done with them

To do:
- [ ] ~~Update project readme gradle instructions~~ _Had already been updated via #313_

volo-droid:
There's a new version of Stately available `2.0.0` (so far there's a release candidate `2.0.0-rc2`), which uses the new memory model, deprecates `stately-iso-collections`, and suggests to use [`stately-concurrent-collections`](https://github.com/touchlab/Stately#stately-concurrent-collections) instead. 

It would be cool if Kable migrates to that version (or alternatively removes Stately all together) otherwise it's impossible to use Stately concurrent collections together with Kable in one project because of a linking error for iOS/Mac targets:
```
e: Module "co.touchlab:stately-isolate" has a reference to symbol co.touchlab.stately.concurrency/GuardedStableRef|null[0]. Neither the module itself nor its dependencies contain such declaration.
```

twyatt:
Thanks for the info, I had planned to remove Stately entired but I was unaware of `stately-concurrent-collections`. That looks like a good choice. Super helpful that you pointed that out!
It is high up on my to do list to get these changes in.

twyatt:
@volo-droid I published a `SNAPSHOT` version for testing (from #437) that has Stately removed:

```
repositories {
    maven("https://oss.sonatype.org/content/repositories/snapshots")
}

dependencies {
    implementation("com.juul.kable:core:0.21.0-issue-237-1-SNAPSHOT")
}
```

volo-droid:
Thanks @twyatt . The project can be built with that version, Mac version crashes on consuming `scanner.advertisements` but I have to figure out if the issue is on our side or caused by the library.

PR #246, opened 2022-01-18
Prevent connection state monitor from being cancelled on disconnect

### LINKED ISSUE #226: Can't reconnect after disconnect
Hi

On iOS when I call disconnect() the state changes directly to Disconnected, if I Connect again the state changes to Connected but on write if get the com.juul.kable.NotReadyException because the peripheral is disconnected

```
Debug: (Kermit) New state Disconnected(status=null)
Debug: (Kermit) New state com.juul.kable.State.Connecting.Bluetooth@83761768
Debug: (Kermit) New state com.juul.kable.State.Connecting.Services@83761528
Debug: (Kermit) New state com.juul.kable.State.Connected@83761718```

Now I call disconnect() the state never updates, if I call connect() again it updates directly to Connected

Debug: (Kermit) New state com.juul.kable.State.Connected@83761718

Now I write to the peripheral but I get the exception because the peripheral is disconnected.

com.juul.kable.NotReadyException

twyatt:
Added a dedicated connection `CoroutineScope`, so that in-flight connection attempts can be cancelled (via `Job.cancelChildren`) without cancelling the Coroutine that pipes `CentralManager` delegate callbacks to the `Peripheral`'s connection state `Flow`.

Closes #226

### DIFF
--- a/core/src/appleMain/kotlin/Peripheral.kt
+++ b/core/src/appleMain/kotlin/Peripheral.kt
@@ -34,6 +34,7 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.CoroutineStart.LAZY
 import kotlinx.coroutines.CoroutineStart.UNDISPATCHED
 import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.NonCancellable
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.async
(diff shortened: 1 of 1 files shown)


cedrickcooke (approved):
Just checking. Should this change be mirrored to (or is it already present in) the other targets?

twyatt:
> Just checking. Should this change be mirrored to (or is it already present in) the other targets?

Good callout.

On Android and Javascript, we update the `StateFlow` directly from the callbacks/listeners.
From what I vaguely remember, this was not done on Apple due to freezing issues. So, something to revisit after #237.

tl;dr this change is not needed on the other platforms, thanks for double checking though!

sdonn3 (approved):

davidtaylor-juul (approved):

PR #254, opened 2022-02-04
Add support for handling observation exceptions

### LINKED ISSUE #188: `Peripheral.observe` error propagation
With the current design, errors related to an observation (`Peripheral.observe`), are propagated via an exception being thrown in the flow. This allows for handling of failures via flow operators such as `catch` or `retry`, but exceptions thrown in a flow are terminal events.

Having the `observe` flow terminate can be problematic, as the flow is a kind of hybrid cold/hot flow, whereas it reacts (performs actions) based on the number of subscribers. When subscribers goes from `0` to `1`, the characteristic notifications/indications will be enabled (and appropriate descriptor written to) and the converse (when subscribers goes from `1` to `0`) will perform the inverse (unwinding the notification/indication setup).

`observe` was designed to be long-living, as in: you can subscribe and it will remain active across disconnects/reconnects (automatically re-configuring any notifications/indications as needed on reconnect). This simplifies its usage considerably but having the flow remain active across some errors states (connection drop) but not others (failure to write to config descriptor) is inconsistent. Combined with the automatic handling of spinning up/down indications/notifications, makes error handling difficult and error prone. Using `retry` would keep the flow alive for downstream subscribers but would have the adverse side-effect of spinning down/up any notifications/indications (because subscribers will momentarily go from `1` to `0`, then back to `1`) — these operations can be expensive and are unnecessary.

The design isn't yet solidified but [`materialize` operator](https://github.com/Kotlin/kotlinx.coroutines/issues/2092) might provide a simple error handling mechanism. The challenge is being able to propagate failures **without** terminating the `observe` flow. It isn't clear if `materialize` would offer such functionality.

Rather than wait for `materialize` operator to be available, the `observe` flow should propagate a `sealed class` of event types (either **data** or **error**). This will allow the `observe` flow to remain active when, for example, failures to spin up notifications/indications occur. Instead of throwing the exception in the flow, a `Observation.Error` (or similarly named class) will be propagated. Down stream subscribers can dematerialize failures if desired, but the flow would remain active and only unexpected ("exceptional") `Exception`s would be thrown from an `observe` flow (i.e. the flow would only terminate via exception due to a bug in Kable).

twyatt:
Closes #188


cedrickcooke (approved):

davidtaylor-juul (approved):
🚀

Issue #251, opened 2022-01-31
Reconnecting Too Quickly Hangs on Javascript


davidtaylor-juul:
Once _Disconnected_ is received, if `connect` is called immediately it may hang forever.

A workaround found by trial and error is to delay for at least 1 second prior to calling `connect`.

davidtaylor-juul:
Closing this as we suspect it is peculiar to our devices and not necessarily an underlying issue with Web BLE.

PR #255, opened 2022-02-04
Refactor Javascript Connect/Disconnect

### LINKED ISSUE #250: Open Hangs on Javascript After Unexpected Disconnect
When a peripheral goes out of range or otherwise unexpectedly disconnects (say if rebooted while connected) it may become unable to be reconnected when back in range.

There seems to be a couple of failure modes depending on what action is taken.

Scenario a)
1. Device goes out of range
2. `Peripheral.state` emits _Disconnected_
3. We invoke `Peripheral.connect` to try to reconnect

Result: hangs forever (suspect inside `gatt.connect`). Also trying to obtain a new peripheral via `bluetooth.requestDevice` also black-holes. Only recourse is to reload the web page.

Scenario b)
1. Device goes out of range
2. `Peripheral.state` emits _Disconnected_
3. We invoke `Peripheral.disconnect` in order to "clean up" the Web BLE machinery
4. We invoke `Peripheral.connect` to try to reconnect

Result: `DOMException` is thrown with an "Bluetooth Device is no longer in range" message. This continues even when the device is back in range. A new peripheral must be obtained in order to reconnect (but in this case there is no need to reload the page).

davidtaylor-juul:
Fixes #250 

- Fixes a race condition where the connect job could be cleared without cancelling, so sometimes there would be two connect jobs running simultaneously
- Fixes issue where observation listeners are cleared from `observationListeners` without stopping notifications and removing the listeners
- Invoking `gatt` methods from the disconnect listener context is problematic. Made the disconnect process an asynchronous job so it can be launched outside the listener context and tracked in the same manner as the connect job.
- Enforce the state transitions more strictly so that connect and disconnect jobs can avoid collisions

~Also adds a workaround for #251 in the form of a delay after disconnecting. It is very curious that 1 second is not long enough to avoid the issue, but 1.5 seconds is.~ Tests are inconclusive, but this may be specific to a particular device and smells a bit too hacky so leaving this out for now.


core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -128,53 +133,105 @@ public class JsPeripheral internal constructor(
...
+        disconnectSync(cancelConnectJob = true)
+    }
+
+    private suspend fun connectSync() {
+        val job = connectJob ?: connectAsync().apply {
+            connectJob = this
+            invokeOnCompletion { connectJob = null }
+        }
+        job.await()
+    }
```
twyatt, line 174: This appears to change behavior in a way that isn't consistent with the [`connect` documentation](https://juullabs.github.io/kable/core/com.juul.kable/-peripheral/connect.html):

> If already connected, then returns immediately.

Before, we'd `await` the existing connect job, which after being connected, will always just return immediately because `await` on a completed `Deferred` immediately returns the result (i.e. would not attempt to reconnect when the connection previously completed successfully).

Whereas, this clears the `connectJob` upon successful connection, meaning subsequent calls will attempt to connect again, even though we're already connected.

core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -128,53 +133,105 @@ public class JsPeripheral internal constructor(
...
-            onServicesDiscovered(ServicesDiscoveredPeripheral(this@JsPeripheral))
-            _state.value = State.Connecting.Observes
-            logger.verbose { message = "Configuring characteristic observations" }
-            observers.onConnected()
+            openConnection()
         } catch (t: Throwable) {
             logger.error(t) { message = "Failed to connect" }
-            disconnectGatt()
+            scope.async {
+                disconnectSync(cancelConnectJob = false)
```
twyatt, line 141: I don't follow why it is desirable to _not_ cancel the connect job when library consumer requested a `disconnect`?

core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -128,53 +133,105 @@ public class JsPeripheral internal constructor(
...
+    }
+
+    private fun finalCleanup() {
+        try {
+            unregisterDisconnectedListener()
+            clearObservationsWithoutStopping()
+            bluetoothDevice.gatt?.disconnect()
+        } finally {
+            // Avoid trampling existing `Disconnected` state (and its properties) by only updating if not already `Disconnected`.
+            _state.update { previous -> previous as? State.Disconnected ?: State.Disconnected() }
```
twyatt, line 233: From what I can tell of Web Bluetooth, we don't get any status codes from the disconnected callback, so `Disconnected` on JS will never carry a status. In other words, I realize now that we don't need to do the `update` dance on JS, because we'll never trample an out-of-band disconnect event. We can just do `_status.value = Disconnected()`.

core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -280,10 +337,9 @@ public class JsPeripheral internal constructor(
     private var isDisconnectedListenerRegistered = false
     private val disconnectedListener: (JsEvent) -> Unit = {
         logger.debug { message = GATT_SERVER_DISCONNECTED }
-        _state.value = State.Disconnected()
-        unregisterDisconnectedListener()
-        observationListeners.clear()
-        connectJob = null
+        scope.async {
+            disconnectSync(cancelConnectJob = true)
+        }
```
twyatt, line 342: Why does a disconnect operation need to be spun up asynchronously from this callback?

core/src/jsMain/kotlin/Peripheral.kt
davidtaylor-juul: The problem with the previous implementation is that after a failed connection attempt, all future connection attempts would return an immediate failure on the `await` call, so it becomes impossible to retry the connection. This new implementation attempts to rectify that.

The `openConnection` method first checks if already connected and immediately returns. It is perhaps slightly inefficient to spin up the async job to do so, but it does satisfy the contract as documented.

core/src/jsMain/kotlin/Peripheral.kt
davidtaylor-juul: Testing has shown that it is problematic to perform any `gatt` operations from within the callback. Doing so can call the system to become unresponsive and the `gatt` promises tend to go into limbo and neither reject nor resolve.

This behavior is not documented, but it easily replicated. See [comment 28 on this report](https://bugs.chromium.org/p/chromium/issues/detail?id=621901#c28) for example.

core/src/jsMain/kotlin/Peripheral.kt
davidtaylor-juul: This is inside the connect job itself, and this catch is the path where it is already terminating so there is no need to cancel itself. It also avoid recursively canceling itself which makes for a somewhat confusing stack trace.

core/src/jsMain/kotlin/Peripheral.kt
davidtaylor-juul: Good catch, will remove this.

core/src/jsMain/kotlin/Peripheral.kt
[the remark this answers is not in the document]
davidtaylor-juul: Agreed, this is awkward and perhaps better naming would help.

The thinking here is that `disconnectSync` is an interface into performing a disconnect via a shared disconnection job. This is to ensure that multiple disconnection triggers all join on the same job with synchronous manipulation of state and the BLE API itself. This is also to satisfy the contract for `disconnect` as stated in the header docs.

`disconnectAsync` (like its peer `connectAsync`) creates an instance of the shared disconnection job. It should only ever be called from `disconnectSync`. That could be enforced by scoping if everything wasn't a single giant class, or the functions could be merged or renamed to make the purpose clearer. The fact that there is commentary shows it is not satisfactory as is and any suggestions would be most welcome.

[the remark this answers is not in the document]
davidtaylor-juul: Refactored to make less confusing.

core/src/jsMain/kotlin/SharedRepeatableTask.kt
```diff
@@ -0,0 +1,42 @@
...
+import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.async
+
+/**
+ * A mechanism for launching and awaiting some shared job repeatedly.
+ *
+ * The job is launched by calling [launchAsync]. Subsequent calls to [launchAsync] will return the
+ * same (i.e. shared) job if it is still executing.
+ */
+internal class SharedRepeatableTask(
```
cedrickcooke, line 15: Love this. Could it be used to improve the other platform implementations too? Obviously not gonna block a PR on that improvement.

core/src/jsMain/kotlin/SharedRepeatableTask.kt
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Launches a shared instance of the job if it is not running.
+     * Subsequent calls, while the job is still live, will return the existing job.
+     * Once the job is completed, a subsequent call will launch a new job.
+     */
+    internal fun launchAsync(start: CoroutineStart = DEFAULT): Deferred<Unit> =
+        deferred ?: scope.async(start = start) {
+            job.invoke()
+        }.apply {
```
cedrickcooke, line 30: Just double-checking how this parses. `deferred ?: (scope.async() { }.apply { })`, not `(deferred ?: scope.async() { }).apply { }`, correct?

cedrickcooke (approved):

core/src/jsMain/kotlin/SharedRepeatableTask.kt
davidtaylor-juul: Quite possibly as there are variations on this pattern in the other target implementations.

core/src/jsMain/kotlin/SharedRepeatableTask.kt
davidtaylor-juul: Good point, as it is impossible to tell from the syntax without applying the precedence rules in your head.

According to the [docs](https://kotlinlang.org/docs/reference/grammar.html#expressions) the method invocation (Postfix) operator has precedence over the Elvis operator so it would be parsed in the desired fashion.

core/src/jsMain/kotlin/SharedRepeatableTask.kt
```diff
@@ -0,0 +1,46 @@
...
+) {
+
+    private var deferred: Deferred<Unit>? = null
+
+    /**
+     * Launches a shared instance of the job if it is not running.
+     * Subsequent calls, while the job is still live, will return the existing job.
+     * Once the job is completed, a subsequent call will launch a new job.
+     */
+    internal fun launchAsync(start: CoroutineStart = DEFAULT): Deferred<Unit> =
```
twyatt, line 27: nit: At least in the JVM world, if the class is `internal` then marking all the functions as `internal` is unnecessary. 🤷

core/src/jsMain/kotlin/SharedRepeatableTask.kt
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * A mechanism for launching and awaiting some shared job repeatedly.
+ *
+ * The job is launched by calling [launchAsync]. Subsequent calls to [launchAsync] will return the
+ * same (i.e. shared) job if it is still executing.
+ */
+internal class SharedRepeatableTask(
+    private val scope: CoroutineScope,
+    private val job: suspend () -> Unit
```
twyatt, line 17: nit: rename variable to be clearer that this isn't a reference to a Coroutine `Job`

```suggestion
    private val action: suspend () -> Unit
```

or...

```suggestion
    private val task: suspend () -> Unit
```

core/src/jsMain/kotlin/SharedRepeatableTask.kt
twyatt, line 27: `start` parameter isn't used, so I'd just remove it. We can add it in later if it is needed.

```suggestion
    internal fun launchAsync(): Deferred<Unit> =
```

core/src/jsMain/kotlin/SharedRepeatableTask.kt
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Launches a shared instance of the job if it is not running.
+     * Subsequent calls, while the job is still live, will return the existing job.
+     * Once the job is completed, a subsequent call will launch a new job.
+     */
+    internal fun launchAsync(start: CoroutineStart = DEFAULT): Deferred<Unit> =
+        deferred ?: scope.async(start = start) {
+            job.invoke()
+        }.apply {
```
twyatt, line 30: If you change parameter `job` type to `suspend CoroutineScope.() -> Unit` then you can condense this syntax a little:

```suggestion
        deferred ?: scope.async(block = job).apply {
```

core/src/jsMain/kotlin/SharedRepeatableTask.kt
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * The job is launched by calling [launchAsync]. Subsequent calls to [launchAsync] will return the
+ * same (i.e. shared) job if it is still executing.
+ */
+internal class SharedRepeatableTask(
+    private val scope: CoroutineScope,
+    private val job: suspend () -> Unit
+) {
+
+    private var deferred: Deferred<Unit>? = null
```
twyatt, line 20: A `Deferred` with a return type of `Unit` feels like it should just be a `Job`? (i.e. use `launch` instead of `async`)

It would also help remove the warning at the usage site:

Screen Shot 2022-03-02 at 10 27 44 PM

core/src/jsMain/kotlin/SharedRepeatableTask.kt
twyatt, line 27: If you go w/ my above comment of changing return type to `Job`, then this function should probably be renamed to `launch`.

core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -128,56 +125,76 @@ public class JsPeripheral internal constructor(
...
+        } finally {
+            _state.value = State.Disconnected()
+        }
+    }
+
+    private fun finalCleanup() {
+        try {
+            unregisterDisconnectedListener()
+            clearObservationsWithoutStopping()
+            bluetoothDevice.gatt?.disconnect()
```
twyatt, line 194: Should this be moved into the `finally` block here as well?

core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -128,56 +125,76 @@ public class JsPeripheral internal constructor(
...
-        _state.update { previous -> previous as? State.Disconnected ?: State.Disconnected() }
+    private suspend fun closeConnection() {
+        // Guard if already disconnecting/disconnected do nothing
+        if (_state.value is State.Disconnecting || _state.value is State.Disconnected) return
+
+        try {
+            _state.value = State.Disconnecting
+            unregisterDisconnectedListener()
+            stopAllObservations()
+            bluetoothDevice.gatt?.disconnect()
```
twyatt, line 184: Should this be moved into `finally` block? to make sure that we cleanup underlying BLE connection even if `stopAllObservations` fails?

core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -128,56 +125,76 @@ public class JsPeripheral internal constructor(
...
-        val job = connectJob ?: connectAsync().also { connectJob = it }
-        job.await()
+        connectJob.launchAsync().await()
     }
 
     public override suspend fun disconnect() {
-        job.cancelAndJoinChildren()
-        connectJob = null
-        disconnectGatt()
+        connectJob.cancel()
```
twyatt, line 148: With JS being single threaded, this probably isn't an issue; but generally I'd recommend we use `cancelAndJoin` here, to ensure that any connect cancellation completes before starting the disconnect process.

In other words, if this were multithreaded, then in our `openConnection` we may be performing some actions on non-suspend lines such as `registerDisconnectListener` and it could be called while the `disconnectJob` is running if we don't `join` the cancelling `connectJob`. Whereas `join`ing means that we'll wait til we hit a suspension point (to cooperatively cancel) before the `disconnectJob` is kicked off.

core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -128,56 +125,76 @@ public class JsPeripheral internal constructor(
...
+        disconnectJob.launchAsync().await()
+    }
+
+    private suspend fun openConnection() {
+        // Guard if already connected do nothing
+        if (_state.value is State.Connected) return
+
+        logger.info { message = "Connecting" }
+        // Wait until fully disconnected before proceeding
+        _state.first { it is State.Disconnected }
```
twyatt, line 158: Would `disconnectJob.join()` accomplish this?

twyatt (commented):
Overall I really love the direction of this, definitely buttons up some major gaps in the JS connection handling.

core/src/jsMain/kotlin/SharedRepeatableTask.kt
davidtaylor-juul: Much nicer, great suggestion!

core/src/jsMain/kotlin/SharedRepeatableTask.kt
davidtaylor-juul: Good callout, it can be a Job+launch instead of Deferred+async now after all the other refactoring.

core/src/jsMain/kotlin/Peripheral.kt
davidtaylor-juul: I guess if we move it to the `finally` block and ensure that the state is updated _before_ invoking it in case it fails that may be the best approach.

core/src/jsMain/kotlin/Peripheral.kt
davidtaylor-juul: Perhaps, see other comment.

core/src/jsMain/kotlin/Peripheral.kt
davidtaylor-juul: Good catch, will make this change also.

core/src/jsMain/kotlin/Peripheral.kt
twyatt: Oh, I missed that.
I was wanting to move it to the `finally` block to make sure that a disconnect occurs even if stopping observations failed.

Maybe keeping the state update after the disconnect is appropriate? If `bluetoothDevice.gatt?.disconnect()` fails, should we be updating to `Disconnected` state? 🤷

core/src/jsMain/kotlin/Peripheral.kt
davidtaylor-juul: Probably "yes", but I think it may not be a good idea.

I've deliberately kept the job-to-job interactions out of these `closeConnection` and `openConnection` functions. In my mind this is "logic layer" and is distinct from the invoking async job layer. The logic here is easier to follow if we rely on the `_state` only.

Introducing the invoking job objects compounds the method interdependency and makes it harder to reason about the code IMHO.

core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -128,56 +125,78 @@ public class JsPeripheral internal constructor(
...
-        _state.value = State.Connected
     }
 
-    private fun closeConnection() {
-        observationListeners.clear()
-        disconnectGatt()
-        unregisterDisconnectedListener()
-        _state.value = State.Disconnected()
+    private val disconnectJob: SharedRepeatableTask = scope.sharedRepeatableTask {
+        connectJob.join()
```
twyatt, line 139: Should this be `cancelAndJoin`?

core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -128,56 +125,78 @@ public class JsPeripheral internal constructor(
...
+    }
+
+    private fun finalCleanup() {
+        try {
+            unregisterDisconnectedListener()
+            clearObservationsWithoutStopping()
+        } finally {
+            // IMPORTANT: update state FIRST as the gatt.disconnect may throw
+            _state.value = State.Disconnected()
+            bluetoothDevice.gatt?.disconnect()
```
twyatt, line 198: I'm worried that updating state first could cause other operations to continue (e.g. `_state.first { it is Disconnected }`) before the disconnect is kicked off?

Perhaps it would be better to nest another `try`-`finally`?

```suggestion
            try {
                bluetoothDevice.gatt?.disconnect()
            } finally {
                _state.value = State.Disconnected()
            }
```

core/src/jsMain/kotlin/Peripheral.kt
davidtaylor-juul: Yes we should. These are the places where we are guaranteeing that we are moving to the expected state regardless if the underlying system is borked. 

If `bluetoothDevice.gatt?.disconnect()` if broken we can be certain the connection is broken and need to be in the `Disconnected` state.

core/src/jsMain/kotlin/Peripheral.kt
davidtaylor-juul: Nope. If there is a connect in-flight it is polite to let it complete before commencing the disconnection - less chance of the underlying system getting all twisted up.

core/src/jsMain/kotlin/Peripheral.kt
davidtaylor-juul: Yes you are right. Ugly, but probably the best approach here.

core/src/jsMain/kotlin/Peripheral.kt
twyatt: Gotchya. :+1:

twyatt (approved):
Great job on this!! 💯 
_Tested this locally w/ SensorTag sample app and it works!_

davidtaylor-juul:
> Tested this locally w/ SensorTag sample app and it works!

Nice to confirm I didn't break everything lol. 🎉

PR #265, opened 2022-02-17
Remove unnecessary freezes on Native

twyatt:


Phoenix7351 (approved):
LGTM. Out of curiosity, what's "freeze" in this context? Like a semaphore?

twyatt:
> LGTM. Out of curiosity, what's "freeze" in this context? Like a semaphore?

This document offers a thorough explanation: https://kotlinlang.org/docs/multiplatform-mobile-concurrency-overview.html

This PR in particular, is removing explicit `freeze` calls that were unnecessary (likely just accidentally committed after long debug sessions when I was tracking down mutability exceptions). They are unnecessary because freezing is automatic (at runtime) on Kotlin/Native when an object crosses a thread boundary.

davidtaylor-juul (approved):

Issue #66, opened 2021-02-09
Retrieve peripheral on iOS/macOS from previous identifier peripheral


francismariano:
Hello guys.

Currently the peripheral is retrieved only with advertisement data. I think interesting to be possible to retrieve the peripheral with its UUID identifier even knowing it is not 100% relied upon for peripheral re-identification.

Example:

```
Peripheral.kt
public fun CoroutineScope.peripheral(
    uuid: Uuid
): Peripheral {
    val peripheral = CentralManager.Default.retrievePeripheral(uuid)
    return ApplePeripheral(coroutineContext, peripheral)
}
```

```
CentralManager.kt
internal fun retrievePeripheral(withIdentifiers: Uuid) : CBPeripheral {
    val peripheral = cbCentralManager.retrievePeripheralsWithIdentifiers(listOf(withIdentifiers))
    return peripheral.firstOrNull() as CBPeripheral
}
```

Thank you

twyatt:
This does seem like a useful feature. I'll see if we can provide a common identifier to allow for a common function across platforms for retrieving a known peripheral. If the other platforms doesn't provide a similar functionality then at very least, this addition does make sense as an Apple-only function. :thu

burnhamd:
This would be a useful feature.

Android at least supports something similar.
From the documentation

> To get a BluetoothDevice, use BluetoothAdapter.getRemoteDevice(String) to create one representing a device of a known MAC address (which you can get through device discovery with BluetoothAdapter) or get one from the set of bonded devices returned by BluetoothAdapter.getBondedDevices(). 

I think realistically this should be two features.
getBondedDevices
and this peripheral from identifier.

QT has a similar capability where the identifier property is the actual Mac address on Android and the masked identifier on iOS.

burnhamd:
I have added the ability to get a peripheral from an identifier to my fork. Ill raise another PR shortly.
I think handling the ID as a string makes sense.  Of course this ID cannot be used between different devices but for the same device it seems to persist at least for multiple days as a way to identify the peripheral.

Issue #277, opened 2022-03-11
WillRestoreState for IOS CBCentral Manager


burnhamd:
I need to implement the willRestoreState delegate for the CBCentralManager for an application which makes use of background mode state restoration on IOS.

Has any thought been done on the architecture of this?

Currently I am thinking to add the following to the code base where the delegate produces a shared flow 
```
public data class CBManagerRestoredState(
    val peripherals: List<CBPeripheral>? = null,
    val scannUUIDs: List<CBUUID>? = null,
    //TBD Scan Options
)

val _willRestoreSharedFlow = MutableSharedFlow<CBManagerRestoredState?>()
```

My first thought is to expose this as a package wide variable since there is no Adapter/Manager class.

Alternatively I think it may also be useful to have an Adapter/Manager class for shared code to view things like bluetooth status and this might could go in the IOS implementation.

Lastly, implementing this may break MacOSX compatibility since to restore I will need to create the CBCentralManager with a key for restoring state and this is not accepted on OS X since the app is never suspended.

Wanted to get feedback before implementing in case there is a thought already for this

twyatt:
Thanks for creating this issue. You've identified an apparent gap in the current Kable API. I'm not sure the best way to approach it, but I'll try to find some time soon to dig deeper and give meaningful feedback.

Phoenix7351:
Hi, @burnhamd. We've added iOS background functionality in https://github.com/JuulLabs/kable/pull/551 . `willRestoreState` was implemented as a `No-op`, though, as it was required by the OS to be implemented by the delegate, but there was (seemingly) no work to be done to set back up the BLE stack. Does this cover this issue?

PR #280, opened 2022-03-16
Add ability to create peripheral from identifier

burnhamd:
This PR adds the ability to create a peripheral from a public identifier as mentioned in issue #66 .  
Exposed the previously private "identifier" property in peripheral and added the ability to connect via a uuid in IOS and via a mac address in Android.

As a note, the API is not the same between android and IOS.  Android will return a peripheral for any Mac address passed.  IOS will return an empty list if there are no devices which have the same uuid identified in the system, so this is returned as null in this implementation.  Im not familiar enough with web bluetooth to know if there is equivalent functionality.


burnhamd:
@twyatt  I also have preliminary changes for the IOS willRestoreState from issue #277.  I can take this out if required.

core/src/androidMain/kotlin/Peripheral.kt
```diff
@@ -123,6 +123,24 @@ public fun CoroutineScope.peripheral(
     )
 }
 
+public fun CoroutineScope.peripheral(
+    macAddress: String,
+    builderAction: PeripheralBuilderAction = {},
+): Peripheral {
+    val builder = PeripheralBuilder()
+    builder.builderAction()
+    val bluetoothDevice = BluetoothAdapter.getDefaultAdapter().getRemoteDevice(macAddress)
```
QuantumRand, line 132: While `getDefaultAdapter()` is guaranteed to return a `BluetoothDevice`, there's no guaranty that it can be connected to.

Android builds a cache of BluetoothDevices when scanning, and this cache can be cleared by various events (dependent on manufacture/OS version), so you can't reliably connect to a device with just a Mac Address.

When a device isn't cached, its `deviceType` returns `BluetoothDevice.DEVICE_TYPE_UNKNOWN`

Returning an `AndroidPeripheral` with a `bluetoothDevice` that can't connect is something we should avoid, though I'm not sure the best way to handle that.

Returning `null` as you do with the `CentralManager` implementation may be an option, though I'm leaning more towards surfacing a more actionable `BluetoothDeviceNotCachedException` error instead.

Definitely open to other suggestions though.

@twyatt is this something you've given any thought to?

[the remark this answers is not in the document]
burnhamd: Good point.  I can implement this with null unless there is consensus for an Exception in stead.  Essentially it would work more similarly to the IOS implementation where there is a search for existing devices with same id and return null if none exist.

core/src/androidMain/kotlin/Peripheral.kt
```diff
@@ -496,5 +515,5 @@ private fun checkBluetoothAdapterState(
     }
 }
 
-internal actual val Peripheral.identifier: String
+public actual val Peripheral.identifier: String
```
twyatt, line 518:
```suggestion
public typealias Identifier = String

public actual val Peripheral.identifier: Identifier
```

core/src/appleMain/kotlin/Peripheral.kt
```diff
@@ -86,6 +86,37 @@ public actual fun CoroutineScope.peripheral(
...
+    val builder = PeripheralBuilder()
+    builder.builderAction()
+    return ApplePeripheral(
+        coroutineContext,
+        peripheral,
+        builder.observationExceptionHandler,
+        builder.onServicesDiscovered,
+        builder.logging
+    )
+}
```
twyatt, line 103: This function should piggyback on your function with `CBPeripheral` in the signature.

```suggestion
public actual fun CoroutineScope.peripheral(
    identifier: Identifier,
    builderAction: PeripheralBuilderAction,
): Peripheral {
    val cbPeripheral = CentralManager.Default.retrievePeripheral(identifier)
        ?: throw NoSuchElementException("Peripheral with UUID $identifier not found")
    return peripheral(cbPeripheral, builderAction)
}
```

core/src/commonMain/kotlin/Peripheral.kt
```diff
@@ -241,4 +241,13 @@ internal suspend inline fun <reified T : State> Peripheral.suspendUntilOrThrow()
...
+/**
+ * The identifier of the remote device
+ *
+ * This identifier is the identifier returned by the operating system for the address of the device.
+ * The value of the identifier will differ between operating system implementations.  On iOS
+ * the identifier is a UUID.  On Android the identifier is a MAC address.  For security reasons
+ * the identifier is not a MAC address on iOS and may be a randomized MAC on Android.  The identifier
+ * may also change.
+ */
+public expect val Peripheral.identifier: String
```
twyatt, line 253:
```suggestion
public expect class Identifier

/**
 * The identifier of the remote device
 *
 * This identifier is the identifier returned by the operating system for the address of the device.
 * The value of the identifier will differ between operating system implementations.  On iOS
 * the identifier is a UUID.  On Android the identifier is a MAC address.  For security reasons
 * the identifier is not a MAC address on iOS and may be a randomized MAC on Android.  The identifier
 * may also change.
 */
public expect val Peripheral.identifier: Identifier
```

core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -410,5 +410,5 @@ public class JsPeripheral internal constructor(
     override fun toString(): String = "Peripheral(bluetoothDevice=${bluetoothDevice.string()})"
 }
 
-internal actual val Peripheral.identifier: String
+public actual val Peripheral.identifier: String
     get() = (this as JsPeripheral).platformIdentifier
```
twyatt, line 414:
```suggestion
public typealias Identifier = String

public actual val Peripheral.identifier: Identifier
    get() = (this as JsPeripheral).platformIdentifier
```

core/src/appleMain/kotlin/Peripheral.kt
```diff
@@ -407,5 +438,5 @@ private fun NSError.toStatus(): State.Disconnected.Status = when (code) {
     else -> Unknown(code.toInt())
 }
 
-internal actual val Peripheral.identifier: String
+public actual val Peripheral.identifier: String
     get() = (this as ApplePeripheral).platformIdentifier
```
twyatt, line 442:
```suggestion
public typealias Identifier = Uuid

public actual val Peripheral.identifier: Identifier
    get() = (this as ApplePeripheral).platformIdentifier
```

Also requires that `platformIdentifer` to read:

```
internal val platformIdentifier = uuidFrom(cbPeripheral.identifier.UUIDString)
```

core/src/androidMain/kotlin/Peripheral.kt
```diff
@@ -123,6 +123,25 @@ public fun CoroutineScope.peripheral(
     )
 }
 
+public fun CoroutineScope.peripheral(
+    macAddress: String,
```
twyatt, line 127:
```suggestion
public actual fun CoroutineScope.peripheral(
    identifier: Identifier,
```

core/src/androidMain/kotlin/Peripheral.kt
```diff
@@ -123,6 +123,25 @@ public fun CoroutineScope.peripheral(
...
+    return AndroidPeripheral(
+        coroutineContext,
+        bluetoothDevice,
+        builder.transport,
+        builder.phy,
+        builder.observationExceptionHandler,
+        builder.onServicesDiscovered,
+        builder.logging,
+    )
+}
```
twyatt, line 143: This function can piggyback on the function with `BluetoothDevice` in its signature.

```suggestion
public fun CoroutineScope.peripheral(
    identifier: Identifier,
    builderAction: PeripheralBuilderAction = {},
): Peripheral {
    val bluetoothDevice = BluetoothAdapter.getDefaultAdapter().getRemoteDevice(identifier)
    return peripheral(bluetoothDevice, builderAction)
}
```

twyatt (commented):
Thanks for kicking this effort off @burnhamd!

I reviewed it specifically for the identifier being exposed. The additional (commented out) code should be applied to a separate PR, if possible.

I've added comments re: introducing an `Identifier` `expect class`. I think that will allow for a unified common API for creating `Peripheral` objects across all the platforms.

burnhamd:
@twyatt I think all changes should be included per your review.

Just want to mention that I did not create an expected function for generating with a identifier into the common peripheral class.

I would assume that once the JS implementation is done, then we could add it there and complete the api compatibility for all platforms.

twyatt:
> @twyatt I have a comment about the Identifier. I am using this to store the identifier to disk so that I can recover the address and reconnect. I agree that keeping the Identifier as a UUID type in IOS makes things cleaner, but it also requires that I now handle the identifiers differently when creating them on my application side. I think this will be common case. Thoughts on just keeping Identifier as String?

Although the API doesn't provide this guarantee (which is unfortunate — as I'd hate to unintentionally break the implementation in the future) but you can call `toString` on all the types currently used for `Identifier`. For the platforms that `typealias Indentifier = String` the `toString` will return the same `String`, and on iOS `toString` will return the string representation of the `Uuid` (I believe).

core/src/appleMain/kotlin/CentralManagerDelegate.kt
```diff
@@ -18,17 +18,30 @@ import platform.Foundation.NSNumber
...
+/*
+TBD Not implemented yet.  This will handle CBManagerWillRestoreState
+public data class CBManagerRestoredState(
+    val peripherals: List<CBPeripheral>? = null,
+    val scannUUIDs: List<CBUUID>? = null,
+    //TBD Scan Options
+)
+
+*/
+
```
twyatt, line 30: Please move this to another PR.

core/src/appleMain/kotlin/CentralManagerDelegate.kt
```diff
@@ -18,17 +18,30 @@ import platform.Foundation.NSNumber
...
     internal val onDisconnected = _onDisconnected.asSharedFlow()
-
     private val _state = MutableStateFlow<CBManagerState?>(null)
     val state: Flow<CBManagerState> = _state.filterNotNull()
 
-    sealed class Response {
+    /*
+    TBD Not implemented yet.  This will handle CBManagerWillRestoreState
+    private val _willRestoreStateFlow = MutableSharedFlow<CBManagerRestoredState?>()
+    */
```
twyatt, line 42: This as well.

core/src/appleMain/kotlin/CentralManagerDelegate.kt
```diff
@@ -114,7 +127,18 @@ internal class CentralManagerDelegate : NSObject(), CBCentralManagerDelegateProt
...
+    override fun centralManager(central: CBCentralManager, willRestoreState: Map<Any?, *>){
+        val peripherals =
+                willRestoreState[CBCentralManagerRestoredStatePeripheralsKey] as List<CBPeripheral>?
+        val scanUUIDs =
+                willRestoreState[CBCentralManagerRestoredStateScanServicesKey] as List<CBUUID>?
+
+        _willRestoreStateFlow.tryEmit(CBManagerRestoredState(
+            peripherals, scanUUIDs))
+    }
+    */
```
twyatt, line 141: This too, please.

core/src/appleMain/kotlin/Peripheral.kt
```diff
@@ -407,5 +422,11 @@ private fun NSError.toStatus(): State.Disconnected.Status = when (code) {
...
-internal actual val Peripheral.identifier: String
-    get() = (this as ApplePeripheral).platformIdentifier
+public actual typealias Identifier = Uuid
+
+public actual val Peripheral.identifier: Identifier
+    get() = (this as ApplePeripheral).platformIdentifier.toUuid()
+
+public actual fun String.toIdentifier(): Identifier {
+    return uuidFrom(this)
+}
```
twyatt, line 432: nitpick

```suggestion
public actual fun String.toIdentifier(): Identifier = uuidFrom(this)
```

core/src/appleMain/kotlin/Uuid.kt
```diff
@@ -11,3 +11,4 @@ internal fun CBUUID.toUuid(): Uuid = when (UUIDString.length) {
     4 -> uuidFrom("0000$UUIDString-0000-1000-8000-00805F9B34FB")
     else -> uuidFrom(UUIDString)
 }
+internal fun NSUUID.toUuid(): Uuid = uuidFrom(UUIDString)
```
twyatt, line 14: nitpick: add newline before this line

```suggestion

internal fun NSUUID.toUuid(): Uuid = uuidFrom(UUIDString)
```

core/src/commonMain/kotlin/Peripheral.kt
```diff
@@ -241,4 +241,17 @@ internal suspend inline fun <reified T : State> Peripheral.suspendUntilOrThrow()
...
+ *
+ * This identifier is the identifier returned by the operating system for the address of the device.
+ * The value of the identifier will differ between operating system implementations.  On iOS
+ * the identifier is a UUID.  On Android the identifier is a MAC address.  For security reasons
+ * the identifier is not a MAC address on iOS and may be a randomized MAC on Android.  The identifier
+ * may also change.
+ */
+public expect val Peripheral.identifier: Identifier
+
+public expect fun String.toIdentifier(): Identifier
```
twyatt, line 257: Being that the functionality of `Identifier` grew, I think it makes sense, for organizational purposes, to move this (and the corresponding platform implementations) to `Identifier.kt` files.

twyatt (commented):
Just a few remaining details to address, otherwise looks great!

core/src/appleMain/kotlin/CentralManager.kt
```diff
@@ -41,6 +41,12 @@ public class CentralManager internal constructor() {
         if (cbCentralManager.isScanning) cbCentralManager.stopScan()
     }
 
+    internal fun retrievePeripheral(withIdentifier: Uuid): CBPeripheral? {
+        val peripherals = cbCentralManager.retrievePeripheralsWithIdentifiers(listOf(withIdentifier.toNSUUID()))
+        if (peripherals.isEmpty()) return null
+        return peripherals.first() as CBPeripheral
+    }
```
twyatt, line 48: We can simplify this slight as:

```suggestion
    internal fun retrievePeripheral(withIdentifier: Uuid): CBPeripheral? =
        cbCentralManager
            .retrievePeripheralsWithIdentifiers(listOf(withIdentifier.toNSUUID()))
            .firstOrNull() as? CBPeripheral
```

twyatt (approved):
Looks great! Awesome work!

twyatt:
Looking into if we could define a "common" version of creating a `Peripheral` via `Identifier`, it is really unfortunate that Javascript's `getDevices` returns a `Promise` (meaning we'd want to expose it as a `suspend` function).

So, we could have a `common` function as:

```kotlin
public expect suspend fun CoroutineScope.peripheral(
    identifier: Identifier,
): Peripheral
```

But marking an extension function of `CoroutineScope` as a `suspend` creates Coroutine scope ambiguity:

Screen Shot 2022-03-24 at 11 53 29 AM

Some options we could take:

1. Don't expose a "common" version of creating a `Peripheral` from `Identifier`
    - Pros: Keeps API simple for the exposed methods on Android and Apple
    - Cons: This puts the onus on the library consumer to implement restoring from `Identifier` on their supported platforms
2. Expose creating `Peripheral` from `Identifier` with an alternate method signature (see below)
    - Pros: Allows library consumers to restore `Peripheral`s via `Identifier` in their `common` code
    - Cons: Inconsistent API (does not match existing `Peripheral` creating methods)

Possible alternate method signature (no longer an extension function on `CoroutineScope` — is has been moved as first function parameter):

```kotlin
public expect suspend fun peripheral(
    scope: CoroutineScope,
    identifier: Identifier,
    builderAction: PeripheralBuilderAction,
): Peripheral
```

core/src/appleMain/kotlin/Peripheral.kt
```diff
@@ -74,15 +74,29 @@ import kotlin.coroutines.cancellation.CancellationException
 public actual fun CoroutineScope.peripheral(
     advertisement: Advertisement,
     builderAction: PeripheralBuilderAction,
+): Peripheral = peripheral(advertisement.cbPeripheral, builderAction)
+
+public fun CoroutineScope.peripheral(
+    identifier: Identifier,
+    builderAction: PeripheralBuilderAction,
```
twyatt, line 81: Just noticed that this should have a default parameter to match Android's method signature.

```suggestion
    builderAction: PeripheralBuilderAction = {},
```

burnhamd:
> Looking into if we could define a "common" version of creating a `Peripheral` via `Identifier`, it is really unfortunate that Javascript's `getDevices` returns a `Promise` (meaning we'd want to expose it as a `suspend` function).
> 
> So, we could have a `common` function as:
> 
> ```kotlin
> public expect suspend fun CoroutineScope.peripheral(
>     identifier: Identifier,
>     builderAction: PeripheralBuilderAction = {},
> ): Peripheral
> ```
> 
> But marking an extension function of `CoroutineScope` as a `suspend` creates Coroutine context ambiguity:
> 
> Screen Shot 2022-03-24 at 11 53 29 AM
> 
> Some options we could take:
> 
> 1. Don't expose a "common" version of creating a `Peripheral` from `Identifier` _(current form of this PR)_
>    
>    * Pros: Keeps API simple for the exposed methods on Android and Apple
>    * Cons: This puts the onus on the library consumer to implement restoring from `Identifier` on their supported platforms
> 2. Expose creating `Peripheral` from `Identifier` with an alternate method signature (see below)
>    
>    * Pros: Allows library consumers to restore `Peripheral`s via `Identifier` in their `common` code
>    * Cons: Inconsistent API (does not match existing `Peripheral` creating methods)
> 
> Possible alternate method signature (no longer an extension function on `CoroutineScope` — is has been moved as first function parameter):
> 
> ```kotlin
> public expect suspend fun peripheral(
>     scope: CoroutineScope,
>     identifier: Identifier,
>     builderAction: PeripheralBuilderAction,
> ): Peripheral
> ```

Personally I am in favor of option 2.  We are already using this type of call with the class we are wrapping the peripheral with so at least for my project, it makes things more consistent.  But I would think to do this, all constructors should move to this type of call.  IE a API breaking change.  I'll let others chime in but this is my opinion.

burnhamd:
> > Personally I am in favor of option 2.
> 
> I prefer option 2 as well.
> 
> > But I would think to do this, all constructors should move to this type of call.
> 
> I think long term, we should move all `Peripheral` creation towards this type of function signature (with `CoroutineScope` a function parameter rather than being the extension function receiver). IMHO it feels more explicit and clearer to read.
> 
> > IE a API breaking change.
> 
> We expected larger changes might be needed as we integrate more features; hence why we're still pre-1.0. We do our best to have deprecation cycles, but being in the 0.x.y versioning means hopefully library consumers expect some larger API breaking changes every once in a while. 🤷
> 
> That being said, I think it makes sense to get this merged as-is (under option number 1) with the intent of having deprecation cycles to move towards option 2. A deprecation cycle also gives library consumers time to chime in if there are strong opinions about where the API is going.

Sounds good.

core/src/androidMain/kotlin/Peripheral.kt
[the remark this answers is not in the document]
twyatt: For others following this thread, I responded via https://github.com/JuulLabs/kable/pull/280#discussion_r832888616.

cedrickcooke (approved):

davidtaylor-juul (approved):
🚀

twyatt:
Thanks for the contribution @burnhamd!

twyatt:
> > > Personally I am in favor of option 2.
> > 
> > 
> > I prefer option 2 as well.
> > > But I would think to do this, all constructors should move to this type of call.
> > 
> > 
> > I think long term, we should move all `Peripheral` creation towards this type of function signature (with `CoroutineScope` a function parameter rather than being the extension function receiver). IMHO it feels more explicit and clearer to read.
> > > IE a API breaking change.
> > 
> > 
> > We expected larger changes might be needed as we integrate more features; hence why we're still pre-1.0. We do our best to have deprecation cycles, but being in the 0.x.y versioning means hopefully library consumers expect some larger API breaking changes every once in a while. 🤷
> > That being said, I think it makes sense to get this merged as-is (under option number 1) with the intent of having deprecation cycles to move towards option 2. A deprecation cycle also gives library consumers time to chime in if there are strong opinions about where the API is going.
> 
> Sounds good.

Created #286 to track this effort.

PR #285, opened 2022-03-24
Propagate observation rewire failures during connection as reconnection failures

cedrickcooke:
Fixes a bug where if a device briefly reconnects but then disconnects during rewiring it would end up in the `Connected` state.

### DIFF
--- a/core/src/commonMain/kotlin/Observers.kt
+++ b/core/src/commonMain/kotlin/Observers.kt
@@ -99,6 +99,7 @@ internal class Observers<T>(
                 throw cancellation
             } catch (e: Exception) {
                 characteristicChanges.emit(Error(characteristic, e))
+                throw ConnectionLostException("Failed to rewire observers during connection attempt", e)
             }
         }
     }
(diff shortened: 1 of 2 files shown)


twyatt (approved):

twyatt:
While it is fresh in my mind, the following is a good comment to add to the release:

> **Behavior change:** A failure to spin-up observations on connect will no longer propagate failures to the corresponding observation flows. Instead, the failure will propagate through the `Peripheral.connect` function **and** cancel the underlying connection — leaving observation flows in an active state.

twyatt:
@cedrickcooke perhaps we should reword or add clarification to the documentation?

https://github.com/JuulLabs/kable/blob/c06e6b3985b599979fb462c9b96b384f764af059/core/src/commonMain/kotlin/Peripheral.kt#L179-L184

As it reads now, I think it is incorrect with this change. In that: a `NoSuchElementException` will only be propagated if we spin up the observation when already connected.

I think this might have been why I originally chose to terminate the flow from `onConnected`; it just made failures more consistent/predictable where they would be propagated from.

I still think this behavior change is correct, just warrants us updating documentation accordingly.

sdonn3 (approved):
solid comments!

Issue #165, opened 2021-09-28
(iOS) Peripheral.state not changing on turning off Bluetooth


lammertw:
I'm presenting a device connection state to the user. For that I'm observing `Peripheral.state`. But when I run my app with a device connected and then turn off Bluetooth on my iPhone, the `Peripheral.state` does not change and stays `Connected`. When I turn off my Bluetooth device it does change to `Disconnected` so it would appear I'm not doing anything wrong.

twyatt:
Thanks for reporting the issue.

We probably need to do something similar to what was done for Android in #133 (where we monitor BLE state and explicitly update the state when we detect that BLE has been turned off).

burnhamd:
Looking at this it seems like the best thing is to just add something like this to the constructor of the peripheral
`        centralManager.delegate.state.onEach { event ->
            if(event == CBManagerStatePoweredOff){
                _state.value = State.Disconnected()
            }
        }.launchIn(scope)`

What Im not are about is how to properly handle state resetting. My thought is that this should also be considered disconnected. If anyone else knows if this is appropriate please let me know.

burnhamd:
I have this pretty much ready to go.  Will create PR once #280 is merged in since it is branched from that

PR #287, opened 2022-03-25
Add disconnect to Apple peripheral when CBCentralManager turns off

burnhamd:
This adds a disconnect to the peripheral when central manager turns off.  After IOS version 10 there is no didDisconnect callback issued when the Central manager is returned which makes it seem that the device can still be subscribed to.  This adds a disconnect to the peripheral and leaves it up to the client application to decide to reconnect or not. Refs issue #165


davidtaylor-juul (approved):
Seems sensible to me, provided this has no effect when the peripheral is already disconnected, which should be the case but is worth testing.

cedrickcooke (approved):

twyatt (approved):

Issue #290, opened 2022-03-29
Connecting to device with bluetooth powered off results in device showing connecting status which doesn't change


burnhamd:
Attempting to connect to a device after turning off bluetooth on IOS will result in the peripheral entering a state of CONNECTING.
In the console there will be a print out from CoreBluetooth showing API misuse, but no exception thrown.   The device does not connect after bluetooth is turned back on.

I am thinking that there should be a check for the bluetooth status before connecting.  Since it is suspending I think we could suspend until bluetooth is available like the scanner.  Option 2 would be to implement an exception if bluetooth is off.

Current implementation idea for suspending function

``public override suspend fun connect() {
        //need to check that coreblueooth is turned on and available
        centralManager.delegate.state
            .first { state -> state == CBCentralManagerStatePoweredOn }
        connectJob.updateAndGet { it ?: connectAsync() }!!.await()
    }``

PR #300, opened 2022-04-05
Fix `Peripheral` state not updating upon connection

burnhamd:
References issue #290 

Found issue where if the device is disconnected and delegate does not return a didDisconnect callback on IOS  (for example when bluetooth is turned off) that the state flow would stay in connecting and never show connected in the peripheral side during reconnect.  

This fixes that by using shared flow instead of state flow


twyatt:
Sorry, I don't follow how using a `SharedFlow` vs. a `StateFlow` would fix the issue?

burnhamd:
> Sorry, I don't follow how using a `SharedFlow` vs. a `StateFlow` would fix the issue?

It took us a while to figure this one out.  So after the change which disconnects the peripherals in the event that bluetooth is powered off, the connectionState would remain in the previous state and not change.  This would result in the peripheral state being DISCONNECTED, and the ConnectionState still having the most recent state emitted.  Then once you call a connect on the peripheral, the peripheral state would change to connected, but the ConnectionState would not emit again because the state was the same as before.  

So changing to a shared flow ensures that the ConnectionState emits even if it is the same value as before.

core/src/appleMain/kotlin/CentralManagerDelegate.kt
```diff
@@ -58,7 +58,7 @@ internal class CentralManagerDelegate : NSObject(), CBCentralManagerDelegateProt
         ) : ConnectionEvent()
     }
 
-    private val _connectionState = MutableStateFlow<ConnectionEvent?>(null)
+    private val _connectionState = MutableSharedFlow<ConnectionEvent?>()
```
twyatt, line 61:
```suggestion
    // `SharedFlow` (instead of `StateFlow`) as downstream needs non-distinct items, as it feeds individual `Peripheral`
    // states. If, for example, this flow emits `Disconnected` then downstream `Peripheral` connects and updates its own
    // state to `Connected`, this flow may still hold `Disconnected` but we'll need to emit another `Disconnected` to
    // update the `Peripheral` state with.
    private val _connectionState = MutableSharedFlow<ConnectionEvent>()
    val connectionState: Flow<ConnectionEvent> = _connectionState.asSharedFlow()
```

twyatt (commented):
Thanks @burnhamd! Great investigative work to figure out the problem and fix! 💯

davidtaylor-juul (approved):
LGTM, thank you @burnhamd

twyatt (approved):

Issue #288, opened 2022-03-25
Request: Add bluetooth status for IOS CBCentralManager and Android BluetoothAdapter


burnhamd:
Currently we are required to add code for checking the bluetooth status in the client application for seeing if bluetooth is enabled, turned on, and permissions are granted.  Would be nice to have a unified status flow which represented the state of the bluetooth on the system so that we can present errors and requests to the user in those cases.

Thinking there can be the following cases

- PoweredOn
-  Unauthorized (maybe with list of authorizations required?) This will be trickier for Android since there will also be location service involved.
-  Resetting
-  PoweredOff
-  Unknown
- Unsupported

twyatt:
Related to #737, and sadly: we likely won't be supporting this in Kable, see #737 for the rationale.

PR #303, opened 2022-04-05
Add bluetooth status flow for Android and IOS

burnhamd:
References issue #288.  Implements a bluetooth status flow which exposes the status of the CBCentral in IOS and on Android the BluetoothAdapter.  We have added this for our project so that we know the current status reported by the delegate being used by the CentralManager.default we have created with kable library.

I am still not sure how this should be handled in JS.  Im not super familiar with WebBluetooth but it seems that these types of statuses are not available.  There is a method getAvailability, but this could result in a state that is not accurate so it seems in JS there is a bit of try and see with bluetooth permissions since the browser automatically requests permissions anyway.

Also I realize that there are numerous places that we are checking the status.  They could be refactored to use this, but for now I wanted to check the approach before changing how it is handled in peripherals.


davidtaylor-juul:
I have some code snippets for Web BLE around this that may be useful to you here:
https://gist.github.com/davidtaylor-juul/3856335418e69b312899b416e9295130

The `availabilitychanged` event is experimental and not broadly supported but is really all we have in Web BLE right now.

twyatt:
Sorry, got buried with other work tasks; I will try to look over this soon.

twyatt:
@burnhamd I've created #335 (that was heavily inspired by this PR), it takes on a slightly different different API shape for the purposes of being able to provide support across all supported platforms.

Let me know what you think, and if it would fit your needs.

twyatt:
Superseded by #335.
Thanks @burnhamd for kicking off this effort (#335 was heavily inspired by your work in this PR). ❤️

PR #304, opened 2022-04-07
Retry service discovery if no services are discovered

### LINKED ISSUE #295: Services are sometimes missing
We're seeing the following exception thrown when attempting to `write` to peripheral:

```java
Fatal Exception: java.util.NoSuchElementException: Service <redacted> not found
       at com.juul.kable.ProfileKt.obtain-vK39g5o(Profile.kt:141)
       at com.juul.kable.AndroidPeripheral.write(Peripheral.kt:306)
       at <redacted>.send(<redacted>.kt:190)
       at <redacted>$send$1.invokeSuspend(<redacted>.kt:12)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
       at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:104)
       at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
       at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
       at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
       at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
```

According to [BluetoothGatt returns empty service list after connection](https://stackoverflow.com/questions/32363931/bluetoothgatt-returns-empty-service-list-after-connection), Android may sometimes incorrectly provide an empty collection of services and that retrying a 3-5 times may be necessary.

twyatt:
Closes #295


cedrickcooke (approved):

core/src/androidMain/kotlin/Peripheral.kt
```diff
@@ -254,12 +258,24 @@ public class AndroidPeripheral internal constructor(
...
         logger.verbose { message = "discoverServices" }
-        connection.execute<OnServicesDiscovered> {
-            discoverServices()
-        }
-        _discoveredServices = withContext(connection.dispatcher) {
-            connection.bluetoothGatt.services.map(::DiscoveredService)
+
+        repeat(DISCOVER_SERVICES_RETRIES) { attempt ->
+            connection.execute<OnServicesDiscovered> {
+                discoverServices()
```
QuantumRand, line 264: Is it ok to retry in quick succession with no delay?

twyatt: I _think_ so. But good point. Android BLE always seems pretty delicate in that regard. I'll add a short delay. 👍

twyatt: Thought about this some more, I don't _think_ there is any need to add an artificial delay.

Partly because the `connection.execute` is having to do thread switching so adds a very slight delay anyways, but also I don't really want to add a delay prematurely if it isn't necessary (plus, I don't even know what delay would be appropriate). I figure we'll try this out, and if we still see issues on our internal libraries, then I'll check logs to see if it was super fast failures — one after the other.

It is also only 5 attempts, so it isn't like it is going to flood the system with requests. 🤷

QuantumRand (approved):

PR #323, opened 2022-04-25
Throw `BluetoothDisabledException` instead of `IllegalStateException` on scan

sdonn3:
When the scanner returns null on Android, use BluetoothDisabledException to catch in app.

### DIFF
--- a/core/src/androidMain/kotlin/Scanner.kt
+++ b/core/src/androidMain/kotlin/Scanner.kt
@@ -35,7 +35,7 @@ public class AndroidScanner internal constructor(
         ?: error("Bluetooth not supported")
 
     public override val advertisements: Flow<Advertisement> = callbackFlow {
-        val scanner = checkNotNull(bluetoothAdapter.bluetoothLeScanner) { "Bluetooth disabled." }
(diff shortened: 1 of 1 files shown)


core/src/androidMain/kotlin/Scanner.kt
```diff
@@ -35,7 +35,7 @@ public class AndroidScanner internal constructor(
         ?: error("Bluetooth not supported")
 
     public override val advertisements: Flow<Advertisement> = callbackFlow {
-        val scanner = checkNotNull(bluetoothAdapter.bluetoothLeScanner) { "Bluetooth disabled." }
+        val scanner = bluetoothAdapter.bluetoothLeScanner ?: throw BluetoothDisabledException("Bluetooth disabled.")
```
twyatt, line 38: Class name already describes what went wrong. My preference would be to not be redundant with the `message` (unless we can offer more information).

```suggestion
        val scanner = bluetoothAdapter.bluetoothLeScanner ?: throw BluetoothDisabledException()
```

cedrickcooke (approved):

PR #327, opened 2022-05-06
Fix JS connect exception propagation

cedrickcooke:


core/src/jsMain/kotlin/SharedRepeatableTask.kt
```diff
@@ -1,44 +1,35 @@
...
     private val scope: CoroutineScope,
-    private val task: suspend CoroutineScope.() -> Unit
+    private val task: suspend CoroutineScope.() -> T
 ) {
 
-    private var job: Job? = null
+    private var deferred: Deferred<T>? = null
+
+    /** If there is a running [job][Deferred] returns it. Otherwise, launches and returns a new [job][Deferred]. */
+    fun getOrLaunchAsync() = deferred
```
twyatt, line 22: I think the `Launch` in the name is misleading?
Perhaps just:

```suggestion
    fun getOrAsync() = deferred
```

twyatt (approved):

core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -125,28 +125,28 @@ public class JsPeripheral internal constructor(
...
+            @Suppress("DeferredResultUnused") // Safe to ignore Deferred because the result is shared elsewhere
+            disconnectJob.getOrAsync()
             throw IOException("Connection attempt failed", cause)
         }
     }
 
-    private val disconnectJob: SharedRepeatableTask = scope.sharedRepeatableTask {
-        connectJob.join()
+    private val disconnectJob: SharedRepeatableTask<Unit> = scope.sharedRepeatableTask {
+        connectJob.cancelAndJoin()
```
davidtaylor-juul, line 140: It sounds silly, but it took a long time to find the right combination of cancel/join of these jobs during both explicit and internally triggered disconnect events. Can we please keep this PR to modify the job `await` stuff only, and move any other changes to a separate PR for testing.

I should have added comments at the time to explain why it is like it is, sorry about that. This innocuous looking code is quite brittle.

Strongly suggest to keep `connectJob.join()` here and move to separate PR.

core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -125,28 +125,28 @@ public class JsPeripheral internal constructor(
...
 
     public override suspend fun connect() {
-        connectJob.launch().join()
+        connectJob.getOrAsync().await()
     }
 
     public override suspend fun disconnect() {
-        connectJob.cancelAndJoin()
-        disconnectJob.launch().join()
+        disconnectJob.getOrAsync().await()
```
davidtaylor-juul, line 149: Again, not-obvious and no comments, my bad.

Strongly suggest the `connectJob.cancelAndJoin()` remain here and move that change to a separate PR.

core/src/jsMain/kotlin/Peripheral.kt
```diff
@@ -310,8 +310,8 @@ public class JsPeripheral internal constructor(
     private var isDisconnectedListenerRegistered = false
     private val disconnectedListener: (JsEvent) -> Unit = {
         logger.debug { message = GATT_SERVER_DISCONNECTED }
-        connectJob.cancel()
-        disconnectJob.launch()
+        @Suppress("DeferredResultUnused") // Safe to ignore Deferred because the result is shared elsewhere
+        disconnectJob.getOrAsync()
```
davidtaylor-juul, line 314: And here, strongly suggest the `connectJob.cancel()` remain here and move that change to a separate PR.

davidtaylor-juul (changes requested):
I'm very uncomfortable with the changes to the job interactions. It was a LOT of debugging to find where/how Web BLE would get twisted and seize up during connect/disconnect. 

Happy to experiment with trying a different combination that looks logically correct, but would rather that is in a separate PR so it can be tested in isolation. There are some very subtle and undocumented things going on when the Web BLE `gattserverdisconnected` callback is triggered and "logically correct" very well may lock up Web BLE.

davidtaylor-juul (approved):
🚀

PR #335, opened 2022-06-02
Add `Bluetooth.availability` flow

twyatt:
Provides a flow that can be subscribed to for determining if bluetooth is available on the system:

```kotlin
Bluetooth.availability.collect {
    when (it) {
        Available -> // todo: handle when bluetooth is available
        is Unavailable -> println("Bluetooth is not available because ${it.reason}")
    }
}
```

Checks (and emits updates on changes) that the bluetooth adapter is on.

On Android 11 and earlier, performs an initial check for location services (which are no longer needed for Android 12 and newer). Unfortunately, could not find a way to monitor changes to location services status, so only the initial emission of status is provided.

On Android, bluetooth permissions are not taken into consideration for determining availability, as the mechanisms for checking/asking for permissions vary greatly.

Related to https://github.com/JuulLabs/kable/issues/288.


core/src/appleMain/kotlin/Bluetooth.kt
```diff
@@ -0,0 +1,38 @@
...
+    Unsupported, // CBManagerState.unsupported
+    Unknown, // CBManagerState.unknown
+}
+
+@OptIn(UnsafeNumber::class)
+internal actual val bluetoothAvailability: Flow<Bluetooth.Availability> =
+    CentralManager.Default.delegate.state.map { state ->
+        when (state) {
+            CBCentralManagerStatePoweredOn -> Available
+            CBCentralManagerStatePoweredOff -> Unavailable(reason = Unknown)
```
davidtaylor-juul, line 32: IMHO would be nice to surface `PoweredOff` as a state rather than `Unknown` for situations where the user has enabled Airplane mode so the UI can present advice accordingly.

twyatt: Unfortunately JS (https://developer.mozilla.org/en-US/docs/Web/API/Bluetooth/onavailabilitychanged) only provides a boolean of availability, so it degrades the common API to the same available/unavailable state. 🤷

twyatt: Good call out. 👍

core/src/jsMain/kotlin/Bluetooth.kt
```diff
@@ -1,62 +1,36 @@
...
+        trySend(if (isAvailable) Available else Unavailable(reason = null))
+    }
+
+    bluetooth.apply {
+        addEventListener(AVAILABILITY_CHANGED, listener)
+        awaitClose {
+            removeEventListener(AVAILABILITY_CHANGED, listener)
+        }
+    }
+}
```
davidtaylor-juul, line 27: It looks like this will crash on the `checkNotNull` below if there is no bluetooth object, is that preferred to emitting `Unavailable`?

Also, if the browser does not support the `availabilitychanged` event this will never emit.

Perhaps it would be useful to emit `Unavailable` with a suitable reason for both these scenarios instead?

donadev:
This feature will be very useful once available. Any prevision on when it will be merged to main?

twyatt:
> This feature will be very useful once available. Any prevision on when it will be merged to main?

I'm hoping to dedicate some time to this in the coming days.

twyatt:
Build is blocked by:

```
/Users/runner/work/kable/kable/core/src/androidMain/kotlin/Advertisement.kt:37: Error: Missing permissions required by BluetoothDevice.getName: android.permission.BLUETOOTH_CONNECT [MissingPermission]
        get() = bluetoothDevice.name
                ~~~~~~~~~~~~~~~~~~~~
```

Will be resolved by bumping/supporting Android target API to 31 (which I'll put up another PR for soon).

burnhamd:
Looks good.  Does exactly what I was thinking.

twyatt:
Note to self: integrate into SensorTag sample app before merging.

davidtaylor-juul (approved):

PR #369, opened 2022-08-03
Use shared bluetooth state `BroadcastReceiver`

### LINKED ISSUE #359: Too many receivers, total of 1000
In the context of my app I need to retrieve the peripheral in order to get the identifier.

The problem is that, on Android, it [registers a broadcast receiver](https://github.com/JuulLabs/kable/blob/main/core/src/androidMain/kotlin/Peripheral.kt#L137) every time and I quickly reach the limit of 1000 receivers which is prohibited by the OS.

I open this issue to discuss about ways to solve the problem in order to contribute by a pull request soon.

I suggest to share one broadcast receiver across all peripherals.

As a possible workaround, I will try to cancel the attached coroutine scope in order to [unregister the receiver](https://github.com/JuulLabs/kable/blob/main/core/src/androidMain/kotlin/Peripheral.kt#L146).

twyatt:
Closes #359


nverinaud:
I confirm this fixes #359 !

cedrickcooke (approved):

sdonn3 (approved):

dzgeorgy:
Shouldn't there be `awaitClose` call in callbackFlow?

twyatt:
> Shouldn't there be `awaitClose` call in callbackFlow?

Yikes, I think you're right. Oversight on my part.

twyatt:
Thanks @dzgeorgy, confirmed that `awaitClose` definitely should've been used.
Fixed in #399.

PR #376, opened 2022-09-01
Discard response after GATT operation cancellation

### LINKED ISSUE #326: Cancelling a read operation
I'm not sure if this is caused by Kable or the Android platform. If I cancel a coroutine while it's running a read operation, any attempt to read afterwards immediately fails with a `GattRequestRejectedException` (`message` is `null`). This lasts until I restart the connection. I haven't dared to test the same thing with a write operation 😅

Is this the expected behavior, or a bug in Kable?

My use case is I am reading a characteristic periodically while a specific page is showing, and I want to cancel that polling when navigating away from that page, while keeping the connection open for the next page.

twyatt:
Closes #326


twyatt:
Note to self: fix will likely also need to be applied to Apple platform.

twyatt:
> Hey @twyatt 🙂 Any progress on this issue?

Apologies, I haven't found the time to test it as much as I'd like.

I'll finish the integration on Apple soon, then release a `-beta` version (as a kind of notice to library users that it could use additional testing).

core/src/androidMain/kotlin/Connection.kt
```diff
@@ -52,12 +58,25 @@ internal class Connection(
...
+        }
+
         withContext(dispatcher) {
+            pending = true
             action.invoke(bluetoothGatt) || throw GattRequestRejectedException()
         }
 
         val response = try {
-            callback.onResponse.receive()
+            callback.onResponse.receive().also { pending = false }
```
cedrickcooke, line 79: Should this be a `finally`?

twyatt: No, as I only want to "reset" the `pending` state if we successfully received a response.
If the Coroutine was cancelled then I want `pending` to remain `true` for the next pass.

I wasn't excited about carrying state across multiple request/response operations, but there isn't a way to clear a `Channel` (that I am aware of); which is what I would've done in a `catch (e: CancellationException)` block if that were possible.

cedrickcooke (approved):

sdonn3 (approved):

twyatt:
@joharei this PR is ready (with necessary approvals) but I'm waiting to merge. I'll merge/release 0.20.1 (with these changes) shortly after [0.20.0](https://github.com/JuulLabs/kable/milestone/30) is released.

QuantumRand (approved):

twyatt:
> @joharei this PR is ready (with necessary approvals) but I'm waiting to merge. I'll merge/release 0.20.1 (with these changes) shortly after [0.20.0](https://github.com/JuulLabs/kable/milestone/30) is released.

Sorry, slight delays in this shipping due to other changes getting in. Will release a version with this change in it very soon.

PR #392, opened 2022-10-17
Add support for scan options

robvs:
The Apple implementation of `Scanner.advertisements` uses `CBCentralManager.scanForPeripherals(withServices:options:)` to start scanning for advertisements. The Kable implementation currently passes `null` for `options`. In my application, I need to set the option `CBCentralManagerScanOptionAllowDuplicatesKey` to `true` so that my app sees all advertisements. (this option is set to `false` by default)

This change enables peripheral scanning options to be passed to `scanForPeripherals()`. After reviewing this with @mtrewartha and discussing whether the `scanOptions` property added here could be helpful on other platforms, we did not see a clear path for it due to the individual quirks of each platform .

### DIFF
--- a/core/src/appleMain/kotlin/ScannerBuilder.kt
+++ b/core/src/appleMain/kotlin/ScannerBuilder.kt
@@ -3,6 +3,8 @@ package com.juul.kable
 import com.benasher44.uuid.Uuid
 import com.juul.kable.logs.Logging
 import com.juul.kable.logs.LoggingBuilder
+import platform.CoreBluetooth.CBCentralManagerScanOptionAllowDuplicatesKey
+import platform.CoreBluetooth.CBCentralManagerScanOptionSolicitedServiceUUIDsKey
 
 public actual class ScannerBuilder {
     @Deprecated(
@@ -17,7 +19,18 @@ public actual class ScannerBuilder {
 
     public actual var filters: List<Filter>? = null
 
-    public var scanOptions: AppleScanner.ScanOptions? = null
+    /**
+     * Specifies whether the scan should run without duplicate filtering. This corresponds to
+     * Apple's CBCentralManagerScanOptionAllowDuplicatesKey scanning option.
+     */
+    public var allowDuplicateKeys: Boolean? = null
+
+    /**
+     * Causes the scanner to scan for peripherals soliciting any of the services contained in the
+     * array. This corresponds to Apple's CBCentralManagerScanOptionSolicitedServiceUUIDsKey
+     * scanning option.
+     */
+    public var solicitedServiceUuids: List<Uuid>? = null
 
     private var logging: Logging = Logging()
 
@@ -25,10 +38,20 @@ public actual class ScannerBuilder {
         logging = Logging().apply(init)
     }
 
-    internal actual fun build(): Scanner = AppleScanner(
-        central = CentralManager.Default,
-        services = filters?.filterIsInstance<Filter.Service>()?.map { it.uuid },
-        options = scanOptions,
-        logging = logging,
-    )
+    internal actual fun build(): Scanner {
+        val options = mutableMapOf<String, Any>()
+        allowDuplicateKeys?.also {
+            options[CBCentralManagerScanOptionAllowDuplicatesKey] = it
+        }
+        solicitedServiceUuids?.also {
+            options[CBCentralManagerScanOptionSolicitedServiceUUIDsKey] = it.toTypedArray()
+        }
+
+        return AppleScanner(
+            central = CentralManager.Default,
+            services = filters?.filterIsInstance<Filter.Service>()?.map { it.uuid },
+            options = options.toMap(),
+            logging = logging,
+        )
+    }
 }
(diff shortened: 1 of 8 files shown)


twyatt:
Thanks for the PR, I'll try to review this by this weekend.

robvs:
fyi, the added `scanOptions` property is optional when building an `AppleScanner` and therefore should have no effect on existing implementations.

core/src/appleMain/kotlin/ScannerBuilder.kt
```diff
@@ -16,6 +16,14 @@ public actual class ScannerBuilder {
         get() = filters?.filterIsInstance<Filter.Service>()?.map { it.uuid }
 
     public actual var filters: List<Filter>? = null
+
+    /**
+     * Scanning options for Apple's CBCentralManager.scanForPeripherals(withServices:options:).
+     * This enables peripheral scanning options such as CBCentralManagerScanOptionAllowDuplicatesKey
+     * to be specified when scanning begins.
+     */
+    public var scanOptions: Map<Any?, *>? = null
```
twyatt, line 25: To make the options more discoverable (and less error-prone) it might make sense to have this be a `data class` that defines the available options. Then internally, it would be converted to the `Map<String, *>` that Core Bluetooth is expecting?

For example:

```kotlin
data class ScanOptions(
    val allowDuplicateKeys: Boolean,
    val solicitedServiceUuids: Array<Uuid>,
)
```

Then, in the builder we would have something along the lines of:

```suggestion
    /**
     * Scanning options for Apple's CBCentralManager.scanForPeripherals(withServices:options:).
     * This enables peripheral scanning options such as CBCentralManagerScanOptionAllowDuplicatesKey
     * to be specified when scanning begins.
     */
    public var options: ScanOptions? = null
```

Alternatively, since there are so few options, might make sense to have them be properties directly on the Scanner builder; I'm not quite sure which I like better.

What do you think?

robvs:
@twyatt I prefer your first suggestion of using a `data class` because it feels cleaner to me. Thanks for the review.

@mtrewartha please review my refactoring if you get a chance.

twyatt:
Thought about it some more; it'd be even nicer to provide a DSL, for example:

```kotlin
Scanner {
    options {
        allowDuplicateKeys = true // or false
        solicitedServiceUuids = arrayOf(..) // or null
    }
}
```

With a DSL, any of the options are optional (and if not specified, we can set to defaults).

I think the method signature on the scanner builder could look something like:

```kotlin
private typealias OptionsBuilder = ScannerBuilder.Options.() -> Unit

class ScannerBuilder {

    public data class Options internal constructor(
        var allowDuplicateKeys: Boolean? = null,
        var solicitedServiceUuids: List<Uuid>? = null,
    )

    // ..

    private var options: Options = Options()
    public fun options(builder: OptionsBuilder) {
        options = Options().apply(builder)
    }

    internal actual fun build(): Scanner {
        val scanOptions = mutableMapOf<String, Any>()
        options.allowDuplicateKeys?.also {
            scanOptions[CBCentralManagerScanOptionAllowDuplicatesKey] = it
        }
        options.solicitedServiceUuids?.also {
            scanOptions[CBCentralManagerScanOptionSolicitedServiceUUIDsKey] = it.toTypedArray()
        }
        return AppleScanner(
            // ..
        )
    }
}
```

core/src/appleMain/kotlin/ScannerBuilder.kt
```diff
@@ -17,18 +19,39 @@ public actual class ScannerBuilder {
 
     public actual var filters: List<Filter>? = null
 
-    public var scanOptions: AppleScanner.ScanOptions? = null
+    /**
+     * Specifies whether the scan should run without duplicate filtering. This corresponds to
+     * Apple's CBCentralManagerScanOptionAllowDuplicatesKey scanning option.
```
twyatt, line 24: nitpick:
- "Core Bluetooth" is slightly more specific 🤷 
- brackets around the constants make them clickable in IDE (and might allow Dokka to link them in the generated KDoc documentation)

```suggestion
     * Core Bluetooth's [CBCentralManagerScanOptionAllowDuplicatesKey] scanning option.
```

core/src/appleMain/kotlin/ScannerBuilder.kt
```diff
@@ -17,18 +19,39 @@ public actual class ScannerBuilder {
...
+     */
+    public var allowDuplicateKeys: Boolean? = null
+
+    /**
+     * Causes the scanner to scan for peripherals soliciting any of the services contained in the
+     * array. This corresponds to Apple's CBCentralManagerScanOptionSolicitedServiceUUIDsKey
```
twyatt, line 30: nitpick

```suggestion
     * array. This corresponds to Core Bluetooth's [CBCentralManagerScanOptionSolicitedServiceUUIDsKey]
```

twyatt (approved):
Looks great to me!
I don't have anyway to test these new options, so I'd appreciate any testing you're willing to do.
Thank You!

cedrickcooke (approved):

twyatt:
Thanks for the contribution! It will be released with [0.20.0](https://github.com/JuulLabs/kable/milestone/30).

PR #395, opened 2022-11-01
Expose Peripheral identifier

### LINKED ISSUE #394: Expose Peripheral's identifier
I believe there is a need to expose the Peripheral's identifier which is not public at this point of time.

There are some usecases (like when using multiple BLE libraries that do different things, that's a usecase I came across) where the identifier is needed and it's not exposed in the `Peripheral` object.

PavlosTze:
Closes #394


cedrickcooke (approved):

PavlosTze:
I just realized this is available already. I'm sorry, I must have missed it. Therefore I am closing this PR.

Issue #156, opened 2021-08-31
Bluetooth device name on iOS is not updated


ln-12:
I am developing a bluetooth device which my app should connect to. During testing, I changed the name of my bluetooth device. On Android, it works absolutely fine and as expected. However, as iOS always does things differently, it does not work.  In fact, iOS caches bluetooth names according to this [SO answer](https://stackoverflow.com/a/47129088/13440564). The author mentions a different approach  of getting the real (updated) name which could be used [here](https://github.com/JuulLabs/kable/blob/main/core/src/appleMain/kotlin/Advertisement.kt#L23). What do you think?

twyatt:
I think this is either a duplicate of, or closely related to, #143.

In either case, yes, we need to make some changes there to make the uncached name available.
We've been swamped with some deadlines lately, but I hope to be able to put some more time to Kable in the coming weeks.

ln-12:
Alright, I'll close this issue in favor of #143. Thanks!

Issue #143, opened 2021-07-14
Consider exposing uncached ble device name


BluestormDNA:
Hello there!

Not sure if related to #59 so i opened a new issue, feel free to close it if it actually is.

Because Ble is so weird, at least on Android (and probably on iOS too) the name returned by the peripheral is the name the BLE Device had when it connected the first time. Then is cached internally (and is reseted on certain conditions...)

This is usually not a problem, but on some scenarios BLE Devices change names to signal things and you actually need the name the device is actually advertising and not the one you connected to originally. On Android that uncached name is on `scanRecord?.deviceName?` and on iOS is on `kCBAdvDataLocalName` but at least on android that field on Advertisement is private.

Even thought i can get them with reflection it could be nice to maybe have something like an `advertisedName` field or actually the scanRecord data exposed.

twyatt:
On Android, is [`scanRecord.deviceName`](https://developer.android.com/reference/android/bluetooth/le/ScanRecord#getDeviceName()) more accurate than [`BluetoothDevice.name`](https://developer.android.com/reference/android/bluetooth/BluetoothDevice#getName())? Is that what you're saying?

If so, then perhaps it would make sense to have the following:

https://github.com/JuulLabs/kable/blob/f34f54e0a9125a637f6ea4615b4028ccf0a1f3b8/core/src/androidMain/kotlin/Advertisement.kt#L24-L25

...instead be written as?:

```kotlin
public actual val name: String?
    get() = scanRecord?deviceName
```

Or, are there use-cases for using one of the names vs. the other?

Devinete:
> On Android, is scanRecord.deviceName more accurate than BluetoothDevice.name? Is that what you're saying?

For scanning purpose it actually is. It haves the current emitted deviceName wich may be the same OR NOT of what you have cached on BluetoothDevice.name

This is some niche use case and relies on ble devices that change names. (My use case itself is a device that acts as a bridge between a server and a client to send data and it changes names dinamically, conserving the same mac, to expose to clients or server...)

Consider a BLE device that changes its name every 24 hours and it starts as "Monday Device" then "Tuesday Device" and so on...
If you happen to connect to it the first time on Wednesday. The bluetoothDevice.name will be "Wednesday Device".
If you come back on Sunday, assuming here the ble hasn't invalidated the cache on reboot or magic, bluetoothDevice.name will still be "Wednesday Device" and scanRecord?deviceName will actually have "Sunday Device".

[...]

PR #401, opened 2022-11-12
Return device name from the scan record when available on Android

L3K0V:
TLDR: Check and return device name from the advertisement scan record if available. If not return device name as before.

[As I wrote in the discussions](https://github.com/JuulLabs/kable/discussions/400) I had some issues with iBeacon devices. I noticed that on Apple devices I got more advertisements having names than the Android results. I start digging a bit and investigating and stumble upon this:
https://stackoverflow.com/questions/26290640/android-bluetoothdevice-getname-return-null and more especially:

> Advertisement data is how a client BTLE endpoint discovers a service device. A client can request a scan response and get more data. The device name is optional in this data. However, the BTLE spec requires that all Bluetooth Low Energy endpoints support the Generic Access service which is required to support the Device Name characteristic.
> ...
> Fortunately, most BTLE devices I have worked with do provide their name in the advertisement or scan response.
>
> Another possibility is that the device may place the name in the scan response part of the advertisement....

And

> On Marshmallow, utilize `ScanRecord.getDeviceName()` to retrieve the local name embedded in the scan record.
>
> `BluetoothDevice.getName()` is unreliable if the local name is included in a scan response, rather than in the immediate advertising packet.

### DIFF
--- a/core/src/androidMain/kotlin/Advertisement.kt
+++ b/core/src/androidMain/kotlin/Advertisement.kt
@@ -22,7 +22,7 @@ public actual class Advertisement(
         get() = scanResult.device
 
     public actual val name: String?
-        get() = bluetoothDevice.name
+        get() = scanResult.scanRecord?.deviceName ?: bluetoothDevice.name
 
     public val address: String
         get() = bluetoothDevice.address


cedrickcooke (approved):

twyatt:
@L3K0V this looks like a good change, my only concern is if there are circumstances where `scanResult.scanRecord?.deviceName` value differs from `bluetoothDevice.name`? As this change could break library consumers if they were relying on a specific name (as returned by `bluetoothDevice.name`).

Normally I'd recommend that we'd deprecate the current property (and offer a new property), so that we can warn library consumers (via the deprecation) that the value may differ from prior versions, but:

- I'd imagine they should always be the same, or at least `scanResult.scanRecord?.deviceName` might hold a value when `bluetoothDevice.name` is `null`? — so, it'd be unlikely consumers would have a problem with `Advertisement.name` suddenly returning a non-`null` value
- Kable is still pre 1.0.0, so hopefully consumers understand that APIs could change/break with each version (as much as I'd like to avoid that)

...so, I think this is fine; but if you happen to know of instances where `scanResult.scanRecord?.deviceName` and `bluetoothDevice.name` might both hold a non-`null` value (but be different) then we should maybe consider a deprecation path for this.

L3K0V:
Hey @twyatt,

> I'd imagine they should always be the same, or at least scanResult.scanRecord?.deviceName might hold a value when bluetoothDevice.name is null? — so, it'd be unlikely consumers would have a problem with Advertisement.name suddenly returning a non-null value

Correct.

> but if you happen to know of instances where `scanResult.scanRecord?.deviceName` and `bluetoothDevice.name` might both hold a non-null value (but be different) then we should maybe consider a deprecation path for this.

From what I know on Android via Settings user can change the Bluetooth device name (`bluetoothDevice.name`) which again can broke some Kable consumers, but the name represented in the `scanRecord` is controlled by the manufacturer and the firmware. Notice that in my case if I factory reset the bluetooth device the `scanRecord` name will be changed, as it's suffixed with an unique ID.

Also in my example looks that `bluetoothDevice.name` returns `null` on Android  but in nRF Connect tool they show the `scanRecord` name. Right now I'm not able to filter via Kable - it's working on iOS but not on Android. On iOS in all cases I have the device name.

twyatt (approved):

L3K0V:
@twyatt btw, any deprecation path defined for this? I mean how we proceed with this?

BluestormDNA:
@twyatt 
I saw this issue very late, but this was already reported on https://github.com/JuulLabs/kable/issues/143 (Android / iOS) and https://github.com/JuulLabs/kable/issues/156 (iOS)
And as said there I think it would be more reasonable to have both properties exposed. The cached name and the current advertised name.

Remember that them can and will be different (see https://github.com/JuulLabs/kable/issues/143) example on a device that changes name every 24h.
And users can rely on both of them depending on use cases. Fusing them together it's not a solution.

Exposing both of them would cover all the bases and @L3K0V could have his use case covered with something like:
`advertisedName ?: cachedName`

twyatt:
Thanks @BluestormDNA for catching that and linking to the existing issues. I'll change to expose as two separate properties. 💯

L3K0V:
Many thanks for the new information here @BluestormDNA much appreciated as I was trying to think about other use cases.

So after all @twyatt looks like it will be needed to expose both names, I can help here.

First - Is there similar problem on iOS? Shall we expose a common field or Android only field?
And I wonder how to explain to the users which one to use and when. I was thinking that `scanRecord` name is more likely to be the single source of truth.

I like having `advertisedName` and `localDeviceName`. Can be call the cached name local like local for the device?

PR #403, opened 2022-11-21
Retrieve `Advertisement.name` from advertisement data

### LINKED ISSUE #143: Consider exposing uncached ble device name

twyatt:
Adjusts the value returned by `Advertisement.name` that was recently changed (unreleased) via #401.

As @BluestormDNA kindly pointed out in https://github.com/JuulLabs/kable/pull/401#issuecomment-1319638377, `Advertisement.name` was not sourcing the name from the most up-to-date sources (and that Kable should provide the uncached value, as well as the previously provided cached value).

The `Advertisement.name` now pulls the name from advertisement data. Previously it used the value provided by the platform peripheral object (but in many cases that value comes from cache and may be outdated).

The previous behavior (data source for peripheral name) is now available as `Advertisement.peripheralName: String?`.

Closes #143


davidtaylor-juul (approved):

L3K0V:
Hello, is there are snapshot or nightly version I can test meanwhile this is being merge?

twyatt:
> Hello, is there are snapshot or nightly version I can test meanwhile this is being merge?

```kotlin
repositories {
}

dependencies {
    implementation("com.juul.kable:core:0.20.1-issue-403-1-SNAPSHOT")
}
```

L3K0V:
Hey @twyatt unfortunately this snapshot does not work for me. It looks like the issue with the iBeacon persists.

twyatt:
> Hey @twyatt unfortunately this snapshot does not work for me. It looks like the issue with the iBeacon persists.

Are you seeing that `Advertisement.name` is providing cached value rather than what you expect?
Does `Advertisement.peripheralName` provide a different value (this is the old behavior)?

L3K0V:
> Are you seeing that Advertisement.name is providing cached value rather than what you expect?
> Does Advertisement.peripheralName provide a different value (this is the old behavior)?

TLDR: Unfortunately I cannot validate as all devices around me report same scan and cached name, still I have some missing devices on Android which are available as scan results on iOS.


##  Details

> Are you seeing that Advertisement.name is providing cached value rather than what you expect?
> Does Advertisement.peripheralName provide a different value (this is the old behavior)?

Hard to say right now. For some devices the values are same, for another I got strange logs:

Android:
```
Name(null): PeripheralName(null)
Name(��): PeripheralName(null)
Name(BM1002_cb0275): PeripheralName(BM1002_cb0275)
Name(BM1002_b0a5e4): PeripheralName(BM1002_b0a5e4)
```

Unfortunately I cannot distinguish them just like that....

On iOS I got slightly different results tho:

```
Name(null): PeripheralName(null)
Name(BM1002_b0a5e4): PeripheralName(BM1002_b0a5e4)
Name(BM1002_cb0275): PeripheralName(BM1002_cb0275)
Name(Nuki_2546AC65): PeripheralName(Nuki_2546AC65)
Name(SBF77): PeripheralName(SBF77)
``` 

`Name(Nuki_2546AC65): PeripheralName(Nuki_2546AC65)` this is the iBeacon I'm looking for.

These two are even not coming on Android:

```
Name(Nuki_2546AC65): PeripheralName(Nuki_2546AC65)
Name(SBF77): PeripheralName(SBF77)
```

twyatt:
@L3K0V for a sanity check, can you see if your iBeacon is showing up on Android using [nRF Connect](https://play.google.com/store/apps/details?id=no.nordicsemi.android.mcp&hl=en_US&gl=US&pli=1)?

L3K0V:
Okay.... mystery is resolved @twyatt.

What I can tell is `name` and `peripheralName` reports the same values.

I wonder if it is needed to expose two properties to be honest as it looks like the problem I had unfortunately was very different. I got completely lost in these bluetooth stuff so maybe we should consider having just single name property.

The problem was, that because of the Android permissions changes across API versions and my permissions declarations in the Manifest, all beacons ware skipped from the scan results. Once I change the permissions declared I'm able to see my target device. I will update and put some answer on the discussion as well. 

Let me know how we can proceed here and how I can help. Do we need to revert `peripheralName` in such case?

twyatt:
Thanks for the follow up @L3K0V!

Although somewhat anecdotal, various places on the internet hint that there may be subtle differences between the two name properties; I'm thinking we keep them both, but I'll revert `name` to be what it was in Kable 0.20.1, then an alternate name can be chosen for the property that provides the "name" that comes from the scan record.

In other words, I'll swap the data source of the names as they are currently in this PR (so that `name` keeps its old behavior).

sdonn3 (approved):

twyatt:
@L3K0V I ended up having `Advertisement.name` come from the scan record and `Advertisment.peripheralName` come from the (sometimes cached) bluetooth device object's `name` property.

This means that the value of `Advertisement.name` will potentially change for some library consumers, but I feel like the scan record is more aligned with what would be expected from the `Advertisement` API; and thus the potentially breaking change (for some).

PR #396, opened 2022-11-03
Add name and address filter in Android

PavlosTze:
Implements #377

PR #408, opened 2023-01-03
Rework filter support to have a common API on all platforms

### LINKED ISSUE #377: [Suggestion] Add device name on Scanner Filters
Right now, on the Filters we can only use `Services UUID` and `ManufacturerData`. I think it's important to be able to filter by device name (or at least a prefix of a device name). 

Is that feasible?

twyatt:
Moves and provides all `Filter`s to common code, to provide a uniform scanning experience on all platforms. Uses native filtering support (when available) and falls back to leveraging flow filtering (when not natively supported).

All `Filter`s are supported on all platforms, except for `Filter.Address`, which is only supported on Android.

| Filter           | Android | Apple | JavaScript |
|------------------|:-------:|:-----:|:----------:|
| `Service`          |   ✓✓    |  ✓✓*  |     ✓✓     |
| `Name`             |   ✓✓    |   ✓   |     ✓✓     |
| `NamePrefix`       |    ✓    |   ✓   |     ✓✓     |
| `Address`          |   ✓✓    |       |            |
| `ManufacturerData` |   ✓✓    |   ✓   |     ✓✓     |

✓✓ = Supported natively
✓ = Support provided by Kable via flow filter
✓✓* = Supported natively if the only filter type used, otherwise falls back to flow filter

[Documentation](https://github.com/JuulLabs/kable/tree/twyatt/filter#scanning) has been updated to reflect these changes.

_Diff shows changes from #396; will rebase after #396 is merged to clean up the PR changes._

Closes #377.


twyatt:
Note to self: escalate deprecations related to scanning/filtering.

core/src/androidMain/kotlin/Scanner.kt
```diff
@@ -95,5 +103,11 @@ public class AndroidScanner internal constructor(
                 logger.warn(e) { message = "Failed to stop scan. " }
             }
         }
+    }.filter { advertisement ->
+        // Short-circuit (i.e. don't filter) if no `Filter.NamePrefix` filters were provided.
+        if (namePrefixFilters.isEmpty()) return@filter true
+
+        // Perform `Filter.NamePrefix` filtering here, since it isn't supported natively.
+        namePrefixFilters.any { filter -> filter.matches(advertisement.name) }
```
ebabel, line 111: thought this was a bug until I realized you have `.startsWith` further down

twyatt: I guess I was shooting for a common `interface` feel for `Filter`, whereas they all internally could have a `matches` function (but with different parameter types — which makes the `interface` infeasible). 🤷

core/src/jsMain/kotlin/Bluetooth.kt
```diff
@@ -20,11 +20,11 @@ public fun CoroutineScope.requestPeripheral(
     options: Options,
     builderAction: PeripheralBuilderAction = {},
 ): Promise<Peripheral> = bluetooth
-    .requestDevice(options.toDynamic())
+    .requestDevice(options.toRequestDeviceOptions().also { console.info(it) })
```
davidtaylor-juul, line 23: We should use the logger instead of spamming direct to console

core/src/jsMain/kotlin/Scanner.kt
```diff
@@ -20,41 +22,47 @@ private const val ADVERTISEMENT_RECEIVED_EVENT = "advertisementreceived"
...
+    private val options = filters.toBluetoothLEScanOptions()
+
     public override val advertisements: Flow<Advertisement> = callbackFlow {
         check(supportsScanning) { "Scanning unavailable" }
 
-        val scan = bluetooth.requestLEScan(options.toDynamic()).await()
-        val listener: (Event) -> Unit = {
-            trySend(Advertisement(it as BluetoothAdvertisingEvent)).getOrElse {
+        logger.info { message = "Starting scan" }
+        console.info(options)
```
davidtaylor-juul, line 40: Duplicate logging direct to `console` should be removed

core/src/jsMain/kotlin/Scanner.kt
```diff
@@ -20,41 +22,47 @@ private const val ADVERTISEMENT_RECEIVED_EVENT = "advertisementreceived"
...
+        val scan = bluetooth.requestLEScan(options).await()
+
+        console.info("Configuring listener")
```
davidtaylor-juul, line 43: Prefer logger over direct to `console`

core/src/jsMain/kotlin/Scanner.kt
```diff
@@ -20,41 +22,47 @@ private const val ADVERTISEMENT_RECEIVED_EVENT = "advertisementreceived"
...
+        val listener: (Event) -> Unit = { event ->
+            console.info(event)
```
davidtaylor-juul, line 45: Prefer logger over direct to `console`

core/src/jsMain/kotlin/Scanner.kt
```diff
@@ -20,41 +22,47 @@ private const val ADVERTISEMENT_RECEIVED_EVENT = "advertisementreceived"
...
+            trySend(Advertisement(event.unsafeCast<BluetoothAdvertisingEvent>())).getOrElse {
                 console.warn("Unable to deliver advertisement event due to failure in flow or premature closing.")
```
davidtaylor-juul, line 47: Prefer logger over direct to `console`

core/src/jsMain/kotlin/Scanner.kt
```diff
@@ -20,41 +22,47 @@ private const val ADVERTISEMENT_RECEIVED_EVENT = "advertisementreceived"
...
             }
         }
         bluetooth.addEventListener(ADVERTISEMENT_RECEIVED_EVENT, listener)
+        console.info("Listener configured")
```
davidtaylor-juul, line 51: Prefer logger over direct to `console`

core/src/commonMain/kotlin/Filter.kt
```diff
@@ -1,10 +1,125 @@
...
+        /**
+         * For any bit in the mask, set it to 1 if it needs to match the corresponding bit in [data], otherwise set it
+         * to 0. If [dataMask] is not `null`, then it must have the same length as [data].
+         */
+        public val dataMask: ByteArray?,
+    ) : Filter() {
+        init {
+            if (dataMask != null) checkDataAndMask(data, dataMask)
+        }
+    }
```
davidtaylor-juul, line 96: Maybe the comment is throwing me off, but I got very confused that a `ByteArray` is a bit-mask?

I'm not familiar with all the use cases, but it makes me wonder if this Is too specialized and perhaps the raw lambda comparator should be exposed for this `ManufacturerData` filter API instead?

davidtaylor-juul (changes requested):
Great job on this. Marking as changes-requested pending question around the API but otherwise looks ready to go.

core/src/jsMain/kotlin/Bluetooth.kt
twyatt: Oh yikes. Good catch! All the `console` logs were for debugging that I forgot to remove (or replace with `Log.*` entries). Will fix.

core/src/commonMain/kotlin/Filter.kt
twyatt: The lamba comparator is a cool idea, but strangely the bytes as a mask is the paradigm used natively for both Android and JavaScript. 🤷 

I'm inclined to get the feature in as the byte mask for now (since it is likely familiar for advanced users who have used the feature natively on either Android or JS) — then maybe sometime in the future add additional helpers that provide the lambda approach?

twyatt: I can rewrite/update the comment if you think that would be helpful?

twyatt: @davidtaylor-juul I've updated the documentation, let me know if that clarifies things?

https://github.com/JuulLabs/kable/blob/3bd19ddab03b9025268989668b017d1652f376c5/core/src/commonMain/kotlin/Filter.kt#L72-L109

davidtaylor-juul: > the bytes as a mask is the paradigm used natively for both Android

Was not aware of this, and I also mis-read the comparison logic on first pass :-/  Makes perfect sense now.

davidtaylor-juul (approved):

core/src/appleMain/kotlin/Scanner.kt
```diff
@@ -16,33 +18,78 @@ import platform.CoreBluetooth.CBManagerStateUnsupported
...
+                    central.scanForPeripheralsWithServices(null, options)
+                }
             }
             .onCompletion {
+                logger.info { message = "Stopping scan" }
                 central.stopScan()
             }
             .filterIsInstance<DidDiscoverPeripheral>()
+            .filter { didDiscoverPeripheral ->
+                if (nativeServiceFiltering) return@filter true // Short-circuit (i.e. don't filter) when scan is using native service filtering.
```
sdonn3, line 69: Why is this short-circuiting done? Maybe I dont understand "native service filtering"

sdonn3:
Awesome changes, love the comments as always, makes it so much easier to follow

sdonn3 (approved):

core/src/appleMain/kotlin/Scanner.kt
twyatt: Native service filtering means that the OS is performing the filtering, so we short-circuit (return `true`) in flow `filter` operator so that we let all the events through the flow (since the OS will have already done the filtering for us).

PR #416, opened 2023-01-10
Simplify documentation re: Android permissions setup

twyatt:
Adjusted documentation to be in table format (to be more organized/readable).

Rendered markdown [here](https://github.com/JuulLabs/kable/tree/twyatt/android-permissions-doc#android-permissions).

### DIFF
--- a/README.md
+++ b/README.md
@@ -454,32 +454,88 @@ cancellation is provided to prevent connection leaks._
 
 ## Setup
 
-### Android
+### Android Permissions
 
-Kable declares common bluetooth permissions but doesn't declare that bluetooth hardware is required. If your app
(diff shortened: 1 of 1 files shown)


davidtaylor-juul (approved):
Is there is something special here that requires the HTML syntax over Markdown? See https://www.markdownguide.org/extended-syntax/#tables

I don't have s strong opinion either way, so fine as is too.

twyatt:
> Is there is something special here that requires the HTML syntax over Markdown? See https://www.markdownguide.org/extended-syntax/#tables
> 
> I don't have s strong opinion either way, so fine as is too.

Multiline code blocks don't work with markdown tables. 😢

QuantumRand (approved):

PR #441, opened 2023-03-08
Extract exceptions module

cedrickcooke:
Extracts common `Exception` classes into a separate jvm-without-android friendly module. Exposure via `api` in gradle should make this binary-breaking but not source-breaking.


exceptions/src/commonMain/kotlin/Exceptions.kt
```diff
@@ -1,42 +1,42 @@
 package com.juul.kable
 
 /** Failure occurred with the underlying Bluetooth system. */
-public open class BluetoothException internal constructor(
```
twyatt, line 4: Is there currently a need to make the constructor public?

cedrickcooke: Yes, since we also construct one as the superclass of `GattRequestRejectedException` which is still in `kable-core`

core/src/androidMain/kotlin/GattRequestRejectedException.kt
```diff
@@ -1,13 +1,9 @@
-@file:JvmName("AndroidExceptionsKt") // Removes conflict with commonMain Exceptions.kt
```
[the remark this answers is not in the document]
cedrickcooke: Especially not needed now that the filename changed 😄

twyatt (approved):

PR #447, opened 2023-03-19
Introduce platform specific `Peripheral` interfaces

twyatt:
No change in functionality, just changing how methods are exposed (via `interface` now rather than concrete classes).

Took the opportunity to rename JavaScript and Apple interfaces to names that more appropriately represent the framework they're built on top of.

### DIFF
--- a/core/src/androidMain/kotlin/AndroidPeripheral.kt
+++ b/core/src/androidMain/kotlin/AndroidPeripheral.kt
@@ -1,0 +1,26 @@
+package com.juul.kable
+
+import kotlinx.coroutines.flow.StateFlow
+
+public interface AndroidPeripheral : Peripheral {
+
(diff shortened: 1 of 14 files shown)


davidtaylor-juul (approved):

core/src/androidMain/kotlin/AndroidPeripheral.kt
```diff
@@ -0,0 +1,26 @@
+package com.juul.kable
+
+import kotlinx.coroutines.flow.StateFlow
+
+public interface AndroidPeripheral : Peripheral {
```
ebabel, line 5: could be nice to have address on here to :)

ebabel, line 5: @twyatt could be nice to have address on here to :)

PR #467, opened 2023-04-16
Bluetooth availability - monitor location service changes

mmaleiter:
I want to mention two things in advance.

Starting with is the usage of distinctUntilChange(). The used location filter hits more than once when the location tracking is enabled/disabled. Since we are only interested in this change we only emit on enable/disable change.

The second thing is using a function instead of property for isLocationEnabled. Sometimes, also in this context, the usage of property does not work reliably. I also read about that somewhere but cannot remember when and where.

### DIFF
--- a/core/src/androidMain/kotlin/Bluetooth.kt
+++ b/core/src/androidMain/kotlin/Bluetooth.kt
@@ -43,17 +43,18 @@ private val Context.locationManager: LocationManager
     get() = ContextCompat.getSystemService(this, LocationManager::class.java)
         ?: error("LocationManager system service unavailable")
 
-private fun isLocationEnabled(): Boolean {
-    return LocationManagerCompat.isLocationEnabled(applicationContext.locationManager)
-}
+private val isLocationEnabled: Boolean
+    get() = LocationManagerCompat.isLocationEnabled(applicationContext.locationManager)
 
 private val locationEnabledFlow = when {
     SDK_INT > R -> flowOf(true)
-    SDK_INT == R -> broadcastReceiverFlow(IntentFilter(PROVIDERS_CHANGED_ACTION))
-        .map { intent -> intent.getBooleanExtra(EXTRA_PROVIDER_ENABLED, false) }
-        .onStart { emit(isLocationEnabled()) }
+    else -> broadcastReceiverFlow(IntentFilter(PROVIDERS_CHANGED_ACTION))
+        .map { intent ->
+            if (SDK_INT == R) intent.getBooleanExtra(EXTRA_PROVIDER_ENABLED, false)
+            else isLocationEnabled
+        }
+        .onStart { emit(isLocationEnabled) }
         .distinctUntilChanged()
-    else -> flowOf(isLocationEnabled())
 }
 
 private val bluetoothStateFlow =
--- a/core/src/androidMain/kotlin/Bluetooth.kt
+++ b/core/src/androidMain/kotlin/Bluetooth.kt
@@ -87,4 +87,3 @@ internal actual val bluetoothAvailability: Flow<Bluetooth.Availability> =
     ) { locationEnabled, bluetoothState ->
         if (locationEnabled) bluetoothState else Unavailable(reason = LocationServicesDisabled)
     }
-
--- a/core/src/androidMain/kotlin/Bluetooth.kt
+++ b/core/src/androidMain/kotlin/Bluetooth.kt
@@ -21,9 +21,9 @@ import com.juul.kable.Reason.Off
 import com.juul.kable.Reason.TurningOff
 import com.juul.kable.Reason.TurningOn
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.onStart
 import android.bluetooth.BluetoothAdapter.ACTION_STATE_CHANGED as BLUETOOTH_STATE_CHANGED
@@ -47,7 +47,7 @@ private fun isLocationEnabled(): Boolean {
 
 private val locationNotNeededStateFlow: Flow<Boolean> =
     when {
-        SDK_INT > R -> MutableStateFlow(true)
+        SDK_INT > R -> flow { emit(true) }
         else -> broadcastReceiverFlow(IntentFilter(LocationManager.PROVIDERS_CHANGED_ACTION))
             .map { intent ->
                 intent.getBooleanExtra(LocationManager.EXTRA_PROVIDER_ENABLED, false)
--- a/core/src/androidMain/kotlin/Bluetooth.kt
+++ b/core/src/androidMain/kotlin/Bluetooth.kt
@@ -10,6 +10,8 @@ import android.bluetooth.BluetoothAdapter.STATE_TURNING_ON
 import android.content.Context
 import android.content.IntentFilter
 import android.location.LocationManager
+import android.location.LocationManager.EXTRA_PROVIDER_ENABLED
+import android.location.LocationManager.PROVIDERS_CHANGED_ACTION
 import android.os.Build.VERSION.SDK_INT
 import android.os.Build.VERSION_CODES.R
 import androidx.core.content.ContextCompat
@@ -23,7 +25,7 @@ import com.juul.kable.Reason.TurningOn
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.distinctUntilChanged
-import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.flowOf
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.onStart
 import android.bluetooth.BluetoothAdapter.ACTION_STATE_CHANGED as BLUETOOTH_STATE_CHANGED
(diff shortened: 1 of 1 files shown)


core/src/androidMain/kotlin/Bluetooth.kt
```diff
@@ -38,10 +41,21 @@ private val Context.locationManager: LocationManager
...
+    return LocationManagerCompat.isLocationEnabled(applicationContext.locationManager)
+}
 
-internal actual val bluetoothAvailability: Flow<Bluetooth.Availability> =
+private val locationNotNeededStateFlow: Flow<Boolean> =
+    when {
+        SDK_INT > R -> flow { emit(true) }
+        else -> broadcastReceiverFlow(IntentFilter(LocationManager.PROVIDERS_CHANGED_ACTION))
+            .map { intent ->
+                intent.getBooleanExtra(LocationManager.EXTRA_PROVIDER_ENABLED, false)
```
twyatt, line 53: Unfortunately [`LocationManager.EXTRA_PROVIDER_ENABLED`](https://developer.android.com/reference/android/location/LocationManager#EXTRA_PROVIDER_ENABLED) requires API 30.

Being that we don't have to check location services status after API 30, this really only adds any value for Android 11 (API 30). If we're below API 30, then we unfortunately have to fallback to just getting the single emission from `onStart`. 😢 

If you think that having the updates is valuable (even if only available on a single additional Android version), then I believe the following would work?:

```kotlin
private val isLocationEnabled: Boolean
    get() = LocationManagerCompat.isLocationEnabled(applicationContext.locationManager)

private val locationEnabledFlow = when {
    SDK_INT > R -> flowOf(true)
    SDK_INT == R -> broadcastReceiverFlow(IntentFilter(PROVIDERS_CHANGED_ACTION))
        .map { intent -> intent.getBooleanExtra(EXTRA_PROVIDER_ENABLED, false) }
        .onStart { emit(isLocationEnabled) }
        .distinctUntilChanged()
    else -> flowOf(isLocationEnabled)
}

private val bluetoothStateFlow =
    broadcastReceiverFlow(IntentFilter(BLUETOOTH_STATE_CHANGED))
        .map { intent -> intent.getIntExtra(EXTRA_STATE, ERROR) }
        .map { state ->
            when (state) {
                STATE_ON -> Available
                STATE_OFF -> Unavailable(reason = Off)
                STATE_TURNING_OFF -> Unavailable(reason = TurningOff)
                STATE_TURNING_ON -> Unavailable(reason = TurningOn)
                else -> error("Unexpected bluetooth state: $state")
            }
        }
        .onStart {
            val isEnabled = when (BluetoothAdapter.getDefaultAdapter()?.isEnabled) {
                true -> Available
                else -> Unavailable(reason = Off)
            }
            emit(isEnabled)
        }

internal actual val bluetoothAvailability: Flow<Bluetooth.Availability> =
    combine(
        locationEnabledFlow,
        bluetoothStateFlow,
    ) { locationEnabled, bluetoothState ->
        if (locationEnabled) bluetoothState else Unavailable(reason = LocationServicesDisabled)
    }
```

core/src/androidMain/kotlin/Bluetooth.kt
```diff
@@ -38,10 +41,21 @@ private val Context.locationManager: LocationManager
     get() = ContextCompat.getSystemService(this, LocationManager::class.java)
         ?: error("LocationManager system service unavailable")
 
-private val isLocationEnabled: Boolean
-    get() = LocationManagerCompat.isLocationEnabled(applicationContext.locationManager)
+private fun isLocationEnabled(): Boolean {
+    return LocationManagerCompat.isLocationEnabled(applicationContext.locationManager)
+}
```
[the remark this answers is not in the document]
mmaleiter: Like I mentioned in the commit comment. It is not the first time I stumble across "property implementation is not reliable, but the function implementation is".  I am currently also coding a KmmPluginChecks implementation, where I tested this on android sdk 9 tablet via log statements. A location enable disabled toggle will not be reported.

I set the version for kable in the test app to 3001 with:

./gradlew             -PRELEASE_SIGNING_ENABLED=false             publishToMavenLocal -Pversion=3001

Please give it a test shot. 

https://github.com/mmaleiter/kmmPluginChecks/tree/test/bluetoothAvailability

core/src/androidMain/kotlin/Bluetooth.kt
mmaleiter: Thank you for the review and especially for catching EXTRA_PROVIDER_ENABLE. I was too keen on code reduction. Also providing the flowof(true), I just did not find the appropriate implementation here, I tried a few thing and new I am still not there.

I think the name locationEnabledFlow is a bit missleading. I see the subject of this flow like -- 
 bluetoothAvailability does not need location to be enabled if sdk int is larger or equal  to android version R. 
I think naming is very important, but do not want to force something. Since I am not very good at naming.

mmaleiter: Maybe isLocationIndependent ?

core/src/androidMain/kotlin/Bluetooth.kt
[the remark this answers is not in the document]
twyatt: > It is not the first time I stumble across "property implementation is not reliable, but the function implementation is".

I have never experienced this.

If you can show that this is a problem here, then we can change it, but I'd like to know that it is a problem before changing it.

core/src/androidMain/kotlin/Bluetooth.kt
mmaleiter: Let's leave the naming like you suggested it. 
Who is supposed to resolve the conversation. For me this one is resolved.

twyatt: > Who is supposed to resolve the conversation.

I'm actually not sure what standard convention is, but in this project I'm fine with either party resolving it. 🙂

core/src/androidMain/kotlin/Bluetooth.kt
```diff
@@ -54,13 +69,17 @@ internal actual val bluetoothAvailability: Flow<Bluetooth.Availability> =
...
+            emit(isEnabled)
         }
+
+internal actual val bluetoothAvailability: Flow<Bluetooth.Availability> =
+    combine(
+        locationEnabledFlow,
+        bluetoothStateFlow,
+    ) { locationEnabled, bluetoothState ->
+        if (locationEnabled) bluetoothState else Unavailable(reason = LocationServicesDisabled)
+    }
```
twyatt, line 85: Missing newline at the end of file.

twyatt (commented):
If you can either confirm that the property accessor for `isLocationEnabled` doesn't work, or revert that change, then add the newline at the end-of-file and this looks ready to me.

core/src/androidMain/kotlin/Bluetooth.kt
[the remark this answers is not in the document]
mmaleiter: So now I think everything did fall into its place. I reverted the change from property to function and also realized that we have to listen to the broadcast to report the bluetoothAvailable status for versions under android R.
I tested it, and will do it again do now again. Pretty sure everything is according to our expectation. I will leave comment to report the manual test result. That you very much for the back and forth, it really help to improve the implementation.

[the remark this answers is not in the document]
mmaleiter: Works like a charm. :+1:

core/src/androidMain/kotlin/Bluetooth.kt
```diff
@@ -41,7 +46,21 @@ private val Context.locationManager: LocationManager
...
-internal actual val bluetoothAvailability: Flow<Bluetooth.Availability> =
+private val locationEnabledFlow = when {
+    SDK_INT > R -> flowOf(true)
+    else -> broadcastReceiverFlow(IntentFilter(PROVIDERS_CHANGED_ACTION))
+        .map { intent ->
+            if (SDK_INT == R) {
+                intent.getBooleanExtra(EXTRA_PROVIDER_ENABLED, false)
+            } else {
+                isLocationEnabled
+            }
```
twyatt, line 57: Oh, very clever approach (to fallback to `isLocationEnabled` for APIs that don't provide `EXTRA_PROVIDER_ENABLED`). 💯 

So, broadcast receiver for `PROVIDERS_CHANGED_ACTION` still triggers below Android R? Nice. 👍

twyatt (approved):
I have not tested this yet, but I like how you brought this all together!

core/src/androidMain/kotlin/Bluetooth.kt
mmaleiter: Thank you. I had that in mind before but just did not do it. :-)

mmaleiter:
Thank you very much, it was really a pleasure. We really yielded a good result.

mmaleiter:
Thanks a lot. :-)

mmaleiter:
Sorry for the git configuration mess (have three different contributors). If you want me to close this PR and register a new one with one commit and one contributor please tell me.

twyatt:
No worries. I'll try to get it merged soon.
Been meaning to test it a little before merge and just haven't found the time. I'll try to make sure and merge it soon though.

core/src/androidMain/kotlin/Bluetooth.kt
```diff
@@ -41,7 +46,21 @@ private val Context.locationManager: LocationManager
...
+        }
+        .onStart { emit(isLocationEnabled) }
```
mmaleiter, line 59: @twyatt I was wondering if 
.onStart { emit(isLocationEnabled) }

should be changed to 

.onStart { emit(isLocationEnabled || SDK_INT > R) }

I think I tested it and found it should not be necessary.

twyatt: It isn't necessary because the `locationEnabledFlow` `when` block will use the `flowOf(true)` when the SDK > `R`.
In other words, the `broadcastReceiverFlow` isn't utilized at all on SDK > `R`.

twyatt:
@mmaleiter it sounds like you've tested this a bit?
If so, let me know, as I'm swamped right now and don't have the time to test it, but I'm happy to merge it if you've verified that it works as expected.

core/src/androidMain/kotlin/Bluetooth.kt
mmaleiter: @twyatt Thank you for the information.

mmaleiter:
@twyatt Yes I tested it thoroughly and found nothing fishy. We do not need to hurry though. I am happy when it is merged but do not want to rush it.

twyatt:
> We do not need to hurry though. I am happy when it is merged but do not want to rush it.

No worries. It will be a little while before we do another Kable release; so we might find some time to perform more testing or fix any bugs if we find any before the next release.

Thanks for the contribution! 💯

PR #475, opened 2023-05-01
Use JDK 17 on CI

twyatt:
Newer versions of AGP require JDK 17:

```
An exception occurred applying plugin request [id: 'com.android.library', artifact: 'com.android.tools.build:gradle:null']
> Failed to apply plugin 'com.android.internal.library'.
   > Android Gradle plugin requires Java 17 to run. You are currently using Java 11.
      Your current JDK is located in /Users/runner/hostedtoolcache/Java_Temurin-Hotspot_jdk/11.0.18-10/x64/Contents/Home
      You can try some of the following options:
       - changing the IDE settings.
       - changing the JAVA_HOME environment variable.
       - changing `org.gradle.java.home` in `gradle.properties`.
```

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -19,10 +19,9 @@ plugins {
  */
 kotlin {
     explicitApi()
+    jvmToolchain(libs.versions.jvm.toolchain.get().toInt())
 
-    android {
-        publishAllLibraryVariants()
-    }
+    android().publishAllLibraryVariants()
     js().browser()
     iosX64()
     macosArm64()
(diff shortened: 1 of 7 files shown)


core/build.gradle.kts
```diff
@@ -131,6 +130,13 @@ kotlin {
 }
 
 android {
+    // Workaround (for `jvmToolchain` not being honored) needed until AGP 8.1.0-alpha09.
+    // https://kotlinlang.org/docs/gradle-configure-project.html#gradle-java-toolchains-support
+    compileOptions {
+        sourceCompatibility = JavaVersion.VERSION_11
+        targetCompatibility = JavaVersion.VERSION_11
+    }
```
twyatt, line 138: Not changed, just moved with code comment updated (to indicate when this configuration directive can be removed).

QuantumRand (approved):

sdonn3 (approved):

PR #488, opened 2023-05-26
Introduce `SharedRepeatableAction` for handling connect

### LINKED ISSUE #339: `connect()` freezes when invoked multiple times concurrently
I am using your library for quite some time now and never had this issue on Android. However, on iOS I rarely get the following error when the connection to my other device fails:

```swift
[DEBUG] ===== CONNECT =====
2022-06-08 16:47:23.481038+0200 Debug[1821:62982] I/Kable: 5DAB21AF-411C-3802-5F8A-ABDA17408F8A Connecting

2022-06-08 16:47:23.702496+0200 Debug[1821:62982] Sentry - debug:: Add breadcrumb: <SentryBreadcrumb: 0x281878c80>
2022-06-08 16:47:23.823228+0200 Debug[1821:62982] D/Kable: 5DAB21AF-411C-3802-5F8A-ABDA17408F8A CentralManagerDelegate state change
  state: DidConnect(identifier=5DAB21AF-411C-3802-5F8A-ABDA17408F8A)
2022-06-08 16:47:23.823619+0200 Debug[1821:62982] V/Kable: 5DAB21AF-411C-3802-5F8A-ABDA17408F8A discoverServices

2022-06-08 16:47:23.834300+0200 Debug[1821:62982] Sentry - debug:: Add breadcrumb: <SentryBreadcrumb: 0x281883380>
2022-06-08 16:47:24.586824+0200 Debug[1821:62982] Sentry - debug:: currentOrientation is unknown.
2022-06-08 16:47:25.159094+0200 Debug[1821:62982] I/Kable: 5DAB21AF-411C-3802-5F8A-ABDA17408F8A Disconnected

2022-06-08 16:47:25.166130+0200 Debug[1821:62982] E/Kable: 5DAB21AF-411C-3802-5F8A-ABDA17408F8A Failed to connect

kotlinx.coroutines.JobCancellationException: LazyDeferredCoroutine was cancelled; job=LazyDeferredCoroutine{Cancelling}@ab111d0
    at 0   shared                              0x10719d4cb        kfun:kotlinx.coroutines.JobCancellationException#<init>(kotlin.String;kotlin.Throwable?;kotlinx.coroutines.Job){} + 179 
    at 1   shared                              0x10714c107        kfun:kotlinx.coroutines.JobSupport#cancel(kotlin.coroutines.cancellation.CancellationException?){} + 203 
    at 2   shared                              0x1071e09a3        kfun:com.juul.kable.ApplePeripheral.onDisconnected#internal + 647 
    at 3   shared                              0x1071ed057        kfun:com.juul.kable.ApplePeripheral.$connectAsync$lambda-6$lambda-3COROUTINE$475.invokeSuspend#internal + 447 
    at 4   shared                              0x1071ed2d3        kfun:com.juul.kable.ApplePeripheral.$connectAsync$lambda-6$lambda-3COROUTINE$475.invoke#interna

```

The call for `peripheral.connect()` looks like this:
```kotlin
try {
    Log.debug { "===== CONNECT =====" }
    newPeripheral.connect()
    Log.debug { "===== DONE =====" }
} catch(ex: Exception) {
    Log.debug { "===== ERROR =====" }
    Log.debug { "Connecting failed!\n" + ex.stackTraceToString() }
    return@coroutineScope true
} finally {
    Log.debug { "===== FINALLY =====" }
}
```

As you can see from the error message above, `===== CONNECT =====` is printed, by neither `===== ERROR =====` nor `===== FINALLY =====` appears, so I am assuming that `connect()` never returns.

On iOS, I call this code from the UI thread. Do you have any idea why this might happen? When the connection can be established, everything works fine afterwards.

twyatt:
Provides a mechanism for allowing an **action** to be performed, and concurrent threads to all watch for the result; while also allowing for the **action** to be cancelled and started over again. _Useful for handling the connect process._

Based on (evolution of) the [`SharedRepeatableTask`](https://github.com/JuulLabs/kable/blob/0.23.0/core/src/jsMain/kotlin/SharedRepeatableTask.kt) class with the primary differentiator: that `SharedRepeatableAction` provides a `CoroutineScope` to the `action` that allows for coroutines to be spawned that **outlive the `action`**.

`SharedRepeatableAction` is available in `common` code but will initially only be used in Apple (and I'll migrate the other platforms over later).

Closes #339.

### DIFF
--- a/core/src/commonMain/kotlin/SharedRepeatableAction.kt
+++ b/core/src/commonMain/kotlin/SharedRepeatableAction.kt
@@ -33,13 +33,14 @@ internal class SharedRepeatableAction<T>(
     private val action: suspend (scope: CoroutineScope) -> T,
 ) {
 
-    private class Jobs<T>(
+    private class State<T>(
         val root: Job,
         val action: Deferred<T>,
-        var isCancelled: Boolean = false,
-    )
+    ) {
+        var isCancelled: Boolean = false
+    }
 
-    private var jobs: Jobs<T>? = null
+    private var state: State<T>? = null
     private val guard = reentrantLock()
 
     suspend fun await() = getOrAsync().await()
@@ -47,7 +48,7 @@ internal class SharedRepeatableAction<T>(
     internal fun getOrAsync(): Deferred<T> = guard.withLock {
         // ktlint-disable indent
         (
-            jobs?.takeUnless { it.isCancelled } ?: run {
+            state?.takeUnless { it.isCancelled } ?: run {
                 val rootJob = Job(coroutineContext.job)
                 val rootScope = CoroutineScope(coroutineContext + rootJob)
                 val actionJob = rootScope.async { action(rootScope) }.apply {
@@ -55,21 +56,21 @@ internal class SharedRepeatableAction<T>(
                         if (cause != null) this@SharedRepeatableAction.reset()
                     }
                 }
-                Jobs(rootJob, actionJob)
-            }.also { jobs = it }
+                State(rootJob, actionJob)
+            }.also { state = it }
         ).action
         // ktlint-enable indent
     }
 
     fun reset() {
         guard.withLock {
-            jobs?.apply { isCancelled = true }
+            state?.apply { isCancelled = true }
         }?.root?.cancel()
     }
 
     suspend fun resetAndJoin() {
         guard.withLock {
-            jobs?.apply { isCancelled = true }
+            state?.apply { isCancelled = true }
         }?.root?.cancelAndJoin()
     }
 }
--- a/core/src/commonMain/kotlin/SharedRepeatableAction.kt
+++ b/core/src/commonMain/kotlin/SharedRepeatableAction.kt
@@ -13,20 +13,20 @@ import kotlin.coroutines.CoroutineContext
 /**
  * A mechanism for launching and awaiting a shared action ([Deferred]) repeatedly.
  *
- * The [action] is started by calling [getOrAsync]. Subsequent calls to [getOrAsync] will return the
- * same (i.e. shared) [action] until cancelled or failure occurs.
+ * The [action] is started by calling [await]. Subsequent calls to [await] will return the same
+ * (i.e. shared) [action] until cancelled or failure occurs.
  *
  * The [action] is passed a [scope][CoroutineScope] that can be used to spawn coroutines that can
- * outlive the [action]. [getOrAsync] will continue to return the same action until either
- * [cancelled][cancel], or a failures occurs in either the [action] or any coroutines spawned from
+ * outlive the [action]. [await] will continue to return the same action until either
+ * [cancelled][reset], or a failures occurs in either the [action] or any coroutines spawned from
  * the [scope][CoroutineScope] provided to [action].
  *
  * An exception thrown from [action] will cancel any coroutines spawned from the
  * [scope][CoroutineScope] that was provided to the [action].
  *
- * Calling [cancel] or [cancelAndJoin] will cancel the [action] and any coroutines created from the
- * [scope][CoroutineScope] provided to the [action]. A subsequent call to [getOrAsync] will then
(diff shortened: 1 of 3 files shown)


core/src/commonMain/kotlin/SharedRepeatableAction.kt
```diff
@@ -0,0 +1,79 @@
...
+import kotlinx.coroutines.async
+import kotlinx.coroutines.cancelAndJoin
+import kotlinx.coroutines.job
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * A mechanism for launching and awaiting a shared action ([Deferred]) repeatedly.
+ *
+ * The [action] is started by calling [await]. Subsequent calls to [await] will return the same
+ * (i.e. shared) [action] until cancelled or failure occurs.
```
cedrickcooke, line 17: Just checking that I'm reading this correctly: if the action completes successfully, subsequent calls to `await` will return the cached result.

core/src/commonMain/kotlin/SharedRepeatableAction.kt
```diff
@@ -0,0 +1,79 @@
...
+    private val action: suspend (scope: CoroutineScope) -> T,
+) {
+
+    private class Jobs<T>(
+        val root: Job,
+        val action: Deferred<T>,
+        var isCancelled: Boolean = false,
+    )
+
+    private var jobs: Jobs<T>? = null
```
cedrickcooke, line 42: Nit: `Jobs` feels kinda collection-y, while the structure of this class is fixed at exactly 2 jobs (and even then, it's 1 parent and 1 child). `State` might be less confusing, even though it's a little generic.

core/src/commonMain/kotlin/SharedRepeatableAction.kt
```diff
@@ -0,0 +1,79 @@
...
+ */
+internal class SharedRepeatableAction<T>(
+    private val coroutineContext: CoroutineContext,
+    private val action: suspend (scope: CoroutineScope) -> T,
+) {
+
+    private class Jobs<T>(
+        val root: Job,
+        val action: Deferred<T>,
+        var isCancelled: Boolean = false,
```
cedrickcooke, line 39: Nit: Probably shouldn't be in the constructor.

cedrickcooke (approved):

core/src/commonMain/kotlin/SharedRepeatableAction.kt
```diff
@@ -0,0 +1,79 @@
...
+ * the [scope][CoroutineScope] provided to [action].
+ *
+ * An exception thrown from [action] will cancel any coroutines spawned from the
+ * [scope][CoroutineScope] that was provided to the [action].
+ *
+ * Calling [reset] or [resetAndJoin] will cancel the [action] and any coroutines created from the
+ * [scope][CoroutineScope] provided to the [action]. A subsequent call to [await] will then start
+ * the [action] again.
+ */
+internal class SharedRepeatableAction<T>(
```
sdonn3, line 31: Love it

sdonn3 (approved):
Seems useful + the tests inspire confidence

core/src/commonMain/kotlin/SharedRepeatableAction.kt
twyatt: Correct. That is intentional. This is primarily for connection handling.
If connect `action` completed successfully, we'll still return the same `Deferred` to await on. Since it will already be done, the `await` will return immediately.
Since coroutines may have been launched from the `action`, those could still be running, if they fail at any point, then it resets the "repeatable action".

core/src/commonMain/kotlin/SharedRepeatableAction.kt
twyatt: 🤦 ya, that was silly. Will change. 👍

davidtaylor-juul (approved):

PR #490, opened 2023-05-31
Replace `Semaphore` with `Mutex` in `Connection`

twyatt:
I believe `Semaphore` was previously used because the `Mutex` could deadlock under the old memory model.


Phoenix7351 (approved):
Looks good, but just wanted to point out that it looks like Kable doesn't have any branch merge protections to prevent PRs being merged without approvals?

twyatt:
> Looks good, but just wanted to point out that it looks like Kable doesn't have any branch merge protections to prevent PRs being merged without approvals?

It does have branch protection for merges to `main`. This PR is being merged into another branch.

cedrickcooke (approved):

davidtaylor-juul (approved):

PR #491, opened 2023-05-31
Simplify internal lifecycle of `Peripheral`

twyatt:
The internals of `Peripheral` now lean more on structured concurrency (disposal of objects uses coroutine lifecycle mechanisms and parent-child relationships).

### DIFF
--- a/core/src/appleMain/kotlin/Connection.kt
+++ b/core/src/appleMain/kotlin/Connection.kt
@@ -7,19 +7,22 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.async
-import kotlinx.coroutines.cancel
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
-import kotlin.coroutines.CoroutineContext
 
 internal class Connection(
-    parentCoroutineContext: CoroutineContext,
+    scope: CoroutineScope,
     val delegate: PeripheralDelegate,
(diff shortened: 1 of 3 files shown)


core/src/appleMain/kotlin/Connection.kt
```diff
@@ -59,9 +62,4 @@ internal class Connection(
         if (error != null) throw IOException(error.description, cause = null)
         response as T
     }
-
-    fun close() {
```
QuantumRand, line 63: Is the idea now that instead of closing a Connection, you'll cancel its scope?

twyatt: From a consumer standpoint, nothing will change. `Peripheral.disconnect` will continue to disconnect. Ultimately, cancelling a `Peripheral`'s parent scope will also cancel (force dispose) the peripheral.

This is more of an internal book-keeping change. We no longer have to explicitly cancel the `Connection`'s `scope` because we now make its parent the scope of the `SharedRepeatableAction`: https://github.com/JuulLabs/kable/blob/e30220928290d52031ebcb3a17cf5e8cb449bd13/core/src/appleMain/kotlin/Peripheral.kt#L191 _(which will be cleaned up automatically on disconnect)_

`ShareRepeatableAction` compartmentalizes the connection lifecycle nicely: https://github.com/JuulLabs/kable/blob/e30220928290d52031ebcb3a17cf5e8cb449bd13/core/src/commonMain/kotlin/SharedRepeatableAction.kt

QuantumRand (approved):

davidtaylor-juul (approved):

Issue #500, opened 2023-06-12
Unexpected "Suppressed failure: com.juul.kable.NotReadyException" on `observe`


twyatt:
Reported via `#juul-libraries` that `Suppressed failure: com.juul.kable.NotReadyException` failure is being hit when attempting to `observe` despite being connected:

```
2023-06-12 10:54:48.659554+0200 Bodytone[15381:1610952] W/Kable/Scanner: According to Core Bluetooth documentation: "The recommended practice is to populate the serviceUUIDs parameter rather than leaving it nil." This means providing only (and at least 1) filter(s) of type `Filter.Service` to Scanner. See https://developer.apple.com/documentation/corebluetooth/cbcentralmanager/1518986-scanforperipheralswithservices#discussion for more details.

2023-06-12 10:54:48.679700+0200 Bodytone[15381:1610952] I/Kable/Scanner: Starting scan with non-native filtering

2023-06-12 10:54:48.747381+0200 Bodytone[15381:1610952] I/Kable/Scanner: Stopping scan

2023-06-12 10:54:48.749047+0200 Bodytone[15381:1610952] I/Kable: 39D79EE2-1EF7-0AC0-B8BE-4C500C92CBF9 Connecting

2023-06-12 10:54:49.250688+0200 Bodytone[15381:1610952] D/Kable: 39D79EE2-1EF7-0AC0-B8BE-4C500C92CBF9 CentralManagerDelegate state change
  state: DidConnect(identifier=39D79EE2-1EF7-0AC0-B8BE-4C500C92CBF9)
2023-06-12 10:54:49.251319+0200 Bodytone[15381:1610952] V/Kable: 39D79EE2-1EF7-0AC0-B8BE-4C500C92CBF9 discoverServices

2023-06-12 10:54:49.366266+0200 Bodytone[15381:1610958] D/Kable/Delegate: 39D79EE2-1EF7-0AC0-B8BE-4C500C92CBF9 39D79EE2-1EF7-0AC0-B8BE-4C500C92CBF9 didDiscoverServices

2023-06-12 10:54:49.369598+0200 Bodytone[15381:1610958] D/Kable/Delegate: 39D79EE2-1EF7-0AC0-B8BE-4C500C92CBF9 39D79EE2-1EF7-0AC0-B8BE-4C500C92CBF9 didDiscoverCharacteristicsForService
  service: 180A
[...]

PR #501, opened 2023-06-13
Add a unit test that validates that `SharedRepeatableAction` remains active

twyatt:
Created to test #500.
It was later determined that #500 was invalid, but the test will help prevent potential regressions.


core/src/commonTest/kotlin/SharedRepeatableActionTests.kt
```diff
@@ -127,4 +129,21 @@ class SharedRepeatableActionTests {
 
         coroutineContext.cancelChildren()
     }
+
+    @Test
+    fun nothingLaunchedFromScope_remainsActive() = runTest {
```
Phoenix7351, line 134: Total nit, but it'd be nice to have `sharedRepeatableAction` specified in the signature as the thing being tested. Understandably, we get this information through the class name, so I can understand why it's left out. But being verbose is free. ^_^

PR #518, opened 2023-07-04
Add identifier property to Advertisement object

### LINKED ISSUE #517: Expose identifier in Advertisement just like is exposed in Peripheral object
Hello everyone.

What do you think to expose the identifier (macadress, unique identifier) in Advertisement object? The address already is available in AndroidAdvertisement but it is good to have the identifier available in common code as well.

Use case: mount a list with advertisement founded directly from common code.

Thank you.

francismariano:
Closes #517

### DIFF
--- a/core/src/appleMain/kotlin/CBPeripheralCoreBluetoothAdvertisement.kt
+++ b/core/src/appleMain/kotlin/CBPeripheralCoreBluetoothAdvertisement.kt
@@ -11,7 +11,6 @@ import platform.CoreBluetooth.CBPeripheral
 import platform.CoreBluetooth.CBUUID
 import platform.Foundation.NSData
 import platform.Foundation.NSNumber
-import platform.Foundation.NSUUID
 
 internal class CBPeripheralCoreBluetoothAdvertisement(
     override val rssi: Int,
@@ -19,7 +18,7 @@ internal class CBPeripheralCoreBluetoothAdvertisement(
     internal val cbPeripheral: CBPeripheral,
 ) : CoreBluetoothAdvertisement {
 
-    override val identifier: Uuid
+    override val identifier: Identifier
         get() = cbPeripheral.identifier.toUuid()
 
     override val name: String?
@@ -70,14 +69,14 @@ internal class CBPeripheralCoreBluetoothAdvertisement(
     override fun equals(other: Any?): Boolean {
         if (this === other) return true
         if (other !is CoreBluetoothAdvertisement) return false
-        if (identifier.toString() != other.identifier.toString()) return false
+        if (identifier != other.identifier) return false
         return true
     }
 
     override fun hashCode(): Int = identifier.hashCode()
 
     override fun toString(): String =
-        "Advertisement(identifier=${identifier}, name=$name, rssi=$rssi, txPower=$txPower)"
+        "Advertisement(identifier=$identifier, name=$name, rssi=$rssi, txPower=$txPower)"
 }
 
 internal fun NSData.toManufacturerData(): ManufacturerData? = toByteArray().toManufacturerData()
--- a/core/src/appleMain/kotlin/CBPeripheralCoreBluetoothAdvertisement.kt
+++ b/core/src/appleMain/kotlin/CBPeripheralCoreBluetoothAdvertisement.kt
@@ -19,8 +19,8 @@ internal class CBPeripheralCoreBluetoothAdvertisement(
     internal val cbPeripheral: CBPeripheral,
 ) : CoreBluetoothAdvertisement {
 
-    override val identifier: NSUUID
-        get() = cbPeripheral.identifier
+    override val identifier: Uuid
+        get() = cbPeripheral.identifier.toUuid()
 
     override val name: String?
         get() = data[CBAdvertisementDataLocalNameKey] as? String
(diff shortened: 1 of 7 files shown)


core/src/appleMain/kotlin/CBPeripheralCoreBluetoothAdvertisement.kt
```diff
@@ -19,8 +19,8 @@ internal class CBPeripheralCoreBluetoothAdvertisement(
     internal val cbPeripheral: CBPeripheral,
 ) : CoreBluetoothAdvertisement {
 
-    override val identifier: NSUUID
-        get() = cbPeripheral.identifier
+    override val identifier: Uuid
```
twyatt, line 22: To be consistent with other platform overrides, let's use `Identifier`.

```suggestion
    override val identifier: Identifier
```

core/src/appleMain/kotlin/CBPeripheralCoreBluetoothAdvertisement.kt
```diff
@@ -70,14 +70,14 @@ internal class CBPeripheralCoreBluetoothAdvertisement(
     override fun equals(other: Any?): Boolean {
         if (this === other) return true
         if (other !is CoreBluetoothAdvertisement) return false
-        if (identifier.UUIDString != other.identifier.UUIDString) return false
+        if (identifier.toString() != other.identifier.toString()) return false
```
twyatt, line 73: `Uuid` has `equals`, so no need to convert to `String`.

```suggestion
        if (identifier != other.identifier) return false
```

core/src/appleMain/kotlin/CBPeripheralCoreBluetoothAdvertisement.kt
```diff
@@ -70,14 +70,14 @@ internal class CBPeripheralCoreBluetoothAdvertisement(
...
         return true
     }
 
     override fun hashCode(): Int = identifier.hashCode()
 
     override fun toString(): String =
-        "Advertisement(identifier=${identifier.UUIDString}, name=$name, rssi=$rssi, txPower=$txPower)"
+        "Advertisement(identifier=${identifier}, name=$name, rssi=$rssi, txPower=$txPower)"
```
twyatt, line 80:
```suggestion
        "Advertisement(identifier=$identifier, name=$name, rssi=$rssi, txPower=$txPower)"
```

core/src/appleMain/kotlin/CoreBluetoothAdvertisement.kt
```diff
@@ -5,7 +5,7 @@ import platform.Foundation.NSData
 import platform.Foundation.NSUUID
 
 public interface CoreBluetoothAdvertisement : Advertisement {
-    public val identifier: NSUUID
+    public override val identifier: Identifier
```
twyatt, line 8: Since `CoreBluetoothAdvertisement`'s `identifier' now has same type as `Advertisement`, we can remove the `override` entirely (as it will be inherited).

```suggestion
```

core/src/jsMain/kotlin/WebBluetoothAdvertisement.kt
```diff
@@ -4,7 +4,7 @@ import com.benasher44.uuid.Uuid
 import org.khronos.webgl.DataView
 
 public interface WebBluetoothAdvertisement : Advertisement {
-    public val identifier: String
+    public override val identifier: String
```
twyatt, line 7:
```suggestion
```

twyatt (commented):
Looks great, just a few minor changes.

Also, can you type on `BluetoothAdvertisingEventWebBluetoothAdvertisement`'s `identifier` to be `Identifier`?

core/src/appleMain/kotlin/CBPeripheralCoreBluetoothAdvertisement.kt
francismariano: I used Uuid because the Peripheral is in that way.

https://github.com/JuulLabs/kable/blob/07e7e96a615359b04e7817b6d8a549426d862c03/core/src/appleMain/kotlin/Peripheral.kt#L125

So, I think interesting change the peripheral as well.

twyatt: Ya, not having the `Peripheral` consistent was an oversight. I'll update the `Peripheral` to be more consistent in a separate PR.

core/src/appleMain/kotlin/CBPeripheralCoreBluetoothAdvertisement.kt
francismariano: Do you have some idea??

`Operator '!=' cannot be applied to 'Identifier /* = Uuid */' and 'Identifier'`

twyatt: Bizarre, I'm not seeing that error:

Screenshot 2023-07-05 at 10 39 19 AM

Maybe we need to re-Gradle sync the project? 🤷 

Can you try making the change and just seeing if it compiles anyways (even with the error in the IDE)?

francismariano: The error is just in IDE. What android version are you using?

twyatt: Screenshot 2023-07-05 at 12 34 31 PM

twyatt (approved):
Great job on this!

twyatt (changes requested):
Oh, sorry, one more thing actually: https://github.com/JuulLabs/kable/pull/518#pullrequestreview-1513235842

> Also, can you change type on `BluetoothAdvertisingEventWebBluetoothAdvertisement`'s `identifier` to be `Identifier`?

twyatt (approved):

twyatt:
@francismariano you'll want to run `./gradlew apiDump` and commit the changes.

francismariano:
Sorry for so many commits ;)

twyatt:
> Sorry for so many commits ;)

No worries at all. `git` doesn't care. 😄

PR #527, opened 2023-07-08
Update README.md with explicit note about peripheral scope

rocketraman:


README.md
```diff
@@ -108,6 +108,9 @@ connection handling and I/O operations.
 val peripheral = scope.peripheral(advertisement)
 ```
 
+Note that if the scope is closed, the peripheral is automatically disconnected. Ensure that the lifetime of the used
+scope matches the required lifetime of the peripheral connection.
```
twyatt, line 112: I appreciate documentation being updated, but is this not covered in the **Structured Concurrency** section (around line 443) below?

If you feel that it isn't discoverable enough being so low in the documentation, then it does make sense to make mention of this here as well — maybe just mention that more details can be found in the **Structured Concurrency** section?

rocketraman: @twyatt I've updated this to include a reference to the structured concurrency section. I do believe this is important enough to be explicitly called out here.

twyatt (approved):

davidtaylor-juul (approved):

PR #538, opened 2023-07-26
Provide additional details to data processor

### LINKED ISSUE #532: `Logging.DataProcessor` should be provided with details about the data
It would be useful to know where the data came from, or is going. In other words, the `process` function should be provided with details of:
- if data is part of read or write
- if data associated with either:
  - characteristic
  - descriptor

twyatt:
`Logging.DataProcessor` now has 4 additional parameters:

- `operation: Operation?` (either: `Write`, `Read` or `Change`)
- `service: Uuid?`
- `characteristic: Uuid?`
- `descriptor: Uuid?`

These appear after the existing `data: ByteArray` parameter:

```kotlin
DataProcessor { data, operation, service, characteristic, descriptor ->
  // todo: process data
}
```

This allows library consumers to change their data processing behavior based on the characteristic or descriptor that the `data` was associated with.

For library consumers that don't need the new parameters, they'll need to update their custom data processors as follows:

```kotlin
DataProcessor { data, _, _, _, _ ->
  // todo: process data
}
```

Closes #532

---

Unit tests were added because the logging message builder had to go through an overhaul to add this support, so the unit tests provided some confidence that the output remained as expected.


core/src/commonMain/kotlin/logs/Logging.kt
```diff
@@ -41,7 +43,12 @@ public class Logging {
     }
 
     public fun interface DataProcessor {
-        public fun process(data: ByteArray): String
+        public fun process(
+            data: ByteArray,
+            serviceUuid: Uuid?,
+            characteristicUuid: Uuid?,
+            descriptorUuid: Uuid?,
```
cedrickcooke, line 50: If this is likely to get more fields added in the future, would be nice to wrap these up in a `class` of some sort. Would still be breaking now, but would make it non-breaking to add fields in the future.

twyatt: Admittedly I'm taking a "move fast and break things" mindset with this, partly because it feels warranted while pre-1.x — but mostly because I want to ditch the data processor entirely in favor of #539. 🤷

twyatt: Added `@ObsoleteKableApi` to help warn consumers that the API will likely experience churn.

cedrickcooke (approved):

davidtaylor-juul (approved):

sdonn3 (approved):

davidtaylor-juul (approved):

Issue #308, opened 2022-04-08
Kable Hangs IOS app when device disconnects multiple times


burnhamd:
Occasionally while using bluetooth background mode I see that when I return to the foreground, if an event occurred in the background that the app is hung.  Looking at the trace, I see that it is almost always hung up in the following code:
dispatch_sync(dispatchQueue) {
            block.run()
        }

I can recreate this issue by disconnecting a device multiple times in the background and seeing that it is called from the cancelPeripheralConnection() in connectAsync()

burnhamd:
Looking more into this, it is not limited to background.
What I have found is that it fails if there is some action taking place like write, or discovery of services.  

It seems that this is being executed on the ApplePeripheral.connection context.  But the actual write gets executed on the central manager dispatcher context.  The problem seems to be when the connection has been closed but the central manager function is running.  Then it just hangs.  Haven't found a fix.  Im not familiar enough with this code to be able to correct it at this time.

twyatt:
Thanks for investigating! Your detailed explanation of the cause is very helpful. When I have some time, I'll explore possible solutions. 👍

PR #548, opened 2023-08-10
Swap dispatch_sync to dispatch_async in NativeDarwin QueueDispatcher

Phoenix7351:
Resolves https://github.com/JuulLabs/kable/issues/308

`dispatch` being implemented with a synchronous call frequently caused deadlocks--which would lead to hangs or crashes for the NativeDarwin consumer. Changing to an asynchronous addition to the queue fixes this.

### DIFF
--- a/core/src/appleMain/kotlin/QueueDispatcher.kt
+++ b/core/src/appleMain/kotlin/QueueDispatcher.kt
@@ -2,9 +2,9 @@ package com.juul.kable
 
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Runnable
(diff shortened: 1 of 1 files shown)


twyatt:
It looks like Ktor samples also use `async`. 👍 

https://github.com/ktorio/ktor-samples/blob/b89817099f3d58c898fdee162992ec4d9685b253/client-mpp/shared/src/iosMain/kotlin/io/ktor/samples/mpp/client/ApplicationDispatcher.kt#L9-L17

QuantumRand (approved):

twyatt (approved):

PR #551, opened 2023-08-18
CA-2351: Enable iOS Backgrounding

Phoenix7351:
This PR allows Apple consumers to take advantage of being woken up in the background to receive BLE events (i.e. device connect/disconnect). 

Also fixes a crash that would occur when the consumer used solicitedServiceUuids--which is needed by the OS to wake up the app.

### DIFF
--- a/core/src/appleMain/kotlin/CentralManager.kt
+++ b/core/src/appleMain/kotlin/CentralManager.kt
@@ -12,9 +12,11 @@ import platform.CoreBluetooth.CBDescriptor
 import platform.CoreBluetooth.CBPeripheral
 import platform.CoreBluetooth.CBService
 import platform.CoreBluetooth.CBUUID
+import platform.CoreBluetooth.CBCentralManagerOptionRestoreIdentifierKey
 import platform.Foundation.NSData
 
 private const val DISPATCH_QUEUE_LABEL = "central"
+private const val CBCENTRALMANAGER_RESTORATION_ID = "kable-central-manager" // add a unique id to it
 
 public class CentralManager internal constructor() {
 
@@ -24,7 +26,8 @@ public class CentralManager internal constructor() {
 
     private val dispatcher = QueueDispatcher(DISPATCH_QUEUE_LABEL)
     internal val delegate = CentralManagerDelegate()
-    private val cbCentralManager = CBCentralManager(delegate, dispatcher.dispatchQueue)
+    private val cbOptions = mutableMapOf<Any?, Any>(CBCentralManagerOptionRestoreIdentifierKey to CBCENTRALMANAGER_RESTORATION_ID).toMap()
+    private val cbCentralManager = CBCentralManager(delegate, dispatcher.dispatchQueue, cbOptions)
 
     internal suspend fun scanForPeripheralsWithServices(
         services: List<Uuid>?,
--- a/core/src/appleMain/kotlin/CentralManager.kt
+++ b/core/src/appleMain/kotlin/CentralManager.kt
@@ -6,13 +6,13 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.withContext
 import platform.CoreBluetooth.CBCentralManager
+import platform.CoreBluetooth.CBCentralManagerOptionRestoreIdentifierKey
 import platform.CoreBluetooth.CBCharacteristic
 import platform.CoreBluetooth.CBCharacteristicWriteType
 import platform.CoreBluetooth.CBDescriptor
 import platform.CoreBluetooth.CBPeripheral
 import platform.CoreBluetooth.CBService
 import platform.CoreBluetooth.CBUUID
-import platform.CoreBluetooth.CBCentralManagerOptionRestoreIdentifierKey
 import platform.Foundation.NSData
 import platform.Foundation.NSUUID
 import platform.Foundation.NSUserDefaults
--- a/core/src/appleMain/kotlin/CentralManager.kt
+++ b/core/src/appleMain/kotlin/CentralManager.kt
@@ -14,9 +14,12 @@ import platform.CoreBluetooth.CBService
 import platform.CoreBluetooth.CBUUID
 import platform.CoreBluetooth.CBCentralManagerOptionRestoreIdentifierKey
 import platform.Foundation.NSData
+import platform.Foundation.NSUUID
+import platform.Foundation.NSUserDefaults
 
 private const val DISPATCH_QUEUE_LABEL = "central"
-private const val CBCENTRALMANAGER_RESTORATION_ID = "kable-central-manager" // add a unique id to it
+private const val CBCENTRALMANAGER_RESTORATION_ID = "kable-central-manager"
+private const val CBCENTRALMANAGER_CONSUMER_ID_KEY = "kable-central-manager-consumer-id"
 
 public class CentralManager internal constructor() {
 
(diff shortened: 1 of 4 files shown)


core/src/appleMain/kotlin/CentralManager.kt
```diff
@@ -6,25 +6,48 @@ import kotlinx.coroutines.CoroutineScope
...
+    private val consumerId: String
+        get() {
+            val id = userDefaults.stringForKey(CBCENTRALMANAGER_CONSUMER_ID_KEY)
+            return if (id != null) {
+                id
+            } else {
+                val id = NSUUID().UUIDString()
+                userDefaults.setObject(id, CBCENTRALMANAGER_CONSUMER_ID_KEY)
+                id
+            }
```
cedrickcooke, line 44: Nit: intent is a little clearer with mutation

```suggestion
            var id = userDefaults.stringForKey(CBCENTRALMANAGER_CONSUMER_ID_KEY)
            if (id == null) {
                id = NSUUID().UUIDString()
                userDefaults.setObject(id, CBCENTRALMANAGER_CONSUMER_ID_KEY)
            }
            return id
```

cedrickcooke (approved):

core/src/appleMain/kotlin/CentralManager.kt
davidtaylor-juul: Also nit: even clearer in point-free style IMHO:
```suggestion
            get() = userDefaults.stringForKey(CBCENTRALMANAGER_CONSUMER_ID_KEY)
                ?: NSUUID().UUIDString().also {
                    userDefaults.setObject(it, CBCENTRALMANAGER_CONSUMER_ID_KEY)
                }
```

core/src/appleMain/kotlin/ScannerBuilder.kt
```diff
@@ -34,8 +35,8 @@ public actual class ScannerBuilder {
         allowDuplicateKeys?.also {
             options[CBCentralManagerScanOptionAllowDuplicatesKey] = it
         }
-        solicitedServiceUuids?.also {
-            options[CBCentralManagerScanOptionSolicitedServiceUUIDsKey] = it.toTypedArray()
+        solicitedServiceUuids?.also { uuids ->
+            options[CBCentralManagerScanOptionSolicitedServiceUUIDsKey] = uuids.map { it.toCBUUID() } as NSArray
```
davidtaylor-juul, line 39: `map` on `List<T>` will return another `List` which would probably (should?) fail to typecast here. Probably it should be:
```kotlin
    uuids.map { it.toCBUUID() }.toTypedArray()
```

My understanding is the `toTypedArray` will convert `List` to the native array type, certainly that is how it works when bridging arrays on Js for example, and one assumes it is the same for Obj-C.

Phoenix7351: This causes a crash:

```
*** Terminating app due to uncaught exception 'Invalid Object', reason: 'Object kotlin.Array@94b0a00 is invalid'
```

davidtaylor-juul: Wild.

davidtaylor-juul (approved):

core/src/appleMain/kotlin/ScannerBuilder.kt
Phoenix7351: 🤷‍♂️

core/src/appleMain/kotlin/CentralManager.kt
```diff
@@ -6,25 +6,42 @@ import kotlinx.coroutines.CoroutineScope
...
+    private val consumerId: String
+        get() = userDefaults.stringForKey(CBCENTRALMANAGER_CONSUMER_ID_KEY)
+            ?: NSUUID().UUIDString().also {
+                userDefaults.setObject(it, CBCENTRALMANAGER_CONSUMER_ID_KEY)
+            }
+
     private val dispatcher = QueueDispatcher(DISPATCH_QUEUE_LABEL)
     internal val delegate = CentralManagerDelegate()
-    private val cbCentralManager = CBCentralManager(delegate, dispatcher.dispatchQueue)
+    private val cbOptions = mutableMapOf<Any?, Any>(CBCentralManagerOptionRestoreIdentifierKey to CBCENTRALMANAGER_RESTORATION_ID + consumerId).toMap()
```
twyatt, line 43: The way this is concatenated it will make for some ugly values, e.g.: `kable-central-managerd97f4db7-2bbf-424b-99f2-a0dc224d086a`

The concatenation should probably read: `"$CBCENTRALMANAGER_RESTORATION_ID-$consumerId"`

core/src/appleMain/kotlin/CentralManager.kt
```diff
@@ -6,25 +6,42 @@ import kotlinx.coroutines.CoroutineScope
...
+    private val cbCentralManager = CBCentralManager(delegate, dispatcher.dispatchQueue, cbOptions)
```
twyatt, line 44: It doesn't look necessary to have the mutable map, nor the typing defined? This could be simplified as:

```kotlin
    private val cbCentralManager = CBCentralManager(
        delegate,
        dispatcher.dispatchQueue,
        options = mapOf(CBCentralManagerOptionRestoreIdentifierKey to "$CBCENTRALMANAGER_RESTORATION_ID-$consumerId"),
    )
```

core/src/appleMain/kotlin/CentralManagerDelegate.kt
```diff
@@ -120,7 +120,9 @@ internal class CentralManagerDelegate : NSObject(), CBCentralManagerDelegateProt
         _state.value = central.state
     }
 
-    // todo: func centralManager(CBCentralManager, willRestoreState: [String : Any])
+    override fun centralManager(central: CBCentralManager, willRestoreState: Map<Any?, *>) {
+        // NO OP
```
twyatt, line 124: It isn't clear why this can be a "no-op", it should probably explain it a little bit?

```suggestion
        // No-op: Unnecessary because...
```

core/src/appleMain/kotlin/ScannerBuilder.kt
twyatt, line 39: I don't think this cast is necessary and can be simplified to:

```suggestion
            options[CBCentralManagerScanOptionSolicitedServiceUUIDsKey] = uuids.map(Uuid::toCBUUID)
```

twyatt:
I don't see any changes on the public API, it appears this should've been marked as `patch`? Or am I missing something?

core/src/appleMain/kotlin/CentralManager.kt
Phoenix7351: I can change it if you really want, but this value is not meant to be seen by any humans. It shouldn't appear in any logging or queue names; it's strictly just for the OS to be able to tell two instances apart in case Kable happens to be running in another iOS app.

Phoenix7351:
> I don't see any changes on the public API, it appears this should've been marked as `patch`? Or did I miss something?

Wasn't aware that API changes were necessary for something to be marked as `minor`. It seemed to me that background support wasn't initially designed with Kable, so this was a new (but small) feature. That's why I went `minor`. 🤷‍♂️

core/src/appleMain/kotlin/ScannerBuilder.kt
Phoenix7351: I thought this would crash, but just ran with it and it looks good!

twyatt:
> > I don't see any changes on the public API, it appears this should've been marked as `patch`? Or did I miss something?
> 
> Wasn't aware that API changes were necessary for something to be marked as `minor`. It seemed to me that background support wasn't initially designed with Kable, so this was a new (but small) feature. That's why I went `minor`. 🤷‍♂️

Re-reading [semver documentation](https://semver.org/):

> MINOR version when you add functionality in a backward compatible manner

I think you're right, makes sense to mark this as `minor`. 👍

PR #558, opened 2023-08-29
Fix Android's scanner behavior when bluetooth is off

cedrickcooke:
Pre this fix, `startScan` would fail internally on high/recent OS versions and the scanner callbacks would never be called.

### DIFF
--- a/core/src/androidMain/kotlin/BluetoothLeScannerAndroidScanner.kt
+++ b/core/src/androidMain/kotlin/BluetoothLeScannerAndroidScanner.kt
@@ -75,6 +75,8 @@ internal class BluetoothLeScannerAndroidScanner(
             }.build()
         }
 
+        checkBluetoothAdapterState(BluetoothAdapter.STATE_ON)
+
         logger.info {
             message = if (scanFilters.isEmpty()) {
                 "Starting scan without filters"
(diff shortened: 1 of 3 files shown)


twyatt (approved):

core/src/androidMain/kotlin/BluetoothLeScannerAndroidScanner.kt
```diff
@@ -75,6 +76,8 @@ internal class BluetoothLeScannerAndroidScanner(
             }.build()
         }
 
+        checkBluetoothAdapterState(STATE_ON)
```
twyatt, line 79: Move right before the `startScan` call? i.e. just above line 88?

cedrickcooke: Can do, wasn't sure which side of the log this should live on

twyatt: Ya, I often have trouble deciding too.
I think in this case though, it would make sense to have a log of what we're about to do right before a crash — I think it is helpful information in this case.

QuantumRand (approved):

PR #561, opened 2023-09-06
Allow `CentralManager` to be configured

### LINKED ISSUE #554: NSInternalInconsistencyException on foreground scan on iOS
After updating to v0.26.0, an `NSInternalInconsistencyException` seems to be thrown at `CentralManager.kt:43:36` with the following message:

>State restoration of CBCentralManager is only allowed for applications that have specified the \"bluetooth-central\" background mode"

We don't scan in the background, so we don't specify a background mode. I'm guessing this is related to #552? For now, we'll just hold off on the update as we don't need it.

twyatt:
Apple consumers can now _optionally_ configure the internally used global `CentralManager`:

```kotlin
CentralManager.configure {
    stateRestoration = true // `false` by default (when not specified).
}
```

This makes "background support" (state restoration) **opt-in** — in other words: not calling `configure` will not enable state restoration.

_As stated in the [updated documentation](https://github.com/JuulLabs/kable/tree/twyatt/central-manager-initialize#background-support), `configure` must be called before using most of Kable's functionality._

Closes #554


twyatt:
Note-to-self: Update documentation re: API change.

Phoenix7351:
Looks good. Gonna test it real quick.

Phoenix7351 (approved):
Works! 🚢

PR #563, opened 2023-09-07
Add support for `autoConnect`

### LINKED ISSUE #410: Support `autoConnect` on Android
As was mentioned in https://github.com/JuulLabs/kable/issues/365#issuecomment-1243022280, for some users having `autoConnect` hardcoded as `false` may cause connection troubles (specifically, GATT 133 errors).

Kable should provide an option to `connect` with `autoConnect` as `true`.

twyatt:
[Martijn van Welie](https://medium.com/@martijn.van.welie) wrote a great article, [Making Android BLE work — part 2](https://medium.com/@martijn.van.welie/making-android-ble-work-part-2-47a3cdaade07), which includes a very informative section on [`autoConnect = true`](https://medium.com/@martijn.van.welie/making-android-ble-work-part-2-47a3cdaade07#7fcf).

~~Making `autoConnect = true` the default, as it is generally easier to manage and will match how Kable behaves on Apple platforms.~~

**UPDATE:** Default now matches the previous behavior, library consumers will need to opt-in to auto-connect.

Library consumers will be able to opt-in to the new behavior via:

```kotlin
val peripheral = scope.peripheral(..) {
    autoConnectIf { true }
}
peripheral.connect()
```

Auto-connect is a lambda predicate to allow consumers to dynamically decide if a connection should be auto-connect. This is useful in situations where initial connection is desired to be `autoConnect = false` (for faster connection) but having subsequent connections be `autoConnect = true` (for battery conservation).

Closes #410

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,7 @@ On Android targets, additional configuration options are available (all configur
 
 ```kotlin
 val peripheral = scope.peripheral(advertisement) {
+    autoConnectIf { true }
     onServicesDiscovered {
         requestMtu(...)
     }
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -287,11 +287,13 @@ public final class com/juul/kable/Peripheral$DefaultImpls {
 }
 
 public final class com/juul/kable/PeripheralBuilder {
+	public final fun getAutoConnect ()Z
 	public final fun getPhy ()Lcom/juul/kable/Phy;
 	public final fun getTransport ()Lcom/juul/kable/Transport;
 	public final fun logging (Lkotlin/jvm/functions/Function1;)V
 	public final fun observationExceptionHandler (Lkotlin/jvm/functions/Function3;)V
 	public final fun onServicesDiscovered (Lkotlin/jvm/functions/Function2;)V
+	public final fun setAutoConnect (Z)V
 	public final fun setPhy (Lcom/juul/kable/Phy;)V
 	public final fun setTransport (Lcom/juul/kable/Transport;)V
 }
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -287,13 +287,12 @@ public final class com/juul/kable/Peripheral$DefaultImpls {
 }
 
 public final class com/juul/kable/PeripheralBuilder {
-	public final fun getAutoConnect ()Z
+	public final fun autoConnectIf (Lkotlin/jvm/functions/Function0;)V
 	public final fun getPhy ()Lcom/juul/kable/Phy;
 	public final fun getTransport ()Lcom/juul/kable/Transport;
 	public final fun logging (Lkotlin/jvm/functions/Function1;)V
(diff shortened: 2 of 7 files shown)


twyatt:
Note to self: need to integrate reflection based workaround for `autoConnect` sometimes not being honors on Android API < 23.

davidtaylor-juul (approved):
Sheesh, got a bit messy there dealing with the API but LGTM.

Had a question regarding [this comment on that issue thread](https://issuetracker.google.com/issues/36995652#comment24):
> TL;DR;
>
> Always call BluetoothDevice#connectGatt() with autoConnect=false (the race condition does not affect that case). If that connection times out you will get a callback with status=133. Then call BluetoothGatt#connect() to initiate a background connection.

Is that a strategy we should employ, or is it ok for our purposes to just always go in with `autoConnect=true`?

twyatt:
> Had a question regarding [this comment on that issue thread](https://issuetracker.google.com/issues/36995652#comment24):
> 
> > TL;DR;
> > Always call BluetoothDevice#connectGatt() with autoConnect=false (the race condition does not affect that case). If that connection times out you will get a callback with status=133. Then call BluetoothGatt#connect() to initiate a background connection.
> 
> Is that a strategy we should employ, or is it ok for our purposes to just always go in with `autoConnect=true`?

The race condition they are describing only affects Android <= 6 (or, API level <= 23, aka M). Kables minimum support Android version is 5, so the race condition will only hit apps (using Kable) and running on Android 5-6. According to https://apilevels.com/, those versions make up <3% of the phone population. Regardless, I've integrated a workaround for the bug linked, whereas we use reflection (essentially accessing private APIs) to forcefully make sure that the auto connect property is correct, see [Reflection.kt](https://github.com/JuulLabs/kable/blob/twyatt/auto-connect/core/src/androidMain/kotlin/Reflection.kt). Therefore, the race condition described in that issue won't hit the very small % of users on affected Android versions.

As for if we should always use `autoConnect = true`, it is slightly slower, but in my testing it **wasn't** noticeably slower. Because it never times it, the code that calls it tends to be simpler, and I'm hoping that makes for a more reliable user experience (despite being slightly slower to connect).

Additionally, you can get GATT 133 errors for a huge assortment of failures, so to use that to decide to switch to `autoConnect = true` doesn't feel consistent to me.

Ultimately, if we always use `autoConnect = true`, I believe we'll test that codepath more and it will ultimately be a better BLE experience.

The article linked in the PR description ([`autoConnect = true`](https://medium.com/@martijn.van.welie/making-android-ble-work-part-2-47a3cdaade07#7fcf) also recommends it as the "preferred way to reconnect". 🤷

cedrickcooke (approved):

twyatt:
@davidtaylor-juul I ended up backtracking on my decision to have the `autoConnect` be static for the life of a peripheral. I see value in the strategy described in the thread you linked.

Updated to have auto-connect configured via a lambda predicate so that library consumers can employ strategies like the one described in your linked thread.

PR #571, opened 2023-09-17
Prevent propagation of failures to `SharedRepeatableAction` parent scope

twyatt:


twyatt:
Deemed this unnecessary, as all the places where `SharedRepeatableAction` are used, its parent scope is a `SupervisorJob`. So, this PR just adds unnecessary complexity without providing any real value.

Issue #378, opened 2022-09-08
Peripheral blocking scope from finishing


tomnovotny7:
## Bug report

When using `scope.peripheral(...)` it will block scope from finishing.

I am developing an Android application and I've tried versions `0.18.0`, `0.17.2`, `0.16.1` and unreleased version `0.18.0-issue-359-1-SNAPSHOT` was crashing when using different contexts.

### How to reproduce

A coroutine is not finished unless the job is manually canceled.

```kotlin
val device: BluetoothDevice = ...

viewModelScope.launch {
  val peripheral = peripheral(device)

  // some work, connecting and disconnection or nothing
}
```

Even worse when using a different context:
```kotlin
val device: BluetoothDevice = ...

viewModelScope.launch {
  withContext(Dispatchers.IO) {
    val peripheral = peripheral(device)

    // some work, connecting and disconnection or nothing
  }
  
  // unreachable code
}
```

### Expected behavior

Either should not block scope and be closed automatically or should provide `peripheral.close()` method for closing the peripheral.

twyatt:
Thanks for the issue report! It is a good reminder that this portion of Kable is overdue for a redesign.

Short-term: I think providing `peripheral.close()` makes sense.

Long-term: I'd like to remove the reliance on a `CoroutineScope` within a `Peripheral`.

The nitty gritty details are: Initially `CoroutineScope` within a `Peripheral` was used as a crutch, to make passing data across threads simpler/feasible with the old memory model. I believe that with the new memory model, Kable can lean on the threads of the callbacks that are providing the various triggers that need to be acted on (rather than spinning up coroutines). For example, on Apple, we spin up a coroutine to monitor when BLE turns off (to disconnect peripherals) but we should be able to explicitly invoke the appropriate methods from the callbacks themselves (rather than emitting to a flow, to be collected and acted on). In the Android world, a coroutine is spun up to pass along characteristic change events from the current connection (which can change over time); instead of using the `Peripheral`'s `CoroutineScope`, Kable should be using the scope of the collector of characteristic change flow (and it should switch to grabbing events from the current connect, which may change).

twyatt:
I'm hoping to have the Android side of this issue resolved in [0.29.0](https://github.com/JuulLabs/kable/milestone/44), the other platforms will be fixed in later versions (I'll keep this issue open until problem is resolved on all platforms).

PR #573, opened 2023-09-19
Propagate failures from `connect` as failures rather than cancellation

twyatt:
The `SharedRepeatableAction` was propagating cancellation when being reset, causing callers of `connect` to also get cancelled. When `connect` should've been propagating whatever failure occurred when a error was hit. Allowing consumers to handle the exceptions and continue running.

---

**UPDATE:** In solving the above mentioned issue, ended up doing a larger _internal_ refactor to properly handle coroutine exceptions (and error propagation as intended). I tried to comment on the PR the additional changes that were made.


core/src/commonMain/kotlin/SharedRepeatableAction.kt
```diff
@@ -46,34 +48,50 @@ internal class SharedRepeatableAction<T>(
...
     }
 
-    fun reset() {
-        guard.withLock {
-            state?.apply { isCancelled = true }
-        }?.root?.cancel()
+    fun fail(cause: Throwable) {
+        setCancelled()
+            ?.root
+            ?.cancel(CancellationException(null, SharedRepeatableActionException(cause)))
```
twyatt, line 76: It would've been much cleaner to have `SharedRepeatableActionException` extend `CancellationException` and cancel with `SharedRepeatableActionException` directly, but I couldn't figure out how to extend `CancellationException`. 😢 

So, we have to do the dance on line 60, where we check the type of the `cause`.

README.md
```diff
@@ -279,17 +279,6 @@ attempt. The [`disconnect`] function suspends until the peripheral has settled o
...
-stall indefinitely. To prevent this (and ensure underlying resources are cleaned up in a timely manner) it is
-recommended that [`disconnect`] be wrapped with a timeout, for example:_
-
-```kotlin
-// Allow 5 seconds for graceful disconnect before forcefully closing `Peripheral`.
-withTimeoutOrNull(5_000L) {
-    peripheral.disconnect()
-}
-```
-
```
twyatt, line 292: This was an extra precaution that I don't think was needed.
In all my testing, the Android system always properly calls the callback when a device disconnects.

core/src/androidMain/kotlin/BluetoothDeviceAndroidPeripheral.kt
```diff
@@ -69,22 +68,14 @@ internal class BluetoothDeviceAndroidPeripheral(
...
+                invokeOnCompletion(::dispose)
+            },
+    )
 
-    init {
-        bluetoothState
-            .filter { state -> state == STATE_OFF }
-            .onEach { disconnect() }
-            .launchIn(scope)
-    }
```
twyatt, line 85: This was moved within `establishConnection`, so it will now match the lifecycle of a connection rather than living for the entire life of the peripheral.

This is a step in the process of resolving #378.

core/src/androidMain/kotlin/Connection.kt
```diff
@@ -24,20 +22,13 @@ public class OutOfOrderGattCallbackException internal constructor(
...
+    private val scope: CoroutineScope,
     internal val bluetoothGatt: BluetoothGatt,
     internal val dispatcher: CoroutineDispatcher,
     private val callback: Callback,
     logging: Logging,
 ) {
 
-    private val job = Job(scope.coroutineContext[Job]).apply {
-        invokeOnCompletion {
-            bluetoothGatt.close()
```
twyatt, line 36: Not needed (was never needed). We `close` the GATT object in the Callback. This was an unnecessary extra call to `close`.

core/src/androidMain/kotlin/BluetoothDeviceAndroidPeripheral.kt
```diff
@@ -159,19 +177,19 @@ internal class BluetoothDeviceAndroidPeripheral(
...
-        connectAction.resetAndJoin()
-        setDisconnected()
+        _connection?.bluetoothGatt?.disconnect()
+        suspendUntil<Disconnected>()
         logger.info { message = "Disconnected" }
     }
 
-    private fun close() {
-        connectAction.reset()
-        setDisconnected()
```
twyatt, line 169: This was unnecessary, as the state `Flow` is updated via `Callback`.

https://github.com/JuulLabs/kable/blob/63731b40f774d7d6b0715a6d487f5f21f9270422/core/src/androidMain/kotlin/gatt/Callback.kt#L106-L111

core/src/appleMain/kotlin/CBPeripheralCoreBluetoothPeripheral.kt
```diff
@@ -78,31 +83,22 @@ internal class CBPeripheralCoreBluetoothPeripheral(
...
 
     init {
-        centralManager.delegate
-            .state
-            .filter { state -> state == CBManagerStatePoweredOff }
-            .onEach {
-                disconnect()
-                _state.value = State.Disconnected()
-            }
-            .launchIn(scope)
```
twyatt, line 104: This was moved within `establishConnection`, so it will now match the lifecycle of a connection rather than living for the entire life of the peripheral.

This is a step in the process of resolving #378.

core/src/appleMain/kotlin/CBPeripheralCoreBluetoothPeripheral.kt
```diff
@@ -164,35 +160,64 @@ internal class CBPeripheralCoreBluetoothPeripheral(
...
+    }
+
+    private suspend fun closeConnection() {
+        withContext(NonCancellable) {
+            centralManager.cancelPeripheralConnection(cbPeripheral)
+        }
+    }
+
+    private fun dispose(cause: Throwable?) {
+        GlobalScope.launch(start = UNDISPATCHED) {
```
twyatt, line 209: `dispose` is called from an `invokeOnCompletion` (of the peripheral scope) so cannot be `suspend`.
Since everything is shutting down at that point, it seems fine to lean on `GlobalScope` to finish shutting down the connection. We need to run within a coroutine scope because `cancelPeripheralConnection` uses a dispatcher for executing the cancel peripheral connection request.

core/src/androidMain/kotlin/BluetoothDeviceAndroidPeripheral.kt
twyatt: Correction: we still need to explicitly set disconnected state if we only call `close` (and don't call `disconnect`) — which we do in some situations (e.g. BLE gets disabled or peripheral scope is disposed w/o first disconnecting peripheral).

davidtaylor-juul (approved):
LGTM, am wondering if the exception changes warrant marking this for a major version bump, but I don't have a strong opinion on that if you think minor is suitable.

twyatt:
> LGTM, am wondering if the exception changes warrant marking this for a major version bump, but I don't have a strong opinion on that if you think minor is suitable.

I've actually since updated this to be marked as `patch`, since from an API perspective: nothing has changed.

Although you're totally right that it should be communicated that this version includes major changes.
This PR is being bundled in the next release with other PRs that will result in a `minor` version bump (anything pre-1.0 is never technically `major` — but API guarantees are kind of out the window pre-1.0 as it is).

I'll be sure to add a warning to the next release, stating that major changes occurred and should be accompanied with lots of testing upon upgrading. I'm considering marking the next release with a `-rc` suffix to further communicate the need for caution when upgrading.

core/src/commonMain/kotlin/Throwable.kt
```diff
@@ -0,0 +1,13 @@
+package com.juul.kable
+
+import kotlinx.coroutines.CancellationException
+
+internal fun Throwable.unwrap(): Throwable {
```
cedrickcooke, line 5: This function is super confusing. I think I'd rather something like `Throwable.unwrapCancellationCause(): Throwable?` but even that doesn't really capture the semantics. Can we get a doc comment explaining the intent of what this unwraps into and when?

sdonn3 (approved):

cedrickcooke (approved):
I don't want to block a PR over a single `internal` function name, but I'd strongly prefer a change there

PR #582, opened 2023-10-04
Android 14 (compile SDK 34) and BroadcastReceiver changes

QuantumRand:
Upgrades compile-sdk to Android 14 (34) and makes required BroadcastReceiver changes (must include export flag when registering).

~Waiting on Tuulbox `7.0.0` release.~

Internal ticket: CA-2225


core/src/androidMain/kotlin/broadcastReceiverFlow.kt
```diff
@@ -1,23 +1,16 @@
...
-import kotlinx.coroutines.channels.awaitClose
+import com.juul.tuulbox.coroutines.flow.broadcastReceiverFlow
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.callbackFlow
 
+@Deprecated(
+    message = "Now provided by Tuulbox and no longer maintained as part of Kable.",
+    replaceWith = ReplaceWith("com.juul.tuulbox.coroutines.flow.broadcastReceiverFlow"),
+    level = DeprecationLevel.WARNING,
+)
```
twyatt, line 13: You can remove this deprecation. I created #583 to cover this.

```suggestion
```

gradle/libs.versions.toml
```diff
@@ -1,11 +1,11 @@
 [versions]
-android-compile = "33"
+android-compile = "34"
```
twyatt, line 2: This is bumping the target SDK as well?

If so, have you checked that Kable properly handles API 34?

Or, all changes necessary for API 34 are covered by this PR?

gradle/libs.versions.toml
```diff
@@ -16,6 +16,7 @@ kotlinx-coroutines-core = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-c
 kotlinx-coroutines-test = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-test", version.ref = "coroutines" }
 tuulbox-collections = { module = "com.juul.tuulbox:collections", version.ref = "tuulbox" }
 tuulbox-logging = { module = "com.juul.tuulbox:logging", version.ref = "tuulbox" }
+tuulbox-coroutines = { module = "com.juul.tuulbox:coroutines", version.ref = "tuulbox" }
```
twyatt, line 19: Can you move this to be alphabetically sorted?

core/build.gradle.kts
```diff
@@ -36,6 +36,7 @@ kotlin {
                 api(libs.kotlinx.coroutines.core)
                 api(libs.uuid)
                 implementation(libs.tuulbox.collections)
+                implementation(libs.tuulbox.coroutines)
```
twyatt, line 39: Since none of the other platforms need this, this should be moved to `androidMain` dependencies.

core/src/androidMain/kotlin/BluetoothState.kt
```diff
@@ -1,31 +1,16 @@
...
-            val state = intent.getIntExtra(EXTRA_STATE, ERROR)
-            channel.trySend(state)
-        }
-    }
-    applicationContext.registerReceiver(receiver, intentFilter)
-    awaitClose {
-        applicationContext.unregisterReceiver(receiver)
-    }
-}.shareIn(GlobalScope, started = WhileSubscribed(replayExpirationMillis = 0))
+internal val bluetoothState = broadcastReceiverFlow(intentFilter, RECEIVER_NOT_EXPORTED)
```
twyatt, line 15: nit: `RECEIVER_NOT_EXPORTED` is default, right?

```suggestion
internal val bluetoothState = broadcastReceiverFlow(intentFilter)
```

gradle/libs.versions.toml
twyatt: Looked into it, it does **not** bump the target SDK version.
Target defaults to **min** SDK version (see https://github.com/JuulLabs/sensortag/pull/90 for more details).

_For a library, the target SDK doesn't matter much, as the app's manifest will override the target version anyways, see [this Reddit comment](https://www.reddit.com/r/androiddev/comments/46ab5u/should_libraries_define_targetsdkversion/d03w7yd/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) for more details._

core/src/androidMain/kotlin/BluetoothState.kt
QuantumRand: Tons of conflicting information about this.
`ContextCompat` documentation says system BroadcastReceivers need to be exported:
```
flags – If this receiver is listening for broadcasts sent from the system or from other apps—even other apps that you own—use the RECEIVER_EXPORTED flag. If instead this receiver is listening only for broadcasts sent by your app, use the RECEIVER_NOT_EXPORTED flag.
```
 But Google says that's a security vulnerability and isn't allowed in Android 14 (system receivers should have no flag at all):
https://developer.android.com/about/versions/14/behavior-changes-14#system-broadcasts

Documentation in `Context` seems to be most up to date, so I'm sticking with that:
```
For apps targeting Build.VERSION_CODES.UPSIDE_DOWN_CAKE, either RECEIVER_EXPORTED or RECEIVER_NOT_EXPORTED must be specified if the receiver is not being registered for system broadcasts or a SecurityException will be thrown. See registerReceiver(BroadcastReceiver, IntentFilter, int) to register a receiver with flags.
```

twyatt: Maybe you misunderstood my comment? I was saying that we don't really need to specify the parameter because `RECEIVER_NOT_EXPORTED` is the default in Tuulbox: https://github.com/JuulLabs/tuulbox/blob/b592470845918d99e5d6c128243a5038e753fc38/coroutines/src/androidMain/kotlin/flow/BroadcastReceiverFlow.kt#L18

So, I wasn't saying we should not use `RECEIVER_NOT_EXPORTED`, rather: that flag will be used (by Tuulbox) if we omit the parameter?

core/build.gradle.kts
```diff
@@ -63,6 +63,7 @@ kotlin {
                 // Workaround for AtomicFU plugin not automatically adding JVM dependency for Android.
                 // https://github.com/Kotlin/kotlinx-atomicfu/issues/145
                 implementation(libs.atomicfu)
+                implementation(libs.tuulbox.coroutines)
```
twyatt, line 66: nit: newline between lines 65 and 66, to help clearly indicate that comment is only for line 65, and doesn't apply to this dependency as well

```suggestion
                // Workaround for AtomicFU plugin not automatically adding JVM dependency for Android.
                // https://github.com/Kotlin/kotlinx-atomicfu/issues/145
                implementation(libs.atomicfu)

                implementation(libs.tuulbox.coroutines)
```

twyatt (approved):

core/src/androidMain/kotlin/BluetoothState.kt
QuantumRand: I understood the comment, but it got me looking closer at the documentation since I realized it didn't jive with what Google's Android 14 migration docs said.

sdonn3 (approved):

PR #612, opened 2023-12-18
Add basic threading strategy support

### LINKED ISSUE #577: Spin down thread on disconnect
Currently, on Android, Kable spins up a dedicated thread for I/O per peripheral. The thread is not spun down until the peripherals scope is cancelled.

This is not intuitive for consumers. Instead, the thread should be spun down on disconnect (or released back to a pool). This will remove the requirement to dispose of the peripheral scope in order to prevent leaking threads.

twyatt:
Closes #577.


twyatt:
Note-to-self: Add KDoc to new `ThreadingStrategy` classes.

kable-core/src/androidMain/kotlin/ThreadingStrategy.kt
```diff
@@ -0,0 +1,92 @@
...
+    private val pool = mutableListOf<Pair<TimeMark, Threading>>()
+    private val guard = reentrantLock()
+
+    private val job = scope.launch {
+        try {
+            while (true) {
+                guard.withLock {
+                    pool.removeAll { (timeMark) -> timeMark.hasPassedNow() }
+                }
+                delay(evictAfter / 2)
```
QuantumRand, line 67: Why half the eviction time? Sqrt might be a little more flexible.

QuantumRand (approved):

kable-core/src/androidMain/kotlin/ThreadingStrategy.kt
twyatt: Was trying to find a balance between checking too often and not having threads stay alive for a lot longer than needed.

I do like `sqrt` for this, but unfortunately `evictAfter` is a `Duration`; whereas `div` is provided, but a `Duration.sqrt` function is not. I could convert to milliseconds for all durations, do the `sqrt` operation, then convert back, but that seems like an unnecessary amount of complexity for little gain?

Perhaps dividing by `3` would be better?

davidtaylor-juul (approved):

PR #652, opened 2024-03-10
Make `AndroidAdvertisement` parcelable

### LINKED ISSUE #610: Feature request: Parcelable Advertisement on Android
I wonder if you would consider making the `AndroidAdvertisement` implement `Parcelable`?

The [Android native ScanResult](https://developer.android.com/reference/android/bluetooth/le/ScanResult) implements the Parcelable interface, so I think this should be possible, though I don't know if there are internal details of the Advertisement implementation that make it more difficult.

The use case is an android view that shows a list of devices in range, implemented using Jetpack Compose. I would like to be able to seamlessly handle app configuration changes using `rememberSaveable`.

Yes, I could store the list of advertisements in a global singleton, outside of the view hierarchy, but that feels less in line with Android architectural idioms.

twyatt:
Closes #610.

### DIFF
--- a/core/src/androidMain/kotlin/AndroidAdvertisement.kt
+++ b/core/src/androidMain/kotlin/AndroidAdvertisement.kt
@@ -1,12 +1,14 @@
 package com.juul.kable
 
+import android.os.Parcelable
+
 @Deprecated(
     message = "Moved as nested class of `AndroidAdvertisement`.",
     replaceWith = ReplaceWith("AndroidAdvertisement.BondState"),
 )
 public typealias BondState = AndroidAdvertisement.BondState
 
-public interface AndroidAdvertisement : Advertisement {
+public interface AndroidAdvertisement : Advertisement, Parcelable {
 
     public enum class BondState {
         None,
--- a/core/api/android/core.api
+++ b/core/api/android/core.api
@@ -343,6 +343,14 @@ public final class com/juul/kable/ScanFailedException : java/lang/IllegalStateEx
 	public final fun getErrorCode ()I
 }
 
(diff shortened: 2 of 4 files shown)


cedrickcooke:
Any reason you're not just using the [parcelize](https://developer.android.com/kotlin/parcelize) plugin?

twyatt:
> Any reason you're not just using the [parcelize](https://developer.android.com/kotlin/parcelize) plugin?

🤦 I forgot about that plugin! TY!
I'll update the PR.

core/src/androidMain/kotlin/AndroidAdvertisement.kt
```diff
@@ -1,12 +1,14 @@
...
+import android.os.Parcelable
+
 @Deprecated(
     message = "Moved as nested class of `AndroidAdvertisement`.",
     replaceWith = ReplaceWith("AndroidAdvertisement.BondState"),
 )
 public typealias BondState = AndroidAdvertisement.BondState
 
-public interface AndroidAdvertisement : Advertisement {
+public interface AndroidAdvertisement : Advertisement, Parcelable {
```
QuantumRand, line 11: This makes this a `major` change, correct?

QuantumRand (approved):

core/src/androidMain/kotlin/AndroidAdvertisement.kt
twyatt: Thank you; yes.

cedrickcooke (approved):

davidtaylor-juul (approved):

PR #661, opened 2024-03-27
Re-align the filter shape to work correctly with the browser API

davidtaylor-juul:
The existing `sealed class Filter` type does not align with the browser filtering API which has this shape (as defined in `BluetoothLEScanFilterInit.kt`):
```kotlin
external interface BluetoothLEScanFilterInit {
    var services: Array<BluetoothServiceUUID>?
    var name: String?
    var namePrefix: String?
    var manufacturerData: Array<BluetoothManufacturerDataFilterInit>?
    var serviceData: Array<BluetoothServiceDataFilterInit>?
}
```

The current shape of the types prevents converting a `List<Filter>` to a `List<BluetoothLEScanFilterInit>` in a way that preserves the capabilities of the browser API. So instead we introduce a new type `FilterSet` that mirrors the `BluetoothLEScanFilterInit` object shape.

This PR deprecates the `Options.filters: List<KFilter>?` field and replaces it with the new `Options.filterSets: List<FilterSet>?` type.

To smooth the transition, if the deprecated `filters` field is provided things will operate as before. If `filters` is null and `filterSets` is provided, it will be used instead.


core/src/jsMain/kotlin/external/BluetoothServiceDataFilterInit.kt
```diff
@@ -9,6 +9,6 @@ package com.juul.kable.external
  *
  * https://webbluetoothcg.github.io/web-bluetooth/#device-discovery
  */
-internal external interface BluetoothServiceDataFilterInit {
-    val service: BluetoothServiceUUID
+internal external class BluetoothServiceDataFilterInit : BluetoothDataFilterInit {
+    var service: BluetoothServiceUUID
```
davidtaylor-juul, line 13: Although not implemented in Chrome, the API for this filter exists and can be constructed. I wrote a small test to try it out (it did nothing of course) but I did need to make this change to align it with the browser APIs.

core/src/jsMain/kotlin/FilterSet.kt
```diff
@@ -0,0 +1,45 @@
...
+
+import com.benasher44.uuid.Uuid
+import com.juul.kable.external.BluetoothLEScanFilterInit
+import com.juul.kable.external.BluetoothManufacturerDataFilterInit
+
+public data class FilterSet(
+    public val services: List<Filter.Service> = emptyList(),
+    public val name: Filter.Name? = null,
+    public val namePrefix: Filter.NamePrefix? = null,
+    public val manufacturerData: List<Filter.ManufacturerData> = emptyList(),
```
davidtaylor-juul, line 11: To be complete, we would have a `List<Filter.ServiceData>` property here as well, but seeing as that is not implemented in the browser I have omitted it to prevent people from attempting to use it (and subsequently finding out it doesn't work).

core/src/jsMain/kotlin/external/BluetoothServiceDataFilterInit.kt
twyatt: If this were an `interface`, can you not construct it via `jso<BluetoothServiceDataFilterInit> { .. }`?

Perhaps `class` is more aligned with `BluetoothDataFilterInit` being a `dictionary`?

It isn't clear to me what the trade-offs are for defining an external type as an `interface` vs. an `abstract class`. 🤔 

The [Kotlin documentation](https://kotlinlang.org/docs/wasm-js-interop.html#external-interfaces) states that:

> External interfaces don't have runtime type information and are a compile-time-only concept.

...in this particular case, we aren't doing `is` checks for `BluetoothDataFilterInit`, so I don't follow why this needs to change to a `class`?

twyatt (approved):

twyatt:
Thanks for addressing this! I see now the shortcomings of the previous API. 🧐

QuantumRand (approved):

core/src/jsMain/kotlin/external/BluetoothServiceDataFilterInit.kt
davidtaylor-juul: `BluetoothDataFilterInit` is an `abstract class`, so this also has to be a class in order to inherit. I'll see if we can change it, but it means also changing the definition for [BluetoothManufacturerDataFilterInit](https://github.com/JuulLabs/kable/blob/5d91057b4cf0173c4d9f4d4eed78f841854a11f4/core/src/jsMain/kotlin/external/BluetoothManufacturerDataFilterInit.kt#L12-L14) to also be an `interface`.

davidtaylor-juul: > Perhaps class is more aligned with BluetoothDataFilterInit being a dictionary?

It isn't a "dictionary" in the sense we would normally think of (i.e. a Kotlin `Map`). I think this is confusing nomenclature from the W3C when they refer to a "dictionary" for these definitions in Javascript it is just a normal object. Remember that in Javascript _technically_ all objects can be considered "dictionaries" in that you can access the instance members via stringly key lookup e.g.:
```Javascript
const person = {
  first: "Bob",
  last: "Smith"
};
console.log(thing['first']); 
```
Prints "Bob".

twyatt: > `BluetoothDataFilterInit` is an `abstract class`, so this also has to be a class in order to inherit. I'll see if we can change it, but it means also changing the definition for [BluetoothManufacturerDataFilterInit](https://github.com/JuulLabs/kable/blob/5d91057b4cf0173c4d9f4d4eed78f841854a11f4/core/src/jsMain/kotlin/external/BluetoothManufacturerDataFilterInit.kt#L12-L14) to also be an `interface`.

`interface` just "feels" cleaner, perhaps because it feels more like a blueprint, which aligns nicely with an `external` definition — though I don't have a concrete reason why it would be better.

The documentation I could find doesn't really clearly state the pros/cons of either approach.

So, I'm fine with either `abstract class` or `interface` purely because I don't actually know which one is more "correct".

> but it means also changing the definition for [BluetoothManufacturerDataFilterInit](https://github.com/JuulLabs/kable/blob/5d91057b4cf0173c4d9f4d4eed78f841854a11f4/core/src/jsMain/kotlin/external/BluetoothManufacturerDataFilterInit.kt#L12-L14) to also be an `interface`.

I think it would be fine to change `BluetoothManufacturerDataFilterInit` as well (if needed), because it is `internal` so consumer won't be affected by the change.

davidtaylor-juul: Have changed to `interface` 👍

PR #665, opened 2024-04-03
Fix crash in peripheral `<init>` when parent job is already complete

cedrickcooke:
This PR doesn't attempt any big refactors re. scoping, just makes sure that `dispose`/`finalCleanup` are called near the end of peripheral initialization to avoid unexpected nulls.


twyatt (approved):

davidtaylor-juul (approved):
Adding "asynchronous constructor side-effects" to my list of nightmares 😆

PR #688, opened 2024-06-06
K2 support

twyatt:
Changes needed to get #683 compiling (i.e. changes needed to support K2).

Dropped per platform `Scanner` interfaces, instead `Scanner` now carries a generic (`Scanner<T : Advertisement>`) of what advertisement type it will emit via its `advertisements` flow. Each platform scanner implements the common `Scanner<T : Advertisement>` interface rather than each platform (e.g. `AndroidScanner`) interface. This change was necessary because K2 does not allow `expect`/`actual` interfaces to differ (signatures must match exactly). Whereas `AndroidScanner` would have an `advertisements` flow of type `Flow<AndroidAdvertisement>` which did not match the `expect` interface that had `Flow<Advertisement>`.

On JS, some `external` classes (that were used in `actual` type aliases) were made concrete (not `abstract`) in order to match the `expect` side (`PlatformService`, `PlatformCharacteristic`, `PlatformDescriptor`).

Special thanks to @cedrickcooke for helping me wrangle a solution that minimized the need for larger API changes.

### DIFF
--- a/core/api/android/core.api
+++ b/core/api/android/core.api
@@ -11,21 +11,6 @@ public abstract interface class com/juul/kable/Advertisement {
 	public abstract fun serviceData (Ljava/util/UUID;)[B
 }
 
-public abstract interface class com/juul/kable/AndroidAdvertisement : android/os/Parcelable, com/juul/kable/Advertisement {
-	public abstract fun getAddress ()Ljava/lang/String;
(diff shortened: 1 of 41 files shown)


davidtaylor-juul (approved):
🎉

sdonn3 (approved):

PR #699, opened 2024-06-20
Ensure teardown is completed when observation coroutines are canceled

### LINKED ISSUE #677: Cannot reobserve a characteristic after canceling a job containing a call to observe
I have the same use case as in Q&A answer https://github.com/JuulLabs/kable/discussions/592 except that I sometimes need to re-observe the characteristic during the same connection.

Using kable 0.31.1 on Android, I did exactly as the answer suggested, using `cancelAndJoin()` to cancel a `launch`ed job. This does successfully write `DISABLE_NOTIFICATION_VALUE` to the descriptor. However, when I call `observe()` on the characteristic again, kable seems to think the characteristic is already being observed, and does not write `ENABLE_NOTIFICATION_VALUE`.

I've tracked this down (using the debugger) to the fact that `didStartObservation` is never set to `false` -- I set a breakpoint at [this line](https://github.com/JuulLabs/kable/blob/7a58126d6b61d4ff2b014fb7fb2e834560357374/core/src/commonMain/kotlin/Observation.kt#L76), and it is never reached. AFAICT, this is because the cancellation of the job causes the next `suspend` function, which happens to be [`BluetoothDeviceAndroidPeripheral.setConfigDescriptor()`](https://github.com/JuulLabs/kable/blob/7a58126d6b61d4ff2b014fb7fb2e834560357374/core/src/androidMain/kotlin/BluetoothDeviceAndroidPeripheral.kt#L362), to throw a `CancellationException` on completion, which is propagated back to [the previous line](https://github.com/JuulLabs/kable/blob/7a58126d6b61d4ff2b014fb7fb2e834560357374/core/src/commonMain/kotlin/Observation.kt#L74) in `Observation.kt`.

mmeisel:
It turns out that any suspending function called after cancelation of the coroutine will itself be canceled. [The kotlin documentation](https://kotlinlang.org/docs/cancellation-and-timeouts.html#run-non-cancellable-block) recommends using `withContext(NonCancellable) {}` in these cases.

Fixes #677

I wanted to write a test for this, but the observation tests don't use any suspend functions so it wasn't clear to me how. It did fix the problem for me, however.

### DIFF
--- a/core/src/commonMain/kotlin/Observers.kt
+++ b/core/src/commonMain/kotlin/Observers.kt
@@ -83,7 +83,9 @@ internal class Observers<T>(
             .mapNotNull { event -> (event as? CharacteristicChange)?.data }
             .onCompletion {
                 try {
-                    // If this suspending function is canceled the peripheral could end up in an indeterminate state.
+                    // `NonCancellable` used to prevent interruption of resetting the observation
+                    // state, which can prevent subsequent re-observation.
+                    // See https://github.com/JuulLabs/kable/issues/677 for more details.
                     withContext(NonCancellable) {
                         observation.onCompletion(onSubscription)
                     }
--- a/core/src/commonMain/kotlin/Observers.kt
+++ b/core/src/commonMain/kotlin/Observers.kt
@@ -5,6 +5,7 @@ import com.juul.kable.ObservationEvent.Error
 import com.juul.kable.logs.Logging
 import kotlinx.atomicfu.locks.SynchronizedObject
 import kotlinx.atomicfu.locks.synchronized
+import kotlinx.coroutines.NonCancellable
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.filter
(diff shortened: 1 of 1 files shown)


twyatt:
Thanks for the PR @mmeisel! I'll look it over this weekend.

core/src/commonMain/kotlin/Observers.kt
```diff
@@ -81,7 +83,10 @@ internal class Observers<T>(
             .mapNotNull { event -> (event as? CharacteristicChange)?.data }
             .onCompletion {
                 try {
-                    observation.onCompletion(onSubscription)
+                    // If this suspending function is canceled the peripheral could end up in an indeterminate state.
```
twyatt, line 86: nitpick

```suggestion
                    // `NonCancellable` used to prevent interruption of resetting the observation
                    // state, which can prevent subsequent re-observation.
                    // See https://github.com/JuulLabs/kable/issues/677 for more details.
```

twyatt (approved):
I did not test it, but I understand why this should fix the issue. 👍 
Only minor ask I have is to update the code comment slightly.

Phoenix7351 (approved):

twyatt:
@mmeisel I plan on testing this in the next day or two. If everything works as expected then I'll merge and cut a release. Thanks again for the PR!

mmeisel:
Thank *you* @twyatt, happy to help!

PR #704, opened 2024-06-25
Discover descriptors on connect

FabioCornelli:
Hi,

first of all thank you for this library! 
it's really helpful for managing BLE across different platforms using kotlin common code.

We are trying to use Kable in our mobile app (android & iOS) and we noticed an issue / unexpected behavior.

It seems that on Apple platforms, after establishing a connection to a peripheral , descriptors are not discovered.

This is causing issues / unexpected behaviors, because on other platform (e.g. Android) all services, characteristics and descriptors are discovered when calling `peripheral.connect()`

For example, imagine a code like this in **commonMain** kotlin code:

```kotlin
val advertisement = scanner.advertisements.first()
val peripheral = coroutineScope.peripheral(advertisement)
peripheral.connect()
peripheral.services?.forEach { service ->
    println("service ${service.serviceUuid}")
    service.characteristics.forEach { characteristic ->
        println("characteristic ${characteristic.characteristicUuid}")
        val descriptors = characteristic
            .descriptors
            .map { descriptor -> peripheral.read(descriptor).decodeToString() }
        println("descriptors -> $descriptors")
    }
}
```

this code will have different behavior and will perform different BLE operation.

On Android it will discover all services, characteristics and read all descriptors.

On iOS it will only discover all services and characteristics, but will not discover nor read any descriptor.

With this PR I've tried to align the behavior on apple platforms.

I've also fixed  `fun readAsNSData(descriptor: Descriptor ): NSData`  because it assumed that descriptor was always in NSData format, but it wasn't actually always the case.

Let me know if this PR seems ok and can be merged or if it needs any modification.
 
Thank you!

PR #706, opened 2024-07-02
Fix reading of descriptor value as `NSData`

FabioCornelli:
Hi,

this PR originated from #704 , but it is solving a separate issue, hence this PR.

The problem is that, on apple platform, descriptor value is force casted to NSData, but seems that `descriptor.value` type depends on `descriptor.UUID.UUIDString`. 

Apple documentation & online references about this:

- https://developer.apple.com/documentation/corebluetooth/characteristic-descriptors  
- https://stackoverflow.com/questions/52816049/how-to-convert-cbdescriptor-value-to-string

From a quick Look at the code, this issue seems to affect only apple platforms.

On Apple `descriptor.value` is returned as generic object `Any?` 
 by the platform [as documented here](https://developer.apple.com/documentation/corebluetooth/cbdescriptor/value)

On Android `descriptor.value` is returned as `byte[]` by the platform [as documented here](https://developer.android.com/reference/android/bluetooth/BluetoothGattDescriptor#getValue())

On JS `descriptor.value` is returned as `ArrayBuffer` by the platform [as documented here](https://developer.mozilla.org/en-US/docs/Web/API/BluetoothRemoteGATTDescriptor/value)

So we need to properly cast to `ByteArray` only on apple platforms.

Let me know if you think some changes are required


core/src/appleMain/kotlin/CBPeripheralCoreBluetoothPeripheral.kt
```diff
@@ -378,9 +387,42 @@ internal class CBPeripheralCoreBluetoothPeripheral(
...
+
+        // CBDescriptor "value" property type depends on cbDescriptor.UUID.UUIDString
+        // https://stackoverflow.com/questions/52816049/how-to-convert-cbdescriptor-value-to-string
+        // https://developer.apple.com/documentation/corebluetooth/characteristic-descriptors
+        when (updatedDescriptor.UUID.UUIDString) {
+            CBUUIDCharacteristicFormatString -> {
+                return updatedDescriptor.value as NSData
+            }
+            CBUUIDCharacteristicAggregateFormatString,
+            CBUUIDCharacteristicUserDescriptionString,
```
davidtaylor-juul, line 402: For `CBUUIDCharacteristicAggregateFormatString` the apple docs do not specify what the contents of the descriptor value are. Do we know for sure this one has `String` content?

FabioCornelli: No actually I'm not sure, I kind of trusted the  [StackOverflow answer](https://stackoverflow.com/questions/52816049/how-to-convert-cbdescriptor-value-to-string)  for that value.

core/src/appleMain/kotlin/CBPeripheralCoreBluetoothPeripheral.kt
```diff
@@ -378,9 +387,42 @@ internal class CBPeripheralCoreBluetoothPeripheral(
...
+            -> {
+                return (updatedDescriptor.value as String)
+                    .encodeToByteArray()
+                    .toNSData()
+            }
+            CBUUIDCharacteristicExtendedPropertiesString,
+            CBUUIDClientCharacteristicConfigurationString,
+            CBUUIDServerCharacteristicConfigurationString,
+            CBUUIDL2CAPPSMCharacteristicString,
```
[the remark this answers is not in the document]
FabioCornelli: Yes Apple API is awkward, indeed.
 I've also searched online a bit and the only useful doc I've found is [this one](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-54/out/en/host/generic-attribute-profile--gatt-.html), but still is not clear to me which is that correct type we should expect for each different value.

`Do you have a real-world example to test this one on, as that would be ideal, otherwise I suppose we go with our best-guess/stack-overflow comments?`

i've tested with an IOT device that mount an ESP32 BLE Board on it. 
I'm using Kable to rewrite in Kotlin Multiplatform this [Espressif SDK](https://github.com/espressif/esp-idf-provisioning-android) which reads descriptors and use it for different kind of messages.

I've also tested with a Xiaomi Mi Smart Band 5 (nothing work related, its just that i have it with me), but since is not something I need for work, I've just verified that the code didn't crash.

This week I'm kind of busy at work, but maybe in the weekend I can run some tests on these two devices and provide you some extra information about this.

FabioCornelli:
I took some time to dig in the [Gatt Part G document](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-54/out/en/host/generic-attribute-profile--gatt-.html) and match the information  there with [Apple documentation](https://developer.apple.com/documentation/corebluetooth/characteristic-descriptors):

| Apple UUID constants | UUID value | Apple Documented value Type  | GATT Part G document | Tested  |
| :------------ |:---------------: |:---------------:|:---------------:|:---------------:|
| CBUUIDCharacteristicExtendedPropertiesString   | 2900 | NSNumber | 3.3.3.1 - Attribute Value shall be two octets in length and shall contain the Characteristic Extended Properties Bit Field | ✅ |
| CBUUIDCharacteristicUserDescriptionString  | 2901  |   NSString |  3.3.3.2 - Attribute Value shall be set to the characteristic user description UTF-8 string. |  ✅ |
| CBUUIDClientCharacteristicConfigurationString | 2902 |  NSNumber |  3.3.3.3 - The Attribute Value shall be two octets in length and shall be set to the characteristic descriptor value |  ✅  |
|CBUUIDServerCharacteristicConfigurationString | 2903  |  NSNumber |  3.3.3.4 - The Attribute Value shall be two octets in length and shall be set to the characteristic descriptor value.  |  ❌  |
|CBUUIDCharacteristicFormatString (Presentation Format ) | 2904  | NSData |   3.3.3.5 - The characteristic presentation format  value is composed of five parts: format (1 octet), exponent (1 octet), unit (2 octets), name space (1 octet), and description (2 octet). |   ❌  |
|CBUUIDCharacteristicAggregateFormatString  |  2905  |  Not documented 👎🏻  |   3.3.3.6 - The Characteristic Aggregate Format value is composed of a list of Attribute Handles of Characteristic Presentation Format declarations, where each Attribute Handle points to a Characteristic Presentation Format declaration.  |  ❌  |
| CBUUIDL2CAPPSMCharacteristicString  |  ???        |   UInt16 |   The document mention L2CAP, but i didn’t find any useful information.  |   ❌  |


Given these information it seems that:

- CBUUIDCharacteristicExtendedPropertiesString, CBUUIDClientCharacteristicConfigurationString, CBUUIDServerCharacteristicConfigurationString and  CBUUIDL2CAPPSMCharacteristicString are all two bytes in length.
- CBUUIDL2CAPPSMCharacteristicString is returned by Apple as UInt16 instead of NSNumber, so it should be handled separately
- CBUUIDCharacteristicAggregateFormatString while is not documented by Apple, I think we can assume it will be returned as NSData since it is an aggregation of  CBUUIDCharacteristicFormatString, but of course if we can't be sure until we test it.


As shown in the table above, I was able to test some of the differente descriptor types and the value was read it correctly.

I've updated the PR to reflect these changes, let me know your feedback!

twyatt:
Thanks for being so thorough and deep-diving with the testing! I'll review this PR soon.

core/src/appleMain/kotlin/CBPeripheralCoreBluetoothPeripheral.kt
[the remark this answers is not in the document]
davidtaylor-juul: Thanks for the detail. I agree with your deductions in the other comment as well. The only reference I found in the docs is for the type alias https://developer.apple.com/documentation/corebluetooth/cbl2cappsm/ which also hints that this the value for `CBUUIDL2CAPPSMCharacteristicString` is exposed as a raw 16bit integer, and is not a `NSNumber` or `NSData` object. It is for `L2CAP` so to test a real world example would require reading characteristics from a device that supports `L2CAP` channels. I'll update this thread if I find a concrete example.

core/src/appleMain/kotlin/CBPeripheralCoreBluetoothPeripheral.kt
```diff
@@ -393,9 +402,47 @@ internal class CBPeripheralCoreBluetoothPeripheral(
...
+            CBUUIDClientCharacteristicConfigurationString,
+            CBUUIDServerCharacteristicConfigurationString,
+            -> {
+                return (updatedDescriptor.value as NSNumber)
+                    .unsignedShortValue
+                    .toNSData()
+            }
+            CBUUIDL2CAPPSMCharacteristicString,
+            -> {
+                return (updatedDescriptor.value as UInt16).toNSData()
```
davidtaylor-juul, line 432: The docs are saying that the `value` here is `UInt16`, but I doubt the API expects us to dereference/typecast directly as a raw pointer. It's not completely out of the question but feels a little odd for Apple to do that? My best guess is the is already an `NSData` (of two bytes length) so we should just return as `NSData` here.

I'm digging around for a device that exposes this characteristic so I can test a real example to know for sure.

twyatt: Not sure if it confirms your suspicions @davidtaylor-juul, but [this code](https://github.com/clostra/newnode/blob/6e64362fb2dd97cc58ee4b1624c466a85a43dc9f/ios/Bluetooth.m#L196-L231) retrieves the `NSData` as-is (without `UInt16` casting, as far as I can tell, at least).

core/src/appleMain/kotlin/CBPeripheralCoreBluetoothPeripheral.kt
```diff
@@ -393,9 +402,47 @@ internal class CBPeripheralCoreBluetoothPeripheral(
...
+                logger.warn {
+                    message = "Best effort descriptor value conversion for undocumented $CBUUIDCharacteristicAggregateFormatString uuid"
+                }
+                return (updatedDescriptor.value as? NSData) ?: byteArrayOf().toNSData()
+            }
+            else -> {
+                logger.warn { message = "cannot read descriptor for unknown uuid string ${updatedDescriptor.UUID.UUIDString}" }
+                return byteArrayOf().toNSData()
+            }
+        }
```
twyatt, line 445: As I was looking how others read descriptors, I stumbled on [this](https://github.com/vokal/flutter_blue/blob/3b6f1552d87554f47580449b16e8f3252b1832ff/ios/Classes/FlutterBluePlugin.m#L555-L565):

```objc
// TODO: Convert NSNumber types too
if( [descriptor.value isKindOfClass:[NSData class]] ) {
    return descriptor.value;
}
if( [descriptor.value isKindOfClass:[NSString class]] ) {
    return [descriptor.value dataUsingEncoding:NSUTF8StringEncoding];
}
if( [descriptor.value isKindOfClass:[NSNumber class]] ) {
    int value = [descriptor.value intValue];
    return [NSData dataWithBytes:&value length:sizeof(value)];
}
```

Which hints that the data we get from Core Bluetooth carries the type info, meaning we could possibly do something similar to the code snippet above?

```suggestion
        return when (val value = updatedDescriptor.value) {
            is NSData -> value
            is NSString -> value.dataUsingEncoding(NSUTF8StringEncoding)
                ?: byteArrayOf().toNSData().also {
                    logger.warn {
                        message = "Failed to decode descriptor"
                        detail("type", "NSString")
                        detail(descriptor)
                    }
                }
            is NSNumber -> memScoped {
                NSData.dataWithBytes(
                    bytes = alloc<IntVar>().also { it.value = value.intValue }.ptr,
                    length = sizeOf<IntVar>().convert(),
                )
            }
            else -> byteArrayOf().toNSData().also {
                logger.warn {
                    message = "Unable to read descriptor of unknown type"
                    detail(descriptor)
                }
            }
```

_Though, I'm not sure what is meant by the `// TODO: Convert NSNumber types too` code comment._

twyatt: Additional instances where the type of the descriptor value was used to process it:
- [FlutterBluePlusPlugin.m#L2137-L2160](https://github.com/boskokg/flutter_blue_plus/blob/0d439a98d5b5ce76ce51a95c9ca93e03e04c1e8a/ios/Classes/FlutterBluePlusPlugin.m#L2137-L2160)
- [BleManager.m#L121-L142](https://github.com/AndrewDongminYoo/react-native-ble-manager/blob/80634af6ac4cda7e203fd936812c2efe3820ef1e/ios/BleManager.m#L121-L142)

twyatt (commented):
I'm not an expert on iOS BLE, but I may have stumbled on a simpler approach to processing the descriptor value?

I did not test it yet, but wanted to share my findings to get your thoughts.

core/src/appleMain/kotlin/CBPeripheralCoreBluetoothPeripheral.kt
FabioCornelli: Yes this approach of checking types directly can work as well.
It was actually my first implementation, but then I opted on checking the `descriptor.UUID.UUIDString` as it was what apple doc suggested and It probably felt less _arbitrary_ to somebody looking at code for the first time.

> Though, I'm not sure what is meant by the // TODO: Convert NSNumber types too code comment.

I guess this is a drawback of this approach.
I think that what the comment is trying to say is that `NSNumber` is always being converted to an `Int` (i.e. 4 bytes), but it is actually a class that can contain any type of number, such as `Long` (8 bytes), `Short` (2 bytes) etc.
This means that if `NSNumber` actually contains a `Long`, then we lost 4 bytes of information by converting to `Int`. 

This approach is also used on [Dotnet](https://github.com/dotnet-bluetooth-le/dotnet-bluetooth-le/blob/master/Source/Plugin.BLE/Apple/Descriptor.cs) which  tries to workaround this issue by casting to `UInt64` (the biggest possible number type, so no bytes are lost).

But even with or without Dotnet workaround, there a still probably some inconsistencies across platforms.

For example take `CBUUIDClientCharacteristicConfigurationString`.
The Bluetooth specification states:

> The Attribute Value shall be two octets in length and shall be set to the characteristic descriptor value

But with this approach we will return  either 4 bytes (without Dotnet workaround) or 8 bytes (with Donet workaround), when the specification states that should be 2 bytes.
Furthermore, on other platforms (i.e. Android, JS) 2 bytes are returned (I didn't test, it's just my guess, but seems very likely)

I'm not trying to dismiss this approach as it is indeed simpler, less verbose and seems to be widely adopted by other libraries / Multiplatform frameworks, but just wanted to highlight some of the potentials limitations.

twyatt: Thanks for the insight! I'm not attached to one approach over the other. As you've identified, they both have trade offs.

Perhaps a mix of the two approaches might make the most sense? Where we first process based on type, then when an `NSNumber`: determine what size number to process as based on UUID.

Using only UUID, as you currently have it, does seem like a really good approach (and it's nice to follow recommendations set forth by the documentation) but being that the documentation leaves a lot to be desired (at times it isn't clear exactly what type will be provided), it feels safer to process by type first (for the non-number types). 🤷‍♂️

FabioCornelli: > Perhaps a mix of the two approaches might make the most sense? Where we first process based on type, then when an NSNumber: determine what size number to process as based on UUID.

Yes I guess we could go along this way since we couldn't find anywhere the expected type returned by Apple.

I've updated the PR by reworking the conversion approach:

- For types that can be converted directly to NSData without ambiguity (i.e. NSString, NSData, UInt16), I convert it directly without further checks on UUID string
- For NSNumber (that can't be converted directly to NSData without ambiguity):
  -  I convert it to the documented numeric type by checking on the UUID string where possible
  - otherwise I convert it to `ULong` to avoid information loss (Dotnet approach)

Let me know your feedback!

core/src/appleMain/kotlin/CBPeripheralCoreBluetoothPeripheral.kt
```diff
@@ -393,9 +401,41 @@ internal class CBPeripheralCoreBluetoothPeripheral(
...
+            is NSNumber -> {
+                when (updatedDescriptor.UUID.UUIDString) {
+                    CBUUIDCharacteristicExtendedPropertiesString,
+                    CBUUIDClientCharacteristicConfigurationString,
+                    CBUUIDServerCharacteristicConfigurationString,
+                    -> value.unsignedShortValue.toNSData()
+                    else -> value.unsignedLongValue.toNSData()
+                }
+            }
+            is UInt16 -> value.toNSData()
```
twyatt, line 429: I'm guessing you have this here for `CBUUIDL2CAPPSMCharacteristicString`?
Although to documentation lists it as `UInt16`, I'd be surprised if this didn't come in as an `NSNumber`?
Aside from [CBUUIDL2CAPPSMCharacteristicString](https://developer.apple.com/documentation/corebluetooth/cbuuidl2cappsmcharacteristicstring), did you see any indication of what time this comes in as?

FabioCornelli: Yes I've kept `Uint16` for `CBUUIDL2CAPPSMCharacteristicString`

> Aside from [CBUUIDL2CAPPSMCharacteristicString](https://developer.apple.com/documentation/corebluetooth/cbuuidl2cappsmcharacteristicstring), did you see any indication of what type this comes in as?

Only other reference was found by @davidtaylor-juul

>  Thanks for the detail. I agree with your deductions in the other comment as well. The only reference I found in the docs is for the type alias https://developer.apple.com/documentation/corebluetooth/cbl2cappsm/ which also hints that this the value for `CBUUIDL2CAPPSMCharacteristicString` is exposed as a raw 16bit integer, and is not a `NSNumber` or `NSData` object. It is for `L2CAP` so to test a real world example would require reading characteristics from a device that supports `L2CAP` channels. I'll update this thread if I find a concrete example.

and seems to indicate that it might return as `UInt16`, since `CBL2CAPPSM` is an alias for `UInt16`.

On the other hand, what you have found here (although it is for `CBCharacteristic` [value property](https://developer.apple.com/documentation/corebluetooth/cbcharacteristic/value) which is defined as `NSData` and not for `CBDescriptor` value property which is defined as `Any`)

> Not sure if it confirms your suspicions @davidtaylor-juul, but [this code](https://github.com/clostra/newnode/blob/6e64362fb2dd97cc58ee4b1624c466a85a43dc9f/ios/Bluetooth.m#L196-L231) retrieves the NSData as-is (without UInt16 casting, as far as I can tell, at least).

Seems to indicate it might be returned as `NSData`.

Given all these considerations, I think I can add management for `CBUUIDL2CAPPSMCharacteristicString` in the `NSNumber` block so that:

- if it is actually returned as `NSData` as indicated by the code you found for  `CBCharacteristic`  we are already covered by the current code. It will be returned directly
- if it is actually a `UInt16` as Apple doc hints, we are already covered by the current code. It will be converted and returned.
- if it is, for some reason,  actually returned as `NSNumber` I will add specific management and convert it to a two bytes array as documented by GATT Part G document.

I'll update the PR in a moment, let me know what you think about this.

core/src/appleMain/kotlin/CBPeripheralCoreBluetoothPeripheral.kt
```diff
@@ -393,9 +402,42 @@ internal class CBPeripheralCoreBluetoothPeripheral(
...
+                    detail(descriptor)
+                }
+            }
+            is NSNumber -> {
+                when (updatedDescriptor.UUID.UUIDString) {
+                    CBUUIDCharacteristicExtendedPropertiesString,
+                    CBUUIDClientCharacteristicConfigurationString,
+                    CBUUIDServerCharacteristicConfigurationString,
+                    CBUUIDL2CAPPSMCharacteristicString,
+                    -> value.unsignedShortValue.toNSData()
```
twyatt, line 427: This feels like a good approach to me. Thanks for the thorough explanation around this. ❤️

twyatt (approved):
I plan to add a few code comments and make some minor code style related changes prior to merging, but this looks great to me! Thanks so much @FabioCornelli!

I'll get this in after we get another approval on it.

core/src/appleMain/kotlin/CBPeripheralCoreBluetoothPeripheral.kt
davidtaylor-juul: IMHO we could omit this case, but I suppose there is no harm leaving it in.

davidtaylor-juul (approved):
I think we arrived at a sensible solution here - huge thank you for unravelling that stuff!

twyatt:
> I plan to add a few code comments and make some minor code style related changes prior to merging

I'll try to find some time to do this in the coming days, then get this merged in.

PR #722, opened 2024-07-26
Add `Options` builder

### LINKED ISSUE #723: Provide `Options` DSL
Rather than consumers constructing `Options` class via its constructor, `Options { }` builder should be used, for example:

```kotlin
val options = Options {
    filters {
        match {
            name = Filter.Name.Exact("Example")
        }
        optionalServices = listOf(..)
    }
}
```

twyatt:
Brings JavaScript-side API (of specifying filters) in line with the other platforms (see #695) by providing an `Options` builder that contains a `filters { }` bulder.

Closes #723

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ The [`Scanner`] may be configured via the following DSL (shown are defaults, whe
 val scanner = Scanner {
     filters {
         match {
-            name = "My device"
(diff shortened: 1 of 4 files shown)


davidtaylor-juul (approved):
🚀

Phoenix7351 (approved):

PR #727, opened 2024-07-30
Fix closing parenthesis in `Options` builder example in README

khebrati:
fixed closing parenthesis in options builder example for Javascript.


twyatt (approved):
Thanks!

Issue #724, opened 2024-07-26
Restructure exception hierarchy


twyatt:
> You should design your own general-purpose Kotlin APIs in the same way: use exceptions for logic errors, type-safe results for everything else. **Don’t use exceptions as a work-around to sneak a result value out of a function.**[^1]

- `IllegalStateException`
- [ ] `IOException` [common]
    - [ ] `GattException` [common]
        - [ ] `GattRequestRejectedException` [android]
        - [ ] `GattStatusException(status: GattStatus)` [android]
        - [ ] `GattWriteException(result: WriteResult)` [android]
    - [ ] `NotConnectedException` [common]
    - [ ] `UnmetRequirementException(reason: UnmetRequirementReason)` [common]
        - `BluetoothDisabled` [common]
        - `LocationManagerUnavailable` [android]
        - ..
- [ ] `InternalError` [common]

| Exception | Purpose |
|-|-|
| `IllegalStateException` | _(and sub-classes of)_ shall be thrown when an application state is **statically** not as expected (failure state won't exhibit a race condition). For example: "bluetooth not supported" (this state will not change), "missing required bluetooth permissions" (Android and iOS both kill app when de-escalating permissions — a developer can check permissions). When these exceptions are thrown, they are intended to crash the consuming app so that consumers are alerted of the logic error(s) and can address them accordingly. |
| `IOException` | "[..] not logic errors in code. They typically represent external conditions that the code has no control over but must deal with."[^2] `IOExceptions` are expected to be caught and handled accordingly (e.g. retried or inform user of steps to resolve the issue, such as enabling bluetooth). |

### kotlinx.serialization

- IllegalArgumentException
    - public open SerializationException
        - internal CborDecodingException
        - internal open JsonException
            - internal JsonDecodingException
            - internal JsonEncodingException
        - internal ProtobufDecodingException
        - internal UnknownFieldException
        - public MissingFieldException

### kotlinx.coroutines

- IllegalStateException
    - public open CancellationException
        - internal AbortFlowException
        - internal ChildCancelledException
        - internal JobCancellationException
        - public TimeoutCancellationException
    - public ClosedSendChannelException
- NoSuchElementException
    - public ClosedReceiveChannelException
- RuntimeException
    - public CompletionHandlerException

[^1]: https://elizarov.medium.com/kotlin-and-exceptions-8062f589d07#2273
[^2]: https://elizarov.medium.com/kotlin-and-exceptions-8062f589d07#ef58

[...]

Issue #721, opened 2024-07-26
Consider leveraging `JetBrains/kotlin-wrappers` library


twyatt:
[JetBrains/kotlin-wrappers](https://github.com/JetBrains/kotlin-wrappers) can provide types such as [`DOMException`](https://github.com/JetBrains/kotlin-wrappers/blob/master/kotlin-web/src/jsMain/generated/web/errors/DOMException.kt), for example:

https://github.com/JuulLabs/kable/blob/06402413693844653ccc2dff5ec0db3d3f4ca8d1/kable-core/src/jsMain/kotlin/BluetoothDeviceWebBluetoothPeripheral.kt#L336-L344

twyatt:
Closing as #749 extensively leverages `kotlin-wrappers` and it will continue to be used for future JavaScript changes.

Issue #286, opened 2022-03-24
Move `CoroutineScope` as function parameter for `Peripheral` creation


twyatt:
See https://github.com/JuulLabs/kable/pull/280#issuecomment-1077950078 for discussion that spawned this decision.

In summary, we should change `Peripheral` creation functions to have a signature such as:

```kotlin
public expect fun peripheral(
    scope: CoroutineScope,
    advertisement: Advertisement,
    builderAction: PeripheralBuilderAction = {},
): Peripheral
```

This should include a deprecation cycle for the current signatures, such as:

```kotlin
@Deprecated(
    message = "Use `peripheral` function with CoroutineScope as a function parameter.",
    replaceWith = "peripheral(scope, advertisement, builderAction)",
)
public expect fun CoroutineScope.peripheral(
    advertisement: Advertisement,
    builderAction: PeripheralBuilderAction = {},
): Peripheral
```

twyatt:
Realized that we're potentially handing library consumers a footgun by requiring that consumers specify a `CoroutineScope`: it seems common that consumers manage their peripheral lifecycles manually (i.e. call `disconnect` when no longer needed — which is the preferred approach) but since creating a `Peripheral` requires specifying a `CoroutineScope` (to be a child of), leads consumers to often pick a common `CoroutineScope` for all their peripherals _and_ likely also to **launch coroutines for collecting `observe`s**. This can be problematic if a consumer wants to cancel collection of observations (i.e. via `cancelChildren`). If they keep their `Peripheral` connection open, the `cancelChildren` call will cancel the `Peripheral`'s internal coroutine that is used to forward characteristic change events.

By changing `Peripheral` creation to take a `CoroutineScope` as a parameter (**with a default of `GlobalScope`**, or similar) makes the footgun less likely to be encountered (as you have to **opt-in** to picking a `CoroutineScope` that may accidentally cancel the `Peripheral`'s internal coroutine).

PR #733, opened 2024-08-08
Unify scan exceptions

### LINKED ISSUE #103: Advertisement flow should throw an error if location permission is missing
So I started trying out the library following the readme and noticed that after collecting the advertisements flow, a scanner was registered, as per the log: `BluetoothLeScanner: onScannerRegistered() - status=0 scannerId=9 mScannerId=0`.
However I wasn't getting any scan record, or advertisement as the library calls it, and wondered whether I was making some mistake.

Turns out I hadn't allowed location permission, but at the same time I wondered whether the library should check for it instead of silently doing nothing. Maybe throw an error if location permission is missing? As it stands, if bluetooth is disabled when collecting the advertisements flow an error is thrown, I'd expect to do the same if location permissions are missing.

What are your thoughts?

### LINKED ISSUE #239: Scanner requires background location permission on Android 10 and 11
When running on a device using Android 10 or 11, no advertisements are detected by `Scanner` unless the user has set their location permission to "Allow all the time" (i.e. the ACCESS_BACKGROUND_LOCATION permission). This is inconvenient for apps which only use scanning while in the foreground, particularly when [the user is required to manually change this permission in settings on Android 11](https://developer.android.com/training/location/permissions#request-background-location). 

Is it possible to configure `Scanner` to require only foreground location permissions on Android 10 and 11? If not, it might be helpful to include this permission requirement in the README. (It was not immediately obvious that detection of no advertisements was a permission issue, as documented in #103, especially since this issue only occurs on certain Android versions.)

twyatt:
> [!WARNING]
> This PR contains major breaking changes (w/ exception hierarchy changes), and puts the exception hierarchy in an inconsistent state (peripheral I/O exceptions need to be updated to match exception handling in scanning).

Bluetooth machinery is now accessed via standard library provided properties (i.e. using `window.navigator` — no more `js("..")` calls needed) with `Navigator.bluetooth` provided as an extension function (that behaves similarly to the built-in properties — being `undefined` when not available).

Previously, it was inconsistent across platforms when/what exceptions would be thrown when scanning. Even more problematic was: on some platforms, the platform specific exception (e.g. `DOMException` on JavaScript) was allowed to propagate — making catching these exceptions from common code impossible. This PR standardizes the unhappy path so that the `Scanner` contract has the same exceptions thrown across all platforms (and can be used from common code — sans `requestDevice`, which is JavaScript only).

All scanning operations (and `requestDevice` on JavaScript) only throw the following exceptions:

- `IllegalStateException` — thrown if a _static_ state prevents scanning (e.g. bluetooth unavailable, scanning feature not supported or insufficient permission) — essentially, anything that could've/should've been checked prior to initiating the scan
- `UnmetRequirementException` (extends `IOException`) — thrown if a _transient_ state prevented scanning (e.g. bluetooth disabled) — essentially, anything that even if you check it ahead of time, may not be satisfied by the time the scan is initiated
- `InternalException` — if a condition was hit that was unexpected (this is a bug and exception message asks consumers to report when this exception is raised)

_System specific exceptions may be held as the `cause` of the above exceptions._

Consumers may now check for `Bluetooth.availability` then safely only catch `IOException` when performing a scan.

`Bluetooth.availability` would previously throw an exception if bluetooth wasn't available, this has been fixed and that flow should no longer throw. Instead, it should emit the appropriate `Unavailable` event.

Partially addresses #724, #721.
Related to #286 (added suspend `requestDevice` function that takes `CoroutineScope` as parameter).
Closes #103
Closes #239

Downstream SensorTag PR: https://github.com/JuulLabs/sensortag/pull/376


kable-core/src/androidMain/kotlin/Bluetooth.kt
```diff
@@ -15,75 +16,89 @@ import android.os.Build.VERSION.SDK_INT
...
 import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.emitAll
+import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.flowOf
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.onStart
 import android.bluetooth.BluetoothAdapter.ACTION_STATE_CHANGED as BLUETOOTH_STATE_CHANGED
 
-public actual enum class Reason {
+public actual enum class AvailabilityReason {
```
sdonn3, line 37: Maybe  in-availability reason? Seems like these are all reasons why it's inaccessible?

sdonn3 (approved):
Seems like reasonable changes, appreciate the extensive commenting as usual

kable-core/src/jsMain/kotlin/Bluetooth.kt
```diff
@@ -1,103 +1,36 @@
...
-    } else {
-        val isAvailable = bluetooth.getAvailability().await()
-        if (isAvailable) Available else Unavailable(reason = null)
+/**
+ * @return [Bluetooth] object or `null` if bluetooth is [unavailable](https://developer.mozilla.org/en-US/docs/Web/API/Bluetooth#browser_compatibility).
+ */
+internal fun bluetoothOrNull(): Bluetooth? {
+    val navigator = try {
+        window.navigator
+    } catch (e: ReferenceError) {
```
davidtaylor-juul, line 14: It is kinda sad to me that Kotlin doesn't differentiate the Js targets (e.g. browser vs node) and has these definitions in the standard library that just crash if you use them in the wrong environment. 😞

kable-core/src/jsMain/kotlin/Bluetooth.kt
```diff
@@ -1,103 +1,36 @@
...
+        // ReferenceError: window is not defined
+        throw IllegalStateException("Bluetooth unavailable", e)
     }
-    if (!this@toRequestDeviceOptions.optionalServices.isNullOrEmpty()) {
-        optionalServices = this@toRequestDeviceOptions.optionalServices
-            .map(Uuid::toBluetoothServiceUUID)
-            .toTypedArray()
+    val bluetooth = navigator.bluetooth
+    if (bluetooth === undefined) {
+        error("Bluetooth unavailable")
```
davidtaylor-juul, line 33: This seems duplicative - we have the same logic as above, and both `throw IllegalStateException("Bluetooth unavailable", e)` and `error("Bluetooth unavailable")` here. How about we just write what the function name suggests?
```kotlin
internal fun bluetoothOrThrow(): Bluetooth =
    bluetoothOrNull() ?: error("Bluetooth unavailable")
```

davidtaylor-juul (approved):
LGTM, one small issue to address but not a blocker.

kable-core/src/jsMain/kotlin/Bluetooth.kt
davidtaylor-juul: Oh, just realized the `throw` is re-wrapping the exception. Derp. 🫠 Carry on...

kable-core/src/jsMain/kotlin/Bluetooth.kt
twyatt: Granted it'd crash on node if you tried to access `window.navigator` in pure JS. 🤷

kable-core/src/commonMain/kotlin/Filter.kt
```diff
@@ -119,6 +119,16 @@ public sealed class Filter {
...
+
+        override fun toString(): String = buildString {
+            append("ManufacturerData(id=")
+            append(id)
+            append(", data=")
+            append(data.toHexString())
+            append(", dataMask=")
+            append(dataMask?.toHexString())
+            append(')')
+        }
```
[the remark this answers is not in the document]
twyatt: I think at one point, I conditionally included a property? That's my story, and I'm sticking to it.

No good reason for it now, I'll change it to string interpolation. 👍

kable-core/src/androidMain/kotlin/Bluetooth.kt
twyatt: Naming is hard. 😝 

I'll revert this actually, since "availability" is going away entirely soon (#737). So, this `enum` won't even be around for much longer anyways.

kable-core/src/jsMain/kotlin/Bluetooth.kt
davidtaylor-juul: Seems sensible enough. I guess it is too late for JB to split the targets so we could build for `js().browser()` and `js().node()`, but it sucks that they provide an API for `window.navigator` that just crashes. 🫠

PR #738, opened 2024-08-16
Add `Bluetooth.isSupported()` function

twyatt:
> [!WARNING]
> This PR is a part of a larger refactor.

Provides a means for determining if Bluetooth Low Energy is supported on the system/platform.


kable-core/src/appleMain/kotlin/bluetooth/IsSupported.kt
```diff
@@ -0,0 +1,49 @@
...
+import platform.CoreBluetooth.CBCentralManagerOptionShowPowerAlertKey
+import platform.CoreBluetooth.CBManagerState
+import platform.CoreBluetooth.CBManagerStateResetting
+import platform.CoreBluetooth.CBManagerStateUnknown
+import platform.CoreBluetooth.CBManagerStateUnsupported
+import platform.darwin.NSObject
+
+// Prevent triggering permissions dialog.
+// https://chrismaddern.com/determine-whether-bluetooth-is-enabled-on-ios-passively/
+private val options = mapOf(CBCentralManagerOptionShowPowerAlertKey to false) as Map<Any?, *>?
```
Phoenix7351, line 20: This actually works? Has this been tested? The article is 10 years old.

kable-core/src/commonMain/kotlin/Bluetooth.kt
```diff
@@ -34,6 +35,15 @@ public object Bluetooth {
         public data class Unavailable(val reason: Reason?) : Availability()
     }
 
+    /**
+     * Checks if Bluetooth Low Energy is supported on the system. Being supported (a return of
+     * `true`) does not necessarily mean that bluetooth operations will work. The radio could be off
+     * of permissions may be denied.
```
Phoenix7351, line 41: or*

kable-core/src/appleMain/kotlin/bluetooth/IsSupported.kt
twyatt: Yup. I tested it, and it works. 🎉

Phoenix7351 (approved):

kable-core/src/appleMain/kotlin/bluetooth/IsSupported.kt
twyatt: Ya, right?! It was a lucky find, and even luckier that it actually worked. 😆

kable-core/src/appleMain/kotlin/bluetooth/IsSupported.kt
```diff
@@ -0,0 +1,49 @@
...
+                // Silently ignore.
+            }
+        }
+    }.also { delegateRef = it }
+    CBCentralManager(delegate, null, options).also { managerRef = it }
+    awaitClose {
+        managerRef = null
+        delegateRef = null
+    }
+}.first { it.isDetermined }
```
davidtaylor-juul, line 46: I didn't dig into it, but curious if `suspendCoroutine` would be a better fit instead of `callbackFlow` here, did you try that on for size at all?

davidtaylor-juul (approved):
That iOS hack is a nice find! 💯

kable-core/src/appleMain/kotlin/bluetooth/IsSupported.kt
twyatt: I actually started with that, unfortunately the delegate would trigger a 2nd time (before things) had been spun down, and `suspendCoroutine` crashes the app because you can only resume the continuation once. Within `callbackFlow` we can ignore the `onFailure` from the `trySend`.

Issue #737, opened 2024-08-14
Remove `Bluetooth.availability`


twyatt:
The `Bluetooth.availability` is problematic because it provides `Available` or `Unavailable` but it is inconsistent across platforms.

This comes down to the major differences in how Android vs. Core Bluetooth (e.g. iOS) report their bluetooth radio state and bluetooth permissions.

### Core Bluetooth

Core Bluetooth combines BLE supported status, radio state and permissions via [`CBManagerState`](https://developer.apple.com/documentation/corebluetooth/cbmanagerstate):

- `poweredOff` _(radio state)_
- `poweredOn` _(radio state)_
- `resetting`
- `unauthorized` _(permission)_
- `unknown`
- `unsupported` _(supported status)_

### Android

Android separates BLE supported status, radio state and permissions.

BLE supported status can be checked via the existence (non-`null`) value for the `BluetoothAdapter`. Radio state is provided via `BluetoothAdapter` states:

- `STATE_OFF`
- `STATE_ON`
- `STATE_TURNING_OFF`
- `STATE_TURNING_ON`

Permissions are determined via the Android permissions API.

### Differences

On Android, there is no way of getting real-time permission status (it is checked on-demand) while having permission on iOS comes in via the `CBCentralManager` delegate (a non-`unsupported`) state.

Requesting permission on Android is explicit, while permission request is presented to the user on construction of `CBCentralManager`.

### Deprecation of `Bluetooth.availability`

Because of the major platform differences, there is no way to provide a consistent _flow_ of when BLE is "ready to use" (i.e. available). As such, detecting the availability of BLE is better left to the consumer.

A possible approach will be provided in the SensorTag sample app.

twyatt:
**UPDATE:** Even providing `isSupported` has proven challenging (due to Apple limitations with being able to query and **always** showing the user an intrusive permission dialog). Providing `isSupported` in Kable is still being considered.

curioustechizen:
I understand why this was closed. I'm wondering if it makes sense for Kable to somehow expose the underlying `CBCentralManager` though. Because unlike Android where BluetoothManager is a singleton, on iOS you can have multiple `CBCentralManager` instances.

[...]

Issue #662, opened 2024-03-28
Services should be re-discovered once onServiceChanged is called


JonatanPlesko:
In API 31, [onServiceChanged](https://developer.android.com/reference/android/bluetooth/BluetoothGattCallback#onServiceChanged(android.bluetooth.BluetoothGatt)) was added to BluetoothGattCallback, which Kable currently does not override and handle.

When this callback is called, Kable needs to rediscover services and **turn on notifications/indications** again. I am currently experiencing an issue where none of the previously set observers work as Android silently clears up all of the existing notification/indication registrations when this occurs - see [L1141 in bta_gattc_act.cc](https://android-review.googlesource.com/c/platform/system/bt/+/1433963).

I'm not sure if rediscovering services and re-enabling notifications/indications is sufficient in this scenario for everything to work properly, but I can confirm that manually calling `BluetoothGatt#setCharacteristicNotification` fixes _my_ issue.

As for reproducing the issue, it should occur when the server changes one (or more) of its services while you're already connected to it and subscribed to some characteristics.

twyatt:
Thanks for reporting this! I'll try to find time soon to work on it.

The device you're experiencing this issue with, and testing with, does it happen to be a publicly available device? As I don't believe any of the devices I currently have perform a peripheral since service change that would trigger this.

If there isn't a publicly available device to test against, I'll probably end up just calling `setNotification` as you suggested.

JonatanPlesko:
Unfortunately, I am not using a publicly available device. It's probably possible to mock such behaviour on Android, but I haven't had the time to try.
BTW, I think calling BluetoothGatt#discoverServices also needs to happen; it's not relevant in my scenario since all of the services and characteristics I was subscribed to are still there, but I think that someone else could run into issues there.

PR #749, opened 2024-09-12
Leverage structured concurrency for connection handling

### LINKED ISSUE #144: Make disconnect timeout configurable instead of requiring use of `withTimeout`
Per [documentation](https://github.com/JuulLabs/kable#connectivity), when `disconnect` is called, it should be wrapped with `withTimeout`. This makes the API error-prone, as forgetting to do so can create very rare and hard to track down bugs.

It would be better to configure the disconnect timeout via the `peripheral` builder lambda (with a sane default).

### LINKED ISSUE #168: Handle out-of-band disconnects on JavaScript
Out-of-band disconnects are not properly handled on JavaScript. If a disconnect occurs in the middle of an I/O operation, the I/O operation may stall indefinitely. Instead, if a disconnect occurs during I/O (similar to Android platform) and exception should be thrown.

### LINKED ISSUE #177: Connection exceptions are not consistent
When attempting to perform an I/O operation, the following exceptions may be thrown to indicate that a connection is not _yet_ available (or has recently become unavailable) to complete the request:

- `NotReadyException`
- `ConnectionLostException`

Unfortunately when they are thrown is not entirely clear. Depending on when the connection drops during an I/O operation, either of the two above exceptions may be thrown.

It would be preferred that it was predictable which exception was thrown, the thinking is that `NotReadyException` should correspond to being unable to fulfill an I/O request because a connection has not yet been established (e.g. while still connecting). Whereas `ConnectionLostException` should be thrown if we previously had a connection but it was recently lost (while trying to perform the I/O operation).

exceptions

It would also likely be helpful if both these exceptions extended from a common exception class so that library consumers could catch a general "connection not established" exception, if desired.

### LINKED ISSUE #286: Move `CoroutineScope` as function parameter for `Peripheral` creation

### LINKED ISSUE #378: Peripheral blocking scope from finishing

### LINKED ISSUE #497: OutOfOrderGattCallbackException: Unexpected response type OnCharacteristicWrite received
I work on Android. Have been using this lib for more than one year. And now I got the case when I catch `OutOfOrderGattCallbackException`. The full text is `OutOfOrderGattCallbackException: Unexpected response type OnCharacteristicWrite received`. The scenario is pretty simple: there are two characteristics. One I use to write byte arrays on a device, the second one is read the status of this "writing". And in one moment I get this exception. I found it is source code but its cause is not clear for me. Why do I get it?

### LINKED ISSUE #567: Worker never ends if I use kable in it
If I just create a `periferal` instance inside worker then this worker never ends. 

I am enqueueing a unique work. For the first time it runs fine (I see in logs `Finished worker`), when I enqueue again the work it does not start. I assume this is because the previous work was not finished.

```kotlin

@HiltWorker
class DfuWorker @AssistedInject constructor(@Assisted private val context: Context,
                                            @Assisted workerParameters: WorkerParameters
) : CoroutineWorker(context, workerParameters) {
    override suspend fun doWork(): Result = withContext(Dispatchers.IO) {
        Timber.tag(TAG).d("Started")

        // If I comment this line all works as expected
        val peripheral = peripheral(Biosensors.bluetoothDevice(deviceId))

        Timber.tag(TAG).d("Finished worker")
        Result.success()
    }

    companion object {
        private const val TAG = "DfuWorker"

        fun WorkManager.runDfuWorker(deviceId: String, deviceType: DeviceType){
            val data = Data
                .Builder()
                .putString(DATA_DEVICE_ID, deviceId)
                .putInt(DATA_DEVICE_TYPE, deviceType.code)
                .build()

            val workRequest = OneTimeWorkRequest
                .Builder(DfuWorker::class.java)
                .addTag("FIRMWARE_UPDATE")
                .setInputData(data)
                .build()

            enqueueUniqueWork(
                "update_firmware",
                ExistingWorkPolicy.KEEP,
                workRequest
            )

            Timber.tag(TAG).d("Requested enqueue worker for dfu.")
        }
    }
}
```

### LINKED ISSUE #622: IllegalStateException when using GlobalScope with CoroutineScope.peripheral in 0.28.0
Trying to get the peripheral using GlobalScope as a receiver for the CoroutineScope.peripheral extension function in the latest version (0.28.0) throws the following exception:
``` 
java.lang.IllegalStateException: Current context doesn't contain Job in it: com.myproject.di.modules.DomainModule$createConnectionScope$$inlined$CoroutineExceptionHandler$1@3595913
    at kotlinx.coroutines.JobKt__JobKt.getJob(Job.kt:645)
    at kotlinx.coroutines.JobKt.getJob(Unknown Source:1)
    at com.juul.kable.BluetoothDeviceAndroidPeripheral.<init>(BluetoothDeviceAndroidPeripheral.kt:74)
    at com.juul.kable.PeripheralKt.peripheral(Peripheral.kt:30)
    at com.juul.kable.PeripheralKt.peripheral(Peripheral.kt:47)
    at com.juul.kable.PeripheralKt.peripheral$default(Peripheral.kt:42)
```

For reference, the `createConnectionScope` function creates the scope which I use for all BLE operations and looks like this:
```
@OptIn(DelicateCoroutinesApi::class)
private fun createConnectionScope(): CoroutineScope = GlobalScope + CoroutineExceptionHandler { _, throwable ->
    Timber.e(throwable)
}
```

Using `CoroutineScope(SupervisorJob() + Dispatchers.Default)` instead of `GlobalScope` works. The issue is not present in 0.27.0.

twyatt:
## Summary

Major changes include:
- Updated JavaScript to use `SharedRepeatableAction`
    - Now all platforms utilize `SharedRepeatableAction` (for triggering connect/disconnect)
- `Peripheral`s are no longer children of a `CoroutineScope`
    - Their lifecycle now must be explicitly ended via `cancel`
- `Peripheral`s are themselves `CoroutineScope`s
    - Tasks may be `launch`ed from `Peripheral`s when the lifecycle of the task is intended to match the `Peripheral`s lifecycle
- `Peripheral.connect` now returns a `CoroutineScope` 
    - Tasks that are intended to match the current connection's lifecycle may be run from this `CoroutineScope`
- Simplified exception hierarchy
    - On the path to extending either `IOException` or `IllegalStateException` for most exceptions (#724)

## Scoped connections

Connection and I/O management, as well as disconnect process was moved to `Connection` class (for every platform). The `Connection` class manages any context switching necessary for I/O and ensuring proper teardown on disconnect. This simplifies connection lifecycle management by leveraging coroutine's structured concurrency for connection lifecycle (proper teardown on either disconnect _or_ peripheral scope completing).

We no longer have to do a `Deferred` [dance (state management) when performing I/O](https://github.com/JuulLabs/kable/blob/355cb41f65ee4d991c777c9138019ac46ee7d424/kable-core/src/androidMain/kotlin/Connection.kt#L55-L90) (when cancelled by caller). When I/O goes through `Connection.execute` (which is scoped to the `Connection`'s `CoroutineScope`), if caller cancels I/O request, then we can continue (asynchronously) waiting for the response (rather than trying to [drop the response on the next I/O operation](https://github.com/JuulLabs/kable/blob/355cb41f65ee4d991c777c9138019ac46ee7d424/kable-core/src/androidMain/kotlin/Connection.kt#L55-L68)).

The following chart shows the approximate relationship between the various `CoroutineScope`s within a `Peripheral`:

```mermaid
flowchart TD
    
    peripheralScope:::scope --> connectionScope:::sscope
    connectionScope --> taskScope
    connectionScope --> action:::sscope

    subgraph Peripheral
    peripheralScope:::scope
    
    subgraph Connection
    connectionScope:::sscope
    taskScope:::sscope
    end
    
    disconnect -.-x connectionScope
    action -.-o Connection
    connect -..-> action
    
        
    end

    peripheralScope("peripheralScope (SupervisorJob)")
    action("action (Deferred)")
    taskScope("taskScope (SupervisorJob)")
    connectionScope("connectionScope (Job)")

    Connection:::klazz
    
    classDef scope fill:#33f
    classDef sscope fill:#a3c
    classDef klazz fill:#333

    linkStyle 0 stroke-width:2px,fill:none,stroke:white;
    linkStyle 1 stroke-width:2px,fill:none,stroke:white;
    linkStyle 2 stroke-width:2px,fill:none,stroke:white;
    linkStyle 3 stroke-width:2px,fill:none,stroke:red;
    linkStyle 4 stroke-width:2px,fill:none,stroke:green;
    linkStyle 5 stroke-width:2px,fill:none,stroke:white;
```

When `Peripheral.connect` is called, the `CoroutineScope` tree (shown in pink) is spun up (via `SharedRepeatableAction`). The `SharedRepeatableAction` creates a `CoroutineScope` (which becomes the `Connection`'s `connectionScope`) and a child `Deferred` (which is executed as an `action` within the `Peripheral.connect` function). In the `action`, the `Connection` is created (using the `connectionScope`) and immediately a child `CoroutineScope` (which is shown as `taskScope`) is created.

In summary:
- `connectionScope` is used for asynchronously executing I/O operations
- `taskScope` is provided to library consumers to launch "tasks" that should match the lifecycle of the connection
- `action` runs the connection process (e.g. connect, discover services, etc)

Some characteristics of this `CoroutineScope` tree that are pertinent:
- A failure of `connectionScope` will shutdown the connection `CoroutineScope` tree (pink elements)
- Tasks (coroutines launched from `taskScope`) are allowed to fail w/o shutting anything else down
- A failure in `action` (during connect process) will shutdown the connection `CoroutineScope` tree (pink elements)

When `Connection` is created, it registers a `Job` completion listener on the `connectionScope` which properly disposes of the underlying peripheral (e.g. on Android, the `BluetoothGatt` is `close`d) — this ensures that after the connection `CoroutineScope` tree has completed that the underlying BLE objects are always properly disposed.

---

Partially addresses #724
Partially addresses #721
Closes #622
Closes #567
Closes #378
Closes #177
Closes #144
Closes #497
Closes #286
Closes #168


kable-core/src/androidMain/kotlin/BluetoothDeviceAndroidPeripheral.kt
```diff
@@ -68,198 +50,131 @@ internal class BluetoothDeviceAndroidPeripheral(
...
+            suspendUntil<State.Connecting.Services>()
             discoverServices()
-            onServicesDiscovered(ServicesDiscoveredPeripheral(this@BluetoothDeviceAndroidPeripheral))
-
-            _state.value = State.Connecting.Observes
-            logger.verbose { message = "Configuring characteristic observations" }
-            observers.onConnected()
+            configureCharacteristicObservations()
         } catch (e: Exception) {
-            closeConnection()
```
twyatt, line 143: We no longer have to explicitly close the connection, because it will be closed implicitly when the connection coroutine scope tree completes. Since the connection process runs within a `Deferred` that is a child of the connection coroutine scope tree, any failure during connect will implicit close down the `Connection`.

kable-core/src/androidMain/kotlin/BluetoothDeviceAndroidPeripheral.kt
```diff
@@ -68,198 +50,131 @@ internal class BluetoothDeviceAndroidPeripheral(
...
-            logger.error(failure) { message = "Failed to connect" }
+            val failure = e.unwrapCancellationException()
+            logger.error(failure) { message = "Failed to establish connection" }
             throw failure
         }
 
         logger.info { message = "Connected" }
         _state.value = State.Connected
 
-        state.watchForConnectionLossIn(scope)
```
twyatt, line 152: This is performed in the `Connection` now.

kable-core/src/androidMain/kotlin/gatt/Callback.kt
```diff
@@ -275,6 +273,11 @@ internal class Callback(
         if (status == GATT_SUCCESS) this.mtu.value = mtu
     }
 
+    override fun onServiceChanged(gatt: BluetoothGatt) {
+        logger.debug { message = "onServiceChanged" }
+        // todo
```
Phoenix7351, line 278: Intentionally vague?

kable-core/src/appleMain/kotlin/CentralManager.kt
```diff
@@ -45,16 +45,16 @@ public class CentralManager internal constructor(
         }
     }
 
-    private val dispatcher = QueueDispatcher(DISPATCH_QUEUE_LABEL)
+    internal val dispatcher = QueueDispatcher(DISPATCH_QUEUE_LABEL)
     internal val delegate = CentralManagerDelegate()
-    private val cbCentralManager = CBCentralManager(delegate, dispatcher.dispatchQueue, options)
+    private val centralManager = CBCentralManager(delegate, dispatcher.dispatchQueue, options)
```
Phoenix7351, line 50: We sure we want this rename? The class here--itself--is named CentralManager. It was actually helpful to see the distinction that `cbCentralManager` was a reference to the Apple API.

kable-core/src/appleMain/kotlin/CentralManagerDelegate.kt
Phoenix7351, line 1: Thought there were going to be changes to didUpdateState here?

kable-core/src/appleMain/kotlin/bluetooth/CheckBluetoothIsOn.kt
```diff
@@ -0,0 +1,37 @@
...
+import platform.CoreBluetooth.CBManagerStatePoweredOn
+import platform.CoreBluetooth.CBManagerStateResetting
+import platform.CoreBluetooth.CBManagerStateUnauthorized
+import platform.CoreBluetooth.CBManagerStateUnknown
+import platform.CoreBluetooth.CBManagerStateUnsupported
+
+/**
+ * @throws UnmetRequirementException If [CentralManager] state is not [CBManagerStatePoweredOn].
+ */
+internal fun CentralManager.checkBluetoothIsOn() {
```
Phoenix7351, line 18: Not sure if this file would be the right place, but it would be really useful to have a flow of the CBManagerState itself. Is that factored into your plan somewhere?

twyatt: I've decided that Kable should not provide bluetooth state for consumers — it is too difficult to provide a unified API.

I'm trying to drive Kable in a direction where the entire API can be used from `common` source sets (write once, use everywhere paradigm).

In this particular instance: `CBManager*` states encompass:
- is BLE supported or not
- is BLE powered on or not
- does user have **permission** to use BLE

Whereas Android has all three of the above bullets available independently (and there is no live-updating permission info available on Android -- it is on-demand).

If `CBManager*` states didn't include the permission aspect, then it is would trivial to provide a unified API (of a flow of "is BLE powered on or not"), but since Apple doesn't separate those states, means that I just can't provide such an API.

So, we're left with only being able to provide an on-demand approach. Which sadly, I may even need to abandon this; as I've found that my findings of being able to suppress the dialog when using BLE isn't actually possible on Apple (I was previously mistaken on my findings: I am able to suppress for BLE on/off dialog, but permissions always comes up).

Making this API super intrusive...ultimately, it is hot garbage that Core Bluetooth doesn't let you query any information about BLE availability without showing the user a dialog. 😢 

I haven't completely abandoned the idea, but a unified API for these things may be impossible; really just trying to find the best compromise (and still have a reasonably friendly API).

kable-core/src/appleMain/kotlin/CentralManagerDelegate.kt
twyatt: Not sure I follow. Can you elaborate?

kable-core/src/appleMain/kotlin/CentralManager.kt
twyatt: Fair point. I generally like to be terse in my variable naming, but I see your point: that I've lost useful/valuable information in this name change. I'll revert. 👍

kable-core/src/androidMain/kotlin/gatt/Callback.kt
twyatt: This is just a stub for future work.
But good call-out, I'll update it to link https://github.com/JuulLabs/kable/issues/662.

kable-core/src/appleMain/kotlin/CentralManagerDelegate.kt
Phoenix7351: Was referring to the flow that tells the consumer the BLE state. Doesn't sound like it's happening.

Phoenix7351 (approved):

kable-core/src/appleMain/kotlin/CentralManagerDelegate.kt
twyatt: Since forwarding state updates is longer living than each connection, those events are forwarded from the central singleton [here](https://github.com/JuulLabs/kable/blob/dc0d94f4127d01f8fb181468e3ed279bf76c46a9/kable-core/src/appleMain/kotlin/CBPeripheralCoreBluetoothPeripheral.kt#L66-L72), with the state being available [here](https://github.com/JuulLabs/kable/blob/dc0d94f4127d01f8fb181468e3ed279bf76c46a9/kable-core/src/appleMain/kotlin/CBPeripheralCoreBluetoothPeripheral.kt#L63).

twyatt: Perhaps there was confusion re: removal of the `connectionState` flow?
There was previously `connectionEvents` _and_ `connectionState`, I decided that having a dedicated flow (`connectionState`) for the connection events was unnecessary, and now they go through `connectionEvents` (along with the other connection events).

kable-core/src/appleMain/kotlin/bluetooth/CheckBluetoothIsOn.kt
twyatt: Additional details can also be found here: https://github.com/JuulLabs/kable/issues/737

kable-core/src/androidMain/kotlin/Connection.kt
```diff
@@ -2,125 +2,271 @@ package com.juul.kable
...
+    )
+
+    val taskScope = connectionScope.childSupervisor("$name/Tasks")
+
+    private val logger =
+        Logger(logging, tag = "Kable/Connection", identifier = gatt.device.toString())
+
+    init {
+        // todo: Move this `require` to the PeripheralBuilder.
+        require(disconnectTimeout > ZERO) { "Disconnect timeout must be >0, was $disconnectTimeout" }
```
davidtaylor-juul, line 87: Is this TODO dependent on some other work or can this be moved to `PeripheralBuilder` as part of this PR?

kable-core/src/androidMain/kotlin/Connection.kt
```diff
@@ -2,125 +2,271 @@ package com.juul.kable
...
-            when (val unwrapped = e.unwrapCancellationCause()) {
-                is ConnectionLostException -> throw ConnectionLostException(cause = unwrapped)
-                else -> throw unwrapped
+            try {
+                connectionScope.async {
+                    callback.onResponse.receive()
+                }.await()
+            } catch (e: CancellationException) {
+                coroutineContext.ensureActive()
+                throw e.unwrapCancellationException()
```
davidtaylor-juul, line 182: This pattern (repeated a few times but I'll only comment on it the once here) is really confusing. If we catch a `CancellationException` and then call `ensureActive`, it seems like that is guaranteed to re-throw the `CancellationException`, and means we would never reach the following code where we re-throw the `e.unwrapCancellationException()`.

I assume I am misunderstanding something here but this code reads like it could be removed and the behavior would not change. 😕

davidtaylor-juul (approved):
This is pretty awesome, much easier to reason about the lifecycle of the various moving parts now. 🚀

kable-core/src/androidMain/kotlin/Connection.kt
twyatt: There are two possible sources for cancellation:
1. We're running in `suspend execute(..)` function (line 149)
2. Suspended `Channel.receive()` could cancel/throw

Essentially, cancellation could occur on either the consumer side _or_ the producer side.

The beauty of `ensureActive` is it says we should propagate a `CancellationException` iff calling side cancelled. i.e. If the coroutine context we're executing in was cancelled then let the cancellation occur.

Whereas, if cancellation occurred on the producer side (from the `Channel.receive()` then the consumer side coroutine context would still be active (and `ensureActive` would not throw).

So, we'll only unwrap the `CancellationException` if the cancellation occurred on the producer side, but we'll honor cancellation that a library consumer initiated.

davidtaylor-juul: Oooh, I see. Makes perfect sense in the broader context, thanks for explaining!

kable-core/src/androidMain/kotlin/Connection.kt
twyatt: I forget why I didn't do it in this PR. I'll try again. If I can't move it quite yet, then I'll merge w/ it as-is (with the `todo`) and try to get to it soon.

twyatt: Looked into it, I could move this, but it'd be duplicated for each platform (which is expected), but I realized that I plan to do some refactoring of `PeripheralBuilder`, so that `Config` objects are created, to make passing of those values around easier. I figured I'd just wait for that small refactor to pull this in. 🤷

PR #752, opened 2024-09-18
Make `isSupported` always `true` on Apple

twyatt:
I had previously found [Determine whether bluetooth is enabled on iOS (passively)](https://chrismaddern.com/determine-whether-bluetooth-is-enabled-on-ios-passively/), and even tested it!

Although, I mistakenly thought `CBCentralManagerOptionShowPowerAlertKey` had suppressed _all_ dialogs, only to later realize I had already granted permission to the app, so it was only suppressing the "turn BLE on" dialog, as mentioned in [What causes this iOS permission prompt for "use Bluetooth for new connections"?](https://stackoverflow.com/q/58600502).

Ultimately, it seems (as far as I can tell) there is no way to passively check if BLE is supported on Core Bluetooth. As a result, we have to blindly assume it is available for this API to have uniform behavior on all platforms.

I'm strongly considering removing `isSupported` from `core` Kable (perhaps providing it as a separate module instead). 🤷 

Alternatively/instead, could go with #753.


sdonn3 (approved):

davidtaylor-juul (approved):
I think this is probably fine given that any current hardware capable of running the supported targets all have bluetooth.

PR #753, opened 2024-09-18
Mark `Bluetooth.isSupported` as experimental and add warning in KDoc

twyatt:
Alternate approach to #752. Rather than fake the BLE support on Apple, provide a KDoc warning.


kable-core/src/commonMain/kotlin/Bluetooth.kt
```diff
@@ -40,8 +40,12 @@ public object Bluetooth {
      * `true`) does not necessarily mean that bluetooth operations will work. The radio could be off
      * or permissions may be denied.
      *
+     * WARNING: On Apple, calling this function will show a Bluetooth permission dialog (if
+     * authorization has not yet been granted).
+     *
      * This function is idempotent.
      */
+    @ExperimentalApi // Due to the inability to query Bluetooth support w/o showing a dialog on Apple, this function may be removed.
     public suspend fun isSupported(): Boolean = isBluetoothSupported()
```
Phoenix7351, line 49: Does this get called anywhere in kable? We don't want to auto prompt just based on pulling in Kable.

davidtaylor-juul (approved):

kable-core/src/commonMain/kotlin/Bluetooth.kt
twyatt: It does not. It is being used by my work-in-progress SensorTag sample refactor.

Phoenix7351 (approved):

twyatt:
Closed in favor of #752.

PR #809, opened 2024-12-03
Adds maximumWriteValueLength to Peripheral

### LINKED ISSUE #729: Feature request: Add maximumWriteValueLengthForType to iOS Peripheral
Hi, 
it would be great if you could add [maximumWriteValueLengthForType](https://developer.apple.com/documentation/corebluetooth/cbperipheral/maximumwritevaluelength(for:)) to the CoreBluetoothPeripheral interface.

Sample: 

```
    override fun maximumWriteValueLengthForType(writeType: WriteType): Int {
        val type = when (writeType) {
            WithResponse -> CBCharacteristicWriteWithResponse
            WithoutResponse -> CBCharacteristicWriteWithoutResponse
        }
        return cbPeripheral.maximumWriteValueLengthForType(type).toInt()
    }
```

ErikHellman:
- Calls maximumWriteValueLength on iOS
- Returns latest know value from onMtuChanged() - 3 on Android, or the default (23 - 3) if the callback haven't been called
- Return default (23 -3) on Web

Closes #729

### DIFF
--- a/kable-core/src/commonMain/kotlin/Peripheral.kt
+++ b/kable-core/src/commonMain/kotlin/Peripheral.kt
@@ -154,6 +154,15 @@ public interface Peripheral : CoroutineScope {
     public val services: StateFlow<List<DiscoveredService>?>
 
     /**
+     * Return the current ATT MTU size, minus the size of the ATT headers (3 bytes).
+     *
+     * On Android, this will be the default (23 - 3) unless you called [requestMtu] when connecting.
+     * For iOS, this is automatically negotiated, and can also vary depending on the writeType.
+     * On JavaScript, this will return the default (23 - 3) every time as there is no ATT MTU property available.
+     */
+    public suspend fun maximumWriteValueLengthForType(writeType: WriteType): Int
+
+    /**
      * On JavaScript, requires Chrome 79+ with the
      * `chrome://flags/#enable-experimental-web-platform-features` flag enabled.
      *
--- a/kable-core/api/android/kable-core.api
+++ b/kable-core/api/android/kable-core.api
@@ -319,6 +319,7 @@ public abstract interface class com/juul/kable/Peripheral : kotlinx/coroutines/C
 	public abstract fun getName ()Ljava/lang/String;
 	public abstract fun getServices ()Lkotlinx/coroutines/flow/StateFlow;
 	public abstract fun getState ()Lkotlinx/coroutines/flow/StateFlow;
+	public abstract fun maximumWriteValueLengthForType (Lcom/juul/kable/WriteType;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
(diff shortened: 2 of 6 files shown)


kable-core/src/commonMain/kotlin/Peripheral.kt
```diff
@@ -153,6 +153,15 @@ public interface Peripheral : CoroutineScope {
...
     public val services: StateFlow<List<DiscoveredService>?>
 
+    /**
+     * Return the current ATT MTU size, minus the size of the ATT headers (3 bytes).
+     *
+     * On Android, this will be the default (23 - 3) unless you called [requestMtu] when connecting.
+     * For iOS, this is automatically negotiated, and can also vary depending on the writeType.
+     * On JavaScript, this will return the default (23 - 3) every time as there is no ATT MTU property available.
+     */
+    public suspend fun maximumWriteValueLengthForType(writeType: WriteType): Int
```
twyatt, line 163: I'm torn on the best approach for getting MTU:
1. via `suspend` function (as you have here)
2. via a flow (as [currently implemented for Android](https://github.com/JuulLabs/kable/blob/61bfcb87b610a1e6f96fd03a1865f53bf6a2ec43/kable-core/src/androidMain/kotlin/AndroidPeripheral.kt#L157-L162))

The current Android implementation provides the raw MTU, but I'm not aware of any use-cases that need the raw value, so I think it makes more sense to only provide the value as you have it here (maximum write size).

The Android flow implementation also should be moved to common code (on the `Peripheral` as you have it here).

Do you have any experience in needing to know the max write size that favors one approach over the other?

twyatt:
Note: `:kable-core:androidApiCheck` Gradle task failure can be fixed by running `./gradlew apiDump` then pushing the updated API dump files.

kable-core/src/commonMain/kotlin/Peripheral.kt
ErikHellman: The reason I didn't use the Flow is that the MTU should never be changed. According to the BLE spec, you should negotiate (request) the MTU immediately when connected. This is also why iOS only have this method instead of a callback.

twyatt: Makes sense. 👍 

Outside the scope of this PR, but I'll plan on dropping `requestMtu` function from `AndroidPeripheral` and instead have MTU configured via the peripheral builder.

twyatt (approved):

ErikHellman:
It would be awesome if we could get this in a near release as I need this for my current project. @cedrickcooke does this solution look ok for you?

cedrickcooke (approved):
Sorry for the delay on this. Yup, looks good

twyatt:
I'll release this w/ `0.35.0` in the coming days.

twyatt:
@ErikHellman released with [`0.35.0`](https://github.com/JuulLabs/kable/releases/tag/0.35.0).

> [!NOTE
> Compared to `0.34.0`, the internals include a major refactor. I believe the `0.35.0` RC releases have been out long enough to get word of any major bugs, but please don't hesitate to report any issues you come across.

PR #831, opened 2025-01-15
Fix error string for `GATT_INVALID_ATTRIBUTE_LENGTH`

angusholder:
Looks like a typo or copy-paste mistake? "Invalid attribute" is a bit different to "Invalid attribute length" - made it sound like the attribute didn't exist, when actually I was just not sending a big enough payload for the write!


twyatt (approved):
Oh yikes. Definitely a mistake. Thanks for catching/fixing this!

davertay-j (approved):

PR #835, opened 2025-01-18
Support Web Bluetooth named UUIDs

twyatt:
Provides the following `Uuid` extension functions:

- `Uuid.service(name: String)`
- `Uuid.characteristic(name: String)`
- `Uuid.descriptor(name: String)`

Examples can be found in the updated [documentation](https://github.com/JuulLabs/kable/tree/twyatt/bluetooth-uuid-names?tab=readme-ov-file#uuids).

Supported names can be seen [here](https://github.com/JuulLabs/kable/blob/twyatt/bluetooth-uuid-names/kable-core/src/commonMain/kotlin/Uuid.kt).


twyatt:
Note-to-self: update documentation.

cedrickcooke (approved):

davertay-j (approved):

PR #839, opened 2025-01-19
Convert discovered services, characteristics, descriptors to interfaces

### LINKED ISSUE #167: Incorrect `Characteristic` equality for `DiscoveredCharacteristic`
On Android at least (have not checked for the other platforms) the equality check of `DiscoveredCharacteristic` may be incorrect if the list of descriptors is non-empty:

https://github.com/JuulLabs/kable/blob/5d6e00e01ac53ea5c76b69fc4b8b897dbb578c75/core/src/commonMain/kotlin/Characteristic.kt#L24-L28

This is because the equality of a `DiscoveredCharacteristic` will be checked against it's list of descriptors, and on Android the `BluetoothGattDescriptor` object does not provide an `equals` function. So, if we have two descriptor objects in memory that represent the same descriptor on a peripheral, then they will erroneously be seen as not equal.

In most cases this will not be an issue, but the `Observers` class relies on the correctness of `Characteristic` equality.

### LINKED ISSUE #802: Mockable Peripheral
It would really help us with our testing if we could make our own implementation of `Peripheral` to test against.

With `Peripheral` being a public interface, we are *so close* to being able to fully mock the device we're connecting to.  However, `DiscoveredService` is platform specific, which interferes with attempts to mock it.

I would like `DiscoveredService` and `DiscoveredCharacteristic` to be interfaces such that they can be mocked, if it is possible.

Thanks!  This library is pretty awesome for our use so far!

twyatt:
Closes #802
Closes #167


kable-core/src/androidMain/kotlin/Profile.kt
```diff
@@ -18,40 +17,97 @@ internal actual typealias PlatformCharacteristic = BluetoothGattCharacteristic
...
+        if (javaClass != other?.javaClass) return false
+        other as PlatformDiscoveredService
+        if (serviceUuid != other.serviceUuid) return false
+        if (instanceId != other.instanceId) return false
+        return true
+    }
+
+    override fun hashCode(): Int {
+        var result = serviceUuid.hashCode()
+        result = 31 * result + instanceId
```
Phoenix7351, line 39: Is this 31 just arbitrary?

Phoenix7351 (approved):

kable-core/src/androidMain/kotlin/Profile.kt
twyatt: I never thought of it much prior to this because I always have the IDE generate the `hashCode`/`equals` functions.

ChatGPT gives some reasons why `31` is used:
- it's a prime number ("helps achieve better distribution of hash codes")
- efficient multiplication ("In binary, `31` is represented as `11111`. Multiplying by `31` can be optimized by the compiler..")
- historical (used for `String.hashCode` in Java standard library)
- odd number advantage ("Since `31` is odd, multiplying by it avoids the risk of canceling out bits during hash code generation)

Phoenix7351: Interesting.

Phoenix7351: Since it's being re-used in other places for the same purpose. Might be worth making it a const and adding this as an explanation? 🤷‍♂️

twyatt: Although I agree that a constant would be nice for clarity, I'm reluctant to take that approach, as it means I'll have to update the generated code any time I update/generate it.

I'll also have to make sure any contributors also use the constant rather than just going w/ the generated code.

Sadly, I don't feel like the juice is worth the squeeze. 😢

QuantumRand (approved):

twyatt:
Note-to-self: integrate equalsverifier

PR #842, opened 2025-01-20
Allow for advertisement manufacturer data length to exceed filter data mask length

### LINKED ISSUE #814: Inconsistent behavior between native and filtered ManufacturerData
Using 0.35.0-rc

If I use one `Filter.ManufacturerData` and I pass a single element `ByteArray` for both `data` and `dataMask` and everything works fine (the actual manufacturer data is 9 bytes).

If I use multiple `Filter.ManufacturerData` and I pass a single element `ByteArray` for both `data` and `dataMask` I get a crash saying that:

> Data mask length (1) must match data length (9)

because `Filter.ManufacturerData.matches(data: ByteArray?)` calls `requireDataAndMaskHaveSameLength` using the actual manufacturer data (as opposed to the `data` parameter).

twyatt:
Closes #814

As seen on the ["Web Bluetooth / Manufacturer Data Filter Sample"](https://googlechrome.github.io/samples/web-bluetooth/manufacturer-data-filter.html?companyIdentifier=0x000D&dataPrefix=03&mask=FF), the actual manufacturer data of a SensorTag (`030000`) is allowed to exceed the length of the specified _filter_ for **manufacturer data** and **mask** (`03` and `FF`, respectively, in the screenshot):

Screenshot 2025-01-19 at 11 26 30 PM


Phoenix7351 (approved):
Nice tests.

davertay-j (approved):

PR #855, opened 2025-02-06
Allow filtering by company ID only

### LINKED ISSUE #854: Filtering on companyIdentifier does not work
I am trying to filter devices by `companyIdentifier` this work for Android, but in JS I get an error.

requestPeripheral options:
```
private val options = Options {
    filters {
        match {
            manufacturerData = listOf(Filter.ManufacturerData(0x0A61, byteArrayOf()))
        }
    }
}
```

Error:
```
SystemLogEngine.kt:35 [Kable/requestDevice] TypeError: Failed to execute 'requestDevice' on 'Bluetooth': 'dataPrefix', if present, must be non-empty.
  options: Options(filters=[FilterPredicate(filters=[ManufacturerData(id=2657, data=, dataMask=null)])], optionalServices=[])
  processed: {"filters":[{"manufacturerData":[{"companyIdentifier":2657,"dataPrefix":{}}]}]}
```

This code in "normal" WebBLE api works:
 ```
       device = await navigator.bluetooth.requestDevice({
          filters: [{ manufacturerData: [{ companyIdentifier: 0x0A61 }] }],
        });
```

### PR DESCRIPTION
**Pull request by twyatt**
Closes #854

Updated scan filtering documentation, rendered markdown [here](https://github.com/JuulLabs/kable/tree/twyatt/companyId-only-filter?tab=readme-ov-file#scanning).

### DIFF
--- a/kable-core/src/commonMain/kotlin/Filter.kt
+++ b/kable-core/src/commonMain/kotlin/Filter.kt
@@ -103,7 +103,8 @@ public sealed class Filter {
          */
         public val id: Int,
 
-        public val data: ByteArray,
+        /** Must be non-`null` if [dataMask] is non-`null`. */
+        public val data: ByteArray? = null,
 
         /**
          * For any bit in the mask, set it to 1 if advertisement manufacturer data needs to match the corresponding bit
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: kable-core/src/commonMain/kotlin/Filter.kt [hunk 1]
```diff
@@ -103,7 +103,8 @@ public sealed class Filter {
          */
         public val id: Int,
 
-        public val data: ByteArray,
+        /** Must be non-`null` if [dataMask] is non-`null`. */
```
Phoenix7351, line 106: Good point. Gonna check to make sure Topaz's decode covers this.

Phoenix7351 (approved):

kable-core/src/commonMain/kotlin/Filter.kt
Phoenix7351: Wasn't covered. Will add.

Phoenix7351: Do we have this for serviceData too?

twyatt: Kable doesn't (yet) support service data filtering. 🙈 
When I do get around to implementing it, then yes, it looks like it would need similar treatment.

twyatt: Created #857 to track.

cedrickcooke (approved):

PR #867, opened 2025-02-21
Configure MTU via peripheral builder

### LINKED ISSUE #811: Drop `Android.requestMtu`
Drop `requestMtu` function from `AndroidPeripheral` and instead have MTU configured via the peripheral builder.

_Originally posted by @twyatt in https://github.com/JuulLabs/kable/pull/809#discussion_r1870305553_

twyatt:
Closes #811

---

> [!NOTE]
> **Medium Risk**
> Breaking API removal and new default MTU on every connect change connection timing and permission requirements; write chunk sizing behavior changes on newer Android versions due to the attribute-length cap.
> 
> **Overview**
> **Android MTU is now configured on the peripheral builder** instead of calling `requestMtu` during `onServicesDiscovered` or on `AndroidPeripheral`. New `PeripheralBuilder.mtu` defaults to **517** (`null` skips any MTU request); the library requests MTU **after GATT connect and before service discovery**.
> 
> **Breaking:** public `AndroidPeripheral.requestMtu` and `ServicesDiscoveredPeripheral.requestMtu` are removed. `connect()` docs note **`BLUETOOTH_CONNECT`** when MTU is non-null. **`maximumWriteValueLengthForType`** is capped by Android max attribute length (512 on API 33+, 600 on older APIs).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bce7baea101f40633bd595e90288eea13dd74e51. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


twyatt:
Note-to-self: double check proper allowable MTU ranges.

- https://github.com/dariuszseweryn/RxAndroidBle/issues/843
- https://issuetracker.google.com/issues/307234027
- https://developer.android.com/about/versions/14/behavior-changes-all#mtu-set-to-517

kable-core/src/androidMain/kotlin/Connection.kt
```diff
@@ -204,7 +205,7 @@ internal class Connection(
             }
             connectionScope.async { callback.onMtuChanged.receive() }.await()
         } catch (e: CancellationException) {
-            coroutineContext.ensureActive()
+            currentCoroutineContext().ensureActive()
```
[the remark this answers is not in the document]
twyatt: It is a new channel for every connection attempt, so a value from a previous connection would not be an issue.
With Android 14, if the system initiates the MTU negotiation, then we'll receive the result of that negotiation on the channel. We cannot overlap because Android:

> disregards all subsequent MTU requests on that ACL connection

https://developer.android.com/about/versions/14/behavior-changes-all#mtu-set-to-517

twyatt:
@cursor review

kable-core/src/androidMain/kotlin/BluetoothDeviceAndroidPeripheral.kt
```diff
@@ -128,6 +140,7 @@ internal class BluetoothDeviceAndroidPeripheral(
             )
 
             suspendUntil<State.Connecting.Services>()
+            if (desiredMtu != null) requestMtu(desiredMtu)
```
QuantumRand, line 143: If `requestMtu()` fails at this point (throwing `GattRequestRejectedException`), is it cancelling the connection? Shouldn't it be able to continue the connection with the default MTU?

twyatt: Yes, it would cancel the connection.
I believe that is desirable though.

The `requestMtu` is just that: _requesting_ an MTU. Assuming properly behaving firmware, the firmware will respond with the MTU to settle on. Hence why Android 14+ **always** (no matter what value you specify) [requests the max (`517`)](https://developer.android.com/about/versions/14/behavior-changes-all#mtu-set-to-517) and lets the firmware respond with what it actually supports.

If the MTU request fails, then I'd imagine there is a bigger problem at hand, as the firmware should really always be OK with any MTU request and simply respond with what it supports. My guess is that any failure seen at this stage will be because of a faulty connection or misbehaving firmware.

Core Bluetooth always negotiates an MTU on connection (you can't disable that). I'm guessing (I have not verified though — it would be very difficult to test), that a failure at the time of requesting an MTU, it would also result in either the firmware telling it what MTU it will accept _or_ a loss of connection.

Phoenix7351 (approved):

kable-core/src/androidMain/kotlin/BluetoothDeviceAndroidPeripheral.kt
QuantumRand: Got it. `GattRequestRejectedException` is only thrown when the _request_ is rejected, not the MTU value.

QuantumRand (approved):

kable-core/src/androidMain/kotlin/BluetoothDeviceAndroidPeripheral.kt
twyatt: Yes. That is my understanding.

Issue #51, opened 2021-01-05
Kable to support advertising as a peripheral


tom-pratt:
Is it or will it be possible in the future to use Kabel to adertise the device as the peripheral? e.g. a mobile app that can communicate with other instances of itself on multiple phones.

twyatt:
There are no short term plans to have Kable act as a GATT server, though if Kable's feature set becomes more complete I could see us working on such support.

One of the hurdles would be that (at least AFAIK) JavaScript does not have GATT server support, so the library platform support would have to diverge a bit.

We'll continue to focus on the GATT client side for a while, so don't expect much movement on this issue for some time; but I do like the idea of Kable being a full turn key BLE solution, so hopefully some day we'll be able to support this.

Happy to discuss possible designs if the community wants to try and tackle this effort and submit PRs.

tom-pratt:
Thanks for the info. Not actually an essential feature for me since the peripheral side is Android only in my project.

Have barely got started so not super familiar with BLE concepts. But my understanding was that being peripheral/central device is not coupled to gatt server/client. e.g. peripheral can be server or client. Is that true? At least that was my interpretation of these docs https://developer.android.com/guide/topics/connectivity/bluetooth-le#roles

twyatt:
> But my understanding was that being peripheral/central device is not coupled to gatt server/client. e.g. peripheral can be server or client. Is that true?

That is my understanding as well.

I suppose my response was a bit wordy; merely trying to convey that things get a little muddy (from an API perspective) to introduce "support advertising as a peripheral", purely because BLE in general is a bit convoluted. So, just feel it takes a bit of care to build up the API to support the requested feature **and do it in a way that is intuitive to developers using Kable**.

PR #875, opened 2025-03-07
Fix iOS Bluetooth read operation suspension issue

rafaelfrancisco-dev:
## Description

This pull request addresses an issue where Bluetooth read operations would suspend indefinitely on iOS devices due to an overly strict characteristic comparison in our event handling logic.

## Problem

In our characteristic event handling, we were performing an identity comparison between two `DiscoveredCharacteristic` objects. On iOS, the system creates new instances of characteristic objects during operations, resulting in different object references despite representing the same logical Bluetooth characteristic. This identity check prevented read operations from completing successfully. This issue may be related to improper implementation of the library code on our side, but the current fix resolves the problem without negatively impacting existing functionality.

## Solution

Removed the redundant identity comparison condition, allowing the code to properly compare characteristics by their UUIDs rather than by object reference. This ensures that characteristics are matched based on their logical identity rather than memory location.


twyatt:
First off, thanks for the thorough explanation and PR!

Unfortunately, I think this will adversely effect the Android implementation, which already [does a similar comparison](https://github.com/JuulLabs/kable/blob/22eab241b7890a79db2d2841f431fe77cf3949b5/kable-core/src/androidMain/kotlin/Profile.kt#L57-L66).

Ultimately, we should lean on the [overridden equality check](https://github.com/JuulLabs/kable/blob/main/kable-core/src/appleMain/kotlin/Profile.kt#L59), but fix it to properly compare by their properties (as you mentioned).

Unfortunately, unlike Android (which provides an instance ID for characteristics), Core Bluetooth exposes no such API, so if two characteristics exist with the same UUID (I admit, this is very unlikely, but it does not violate the spec) then our equality check should properly see them as equal. At this point, aside from object reference, I'm not sure how this can be accomplished w/ Core Bluetooth.

It is an option that we could allow Apple platforms to have a "degraded" implementation, but I was really hoping to avoid that (and keep the platforms aligned) — perhaps it would make sense to have the equality check use UUIDs as well as the characteristic properties? 🤷 It's not perfect, but Core Bluetooth hasn't given us many other options.

twyatt:
I just realized that on Apple platforms, instance IDs can be artificially generated during service discovery process.

I think that would be the best approach here, it would allow proper overriding of `equal`s and fix this issue for you.

Let me know if you want to update this PR for that, or if you'd like me to take over this effort (I'll try to get to it soon, but no promises on when it might be).

twyatt:
**UPDATE:** I'll try to find some time in the coming days to add some commits to this PR to resolve this.

twyatt:
Ok, I am running into some challenges with the instance ID approach. I'll look into it further in the coming days and keep you updated.

rafaelfrancisco-dev:
> On iOS, the system creates new instances of characteristic objects during operations, resulting in different object references despite representing the same logical Bluetooth characteristic.

To give you context, this problem started by our need to connect to a device that has two Device Information services (there is an issue outside our control where we can't override the default service already being advertised by the used hardware and opted to create a new service with the same exact UUID's, this, while an ugly solution, doesn't seem to be explicitly forbidden by the BLE spec). 

So to get the correct service we iterate the discovered services and characteristics and get the last one matching the UUID for the characteristic we need at the moment.

```kotlin
val discoveredServices = services
    .filterNotNull()
    .first()

val infoServiceUuid = Uuid.service("device_information")
val modelCharUuid = Uuid.characteristic("model_number_string")

val modelCharacteristics = discoveredServices.flatMap { service ->
    service.characteristics.filter { char ->
        char.serviceUuid == infoServiceUuid && char.characteristicUuid == modelCharUuid
    }
}
val lastModelChar = modelCharacteristics.lastOrNull()
    ?: throw IllegalStateException("No readable characteristic found for $modelCharUuid")

val model = read(lastModelChar)
```

On Android devices this code snippet works perfectly, on iOS however the read would suspend indefinitely due to equality check failing on the when statement I deleted on this PR:

```kotlin
eventCharacteristic is DiscoveredCharacteristic && characteristic is DiscoveredCharacteristic ->
             eventCharacteristic == characteristic
```

The eventCharacteristic **is** DiscoveredCharacteristic but the equality check would fail in _eventCharacteristic == characteristic_ which I can only surmise it fails due to being a new instance (since when we compare by UUID's the check is valid).

rafaelfrancisco-dev:
No need to apologize, take your time. In the meantime we've been using the fork, so there is no rush at all.

davertay-j:
> Simply comparing UUIDs seems insufficient to me, since 2 characteristics can technically have the same UUID (which is why [Android provides instanceId](https://developer.android.com/reference/android/bluetooth/BluetoothGattCharacteristic#getInstanceId())). I'd like to find a solution that can properly differentiate 2 different characteristics with the same UUID — if possible.

As iOS may return new instances on each delegate response we cannot use object identity here, and we also cannot use solely UUID as some devices have duplicates. I'm wondering if the solution is to uniqify by enumeration which is the [approach we settled on in Topaz browser](https://github.com/JuulLabs/topaz/blob/main/lib/Sources/BluetoothNative/RelayDelegate.swift#L215-L230)

twyatt:
> No need to apologize, take your time. In the meantime we've been using the fork, so there is no rush at all.

Thanks for being understanding/patient.

I've looked into this a lot more, and discussed by my colleagues — being that Core Bluetooth doesn't provide neither an instance ID for services nor characteristics, and sources are not available; it is difficult to know the best approach to workaround Core Bluetooth's limitations.

We have some ideas of possible workarounds[^1] but I'd really like to perform some thorough testing prior to committing to workarounds. As such, I plan to work on #51 before being able to resume work on this issue.

[^1]: The current workaround idea is to artificially create instance IDs upon service discovery — special care will have to be taken to invalidate discovered references if service discovery happens again after connection is established or upon reconnecting.

fgroeger:
Is there any update on this? We currently run into the same issue that our code works perfectly fine on Android, but on iOS the observing does not work. We are now also using a fork for now, but we'd like to use the official implementation as soon as possible again. Another idea which came to our mind was to offer a method to observe on all characteristics, then the filtering could be done on the caller side - what do you think about that?

fgroeger:
It works fine for us, thank you! 🚀

PR #890, opened 2025-04-06
Optimize scan filtering

### LINKED ISSUE #889: Android Scanner should filter by service first, then name when both are provided
Currently kable will use native filtering only if service is provided by itself. It should do service filtering natively and then use post processing for any non-native filters.

twyatt:
Closes #889

Optimizes scan filtering when non-native filter(s) are combined with native filter(s), by favoring native scan filtering and moving non-native supported filters to flow filtering.

Take the following example:

<table>
<tr>
<td align="center"><b>Code</b></td>
<td align="center"><b>Graphical representation</b></td>
</tr>
<tr>
<td>

```kotlin
filter {
    match {
        service = listOf(Filter.Service(B))
        // AND
        name = listOf(Filter.Name.Prefix("A_"))
        // AND
        serviceData = listOf(Filter.ServiceData(..))
    }
}
```
</td>
<td>

Issue 890
</td>
</tr>
</table>

This PR will optimize the scan by establishing a native filter for the `Filter.Service` and `Filter.ServiceData` and having the `Filter.Name.Prefix` filtering occur via flow operator.

| Before the PR | After the PR |
|-|-|
| Issue 890 before | Issue 890 after |

This optimization is only possible when a single `match` is specified (i.e. filters will not be `OR`'d). This is because when there are multiple filter `match` declarations, then it would be difficult to reliably determine which native filter was applied to know what "overlayed" filters should occur at the flow operator level. To illustrate this limitation, take the following example:

<table>
<tr>
<td align="center"><b>Code</b></td>
<td align="center"><b>Graphical representation</b></td>
</tr>
<tr>
<td>

```kotlin
filter {
    match {
        service = listOf(Filter.Service(A))
        // AND
        name = listOf(Filter.Name.Prefix("A_"))
    }
    // OR
    match {
        service = listOf(Filter.Service(B))
        // AND
        name = listOf(Filter.Name.Prefix("B_"))
    }
}
```
</td>
<td>

Issue 890 limitation
</td>
</tr>
</table>

Then, during a scan there isn't a clear way to filter at the flow operator:

Issue 890 limitation2

As such, for these situations (more than one `match` with non-native filters): all filters will occur at the flow operator level.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -19,6 +19,7 @@ kotlinx-coroutines-core = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-c
 kotlinx-coroutines-test = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-test", version.ref = "coroutines" }
 kotlinx-io = { module = "org.jetbrains.kotlinx:kotlinx-io-core", version = "0.7.0" }
 mockk = { module = "io.mockk:mockk", version = "1.13.17" }
+robolectric = { module = "org.robolectric:robolectric", version = "4.14.1" }
 tuulbox-collections = { module = "com.juul.tuulbox:collections", version.ref = "tuulbox" }
 tuulbox-coroutines = { module = "com.juul.tuulbox:coroutines", version.ref = "tuulbox" }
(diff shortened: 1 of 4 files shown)


QuantumRand (approved):

sdonn3 (approved):
lgtm

PR #901, opened 2025-05-12
JVM Support

### LINKED ISSUE #380: Support Linux / Windows JVM targets
Currently, the Kable Library does not support Windows or Linux targets.

There are libraries like https://github.com/weliem/blessed-bluez, that support Windows or Linux. It would be great to investigate if this could be an approach for kable to extend the platform compatibility. I really like the APIs from Kable, so I would prefer using such a library for all platforms instead of using different ones.

cedrickcooke:
Unlinking this from #380 because it does not publish a proper multi-architecture artifact. However, this code _should_ work across platforms and just require additional CI work to fix the publication.

I've tested this as working on both Windows x64 and Mac Arm (the only two machines I have conveniently available).

Parts of this are extremely inefficient: all device scanning is done without OS level filters, and Kotlin's flow-based filtering is used instead. Additionally, reconnecting a disconnected device requires scanning (this is a limitation of btleplug, see deviceplug/btleplug#413). 

# Warning

This is super experimental, and has only been tested on a small number of computers with a small number of peripherals. While I adore Rust, I am _not_ an expert and most of my experience with it is from several years ago.

# Architecture Notes

Massive shoutout to @vermasrijan19 for the proposed strategy & example. This implementation leverages [btleplug](https://github.com/deviceplug/btleplug) as a cross-platform native abstraction and [UniFFI](https://github.com/mozilla/uniffi-rs) as a JVM/JNA binding layer.

## Gradle Plugin

`uniffi-plugin` creates a full-fat gradle plugin inline within this repo. This approach was taken with the original intent of using Rust's excellent cross-compilation support to handle multiple architectures from a single build host, but I ended up giving up on that after struggling with native OS dependencies. Still, this plugin is responsible for integrating `cargo` into the gradle ecosystem and guarantees that `kable-btleplug-ffi` respects all the usual gradle commands (`assemble`, `test`, etc).

## Combo Rust Module/JVM Module

Despite `kable-btleplug-ffi` being compilable as a Kotlin/JVM module via gradle, it follows normal rust conventions for project layout. As such, it should play nice with standard IDE options like LSPs and RustRover. Development can even be done by invoking normal rust commands such as `cargo check` and `cargo test` directly.

When this module is assembled via gradle it outputs a `.jar` file containing the native dynamic library for the platform it was compiled on. Thanks to UniFFI and JNA, the correct dynamic library is loaded automatically as soon as we start calling the kotlin accessors.

### Future Work

The published `.jar` should actually contain _six_ dynamic libraries, one for each architecture-OS pair (x64/arm, windows/linux/mac). On 5/6 of the platforms, we can get away with just running `cargo build` without involving gradle. Each of those platforms needs to upload their dynamic library as an input into the final build task, which will do a proper kotlin compilation and upload the maven artifacts. Getting this to work likely requires a small update to `uniffi-plugin`'s `copyDynamicLibraryResources` task

## Kotlin Implementation

As much as possible, calls to UniFFI Kotlin are kept to a minimal surface area inside the `btleplug` sub-package. All instances of `jvmNotImplementedException` have been removed, but you'll still see exceptions if you try to make FFI calls on an architecture not included in the published artifact.


kable-core/src/jvmMain/kotlin/com/juul/kable/btleplug/BtleplugPeripheral.kt
```diff
@@ -0,0 +1,218 @@
...
+                .asSequence()
+                .flatMap { it.characteristics }
+                .single { it.uuid == uuid }
+            observers.characteristicChanges.emit(
+                CharacteristicChange(BtleplugCharacteristic(characteristic), data),
+            )
+        }
+    }
+
+    internal val ffi = com.juul.kable.btleplug.ffi.Peripheral(identifier.ffi, callbacks)
```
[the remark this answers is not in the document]
cedrickcooke: Ah, that's a good callout. "Foreign function interface." How I've tried to use it in the kotlin, that basically means "this comes from Rust." From Rust's side, it means "Kotlin can see this."

kable-btleplug-ffi/src/characteristic.rs
```diff
@@ -0,0 +1,50 @@
...
+    bits: u8,
+}
+
+impl From<Characteristic> for btleplug::api::Characteristic {
+    fn from(value: Characteristic) -> Self {
+        btleplug::api::Characteristic {
+            uuid: value.uuid.into(),
+            service_uuid: value.service.into(),
+            properties: value.properties.into(),
+            descriptors: value.descriptors.into_iter().map(|x| x.into()).collect(),
```
davertay-j, line 24: nit: no need for the lamda, should be able to pass the method directly:
```suggestion
            descriptors: value.descriptors.into_iter().map(Descriptor::into).collect(),
```

kable-btleplug-ffi/src/characteristic.rs
```diff
@@ -0,0 +1,50 @@
...
+    }
+}
+
+impl From<btleplug::api::Characteristic> for Characteristic {
+    fn from(value: btleplug::api::Characteristic) -> Self {
+        Characteristic {
+            uuid: value.uuid.into(),
+            service: value.service_uuid.into(),
+            properties: value.properties.into(),
+            descriptors: value.descriptors.into_iter().map(|x| x.into()).collect(),
```
davertay-j, line 35: nit:
```suggestion
            descriptors: value.descriptors.into_iter().map(btleplug::api::Descriptor::into).collect(),
```

Plenty more of these in other files but is a nit so I'll not comment on them all, sometimes it reads better, sometimes it doesn't and I don't have a strong opinion on unecessary lamda usage so use whatever style you prefer.

kable-btleplug-ffi/src/peripheral.rs
```diff
@@ -0,0 +1,257 @@
...
+                .is_ok()
+        } else {
+            false
+        }
+    }
+
+    async fn disconnect(&self) -> bool {
+        match self.get_platform().await {
+            Err(_) => true,
+            Ok(platform) => platform.disconnect().await.is_ok(),
```
davertay-j, line 181: This is probably fine, but is confusing to me. Can you please explain why we return `true` on error, and `true` on Ok, and `false` on "not Ok"?

kable-btleplug-ffi/src/peripheral_id.rs
```diff
@@ -0,0 +1,73 @@
...
+pub struct PeripheralId {
+    pub platform: btleplug::platform::PeripheralId,
+}
+
+#[cfg(target_os = "macos")]
+#[uniffi::export]
+impl PeripheralId {
+    #[uniffi::constructor]
+    fn new(value: String) -> Self {
+        PeripheralId {
```
davertay-j, line 20: nit, unsure if more "rusty" or not to use `Self` in constructors? Asking because we have not been consistent here vs other structs in this module.
```suggestion
        Self {
```

kable-core/src/jvmMain/kotlin/com/juul/kable/btleplug/BtleplugAdvertisement.kt
```diff
@@ -0,0 +1,48 @@
...
+        identifier = Identifier(properties.id),
+        rssi = properties.rssi?.toInt() ?: Int.MIN_VALUE,
+        txPower = properties.txPowerLevel?.toInt(),
+        uuids = properties.services.map(Uuid::parse),
+    )
+
+    override val isConnectable: Boolean?
+        get() = null // Doesn't exist in btleplug
+
+    // TODO: Figure out why Btleplug exposes this as a map.
```
davertay-j, line 37: Re this comment, Web APIs also define a map. https://webbluetoothcg.github.io/web-bluetooth/#bluetoothmanufacturerdatamap

Not particularly useful but there you go. :-)

davertay-j (approved):
In the words of ProjectFarm, "Very Impressive!"

kable-core/src/jvmMain/kotlin/com/juul/kable/btleplug/BtleplugAdvertisement.kt
cedrickcooke: > Instances of [BluetoothManufacturerDataMap](https://webbluetoothcg.github.io/web-bluetooth/#bluetoothmanufacturerdatamap) have a [[BackingMap]] slot because they are [maplike](https://webidl.spec.whatwg.org/#dfn-maplike),

Wow, you weren't kidding about not particularly useful.

kable-btleplug-ffi/src/peripheral_id.rs
cedrickcooke: Good point. A quick search says `Self {` for the constructor might be more idiomatic.

kable-btleplug-ffi/src/characteristic.rs
cedrickcooke: Turns out there's a default-off lint enforcing this. I've enabled it for the project so there'll be no decision making in the future

kable-btleplug-ffi/src/peripheral_id.rs
cedrickcooke: Same here re. a default-off lint. Enabled going forward.

kable-btleplug-ffi/src/peripheral.rs
davertay-j: Cool, thanks for explaining.

kable-btleplug-ffi/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "btleplug-ffi"
+edition = "2024"
```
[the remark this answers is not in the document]
cedrickcooke: It's the most up to date version, they're on a 3-yr cadence so far.

Editions are [complicated](https://doc.rust-lang.org/nightly/edition-guide/editions/index.html), but the tl;dr is that new projects should always use the most recent one. Older editions are still supported so the language can change around things like keywords without breaking builds/compatibility.

twyatt (commented):
Pass 2/3

I'll continue reviewing more later tonight.

kable-core/src/jvmMain/kotlin/com/juul/kable/btleplug/BtleplugPeripheral.kt
```diff
@@ -0,0 +1,218 @@
...
+        return withContext(Dispatchers.IO) {
+            ffi.writeDescriptor(
+                getDescriptor(
+                    descriptor.serviceUuid.toString(),
+                    descriptor.characteristicUuid.toString(),
+                    descriptor.descriptorUuid.toString(),
+                ),
+                data,
+            )
+        }
```
twyatt, line 202: How does this operation handle cancellation (i.e. when caller, on the Kotlin side, cancels the coroutine)?
Does the rust side honor the cancellation? Does it even get the signal?

twyatt (approved):
The Rust portions are out of my wheelhouse, so I didn't review those portions too closely, but in general it is fairly readable and easy to follow. 🏅 

When trying run the Gradle tasks to generate the bindings it failed with:

```
> Task :kable-btleplug-ffi:cargoBuild FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':kable-btleplug-ffi:cargoBuild'.
> A problem occurred starting process 'command 'cargo'
```

This is obviously expected, but wondering what your thoughts are on later migrating to using something like: https://ysb33rorg.gitlab.io/rust-gradle-plugin/0.2/docs/product-documentation.html
_(I'd imagine it'd make it easier for contributors to spin up the project?)_

Overall, fantastic and amazing job on this!!! 💯

twyatt:
So that I don't have to do a release branch dance, can you wait to merge this until after #958 is merged?

cedrickcooke:
> This is obviously expected, but wondering what your thoughts are on later migrating to using something like: https://ysb33rorg.gitlab.io/rust-gradle-plugin/0.2/docs/product-documentation.html
(I'd imagine it'd make it easier for contributors to spin up the project?)

I would _love_ to build off of a pre-existing Rust plugin, but everything I found was either super old/unmaintained (the library you linked, as an example, last updated in 2018), or designed for Android NDK (like https://github.com/mozilla/rust-android-gradle)

uniffi-plugin/src/main/kotlin/tasks/UniffiBindgenTasks.kt
```diff
@@ -0,0 +1,92 @@
...
+"""
+
+internal fun TaskContainer.registerUniffiBindgenTasks(accessor: UniffiKotlinExtensionAccessor) {
+    named("clean") { dependsOn("cleanUniffiBindgenCargoProject") }
+    named("compileKotlin") { dependsOn("generateKotlinBindings") }
+
+    register("cleanUniffiBindgenCargoProject") {
+        group = UNIFFI_TASK_GROUP
+
+        doLast { project.uniffiBindgenProject.deleteRecursively() }
```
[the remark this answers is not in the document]
cedrickcooke: The `register` block itself runs at configuration time, so we need a `doLast` (or `doFirst`, doesn't really matter) to specify code to run during task execution.

uniffi-plugin/settings.gradle.kts
```diff
@@ -0,0 +1,15 @@
...
+        mavenCentral()
+    }
+}
+
+pluginManagement {
+    repositories {
+        gradlePluginPortal()
+        mavenCentral()
+    }
+}
```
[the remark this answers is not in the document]
cedrickcooke: Removed what I could. Looks like I don't need mavenCentral, but I do need GPP

uniffi-plugin/build.gradle.kts
```diff
@@ -0,0 +1,27 @@
...
+            displayName = "uniffi-kotlin"
+            implementationClass = "com.juul.kable.uniffi.plugin.UniffiKotlinPlugin"
+        }
+    }
+}
+
+val compileKotlin: KotlinCompile by tasks
+compileKotlin.compilerOptions {
+    languageVersion.set(KotlinVersion.KOTLIN_2_0)
+}
```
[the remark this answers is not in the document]
cedrickcooke: Turns out I didn't 😅

uniffi-plugin/settings.gradle.kts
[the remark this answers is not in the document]
twyatt: Isn't GPP default (i.e. would be used if these blocks were specified at all)?

uniffi-plugin/src/main/kotlin/ProjectExtensions.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import org.gradle.api.Project
+import java.io.File
+
+internal val Project.uniffiOutputDirectory: File
+    get() = this.layout.buildDirectory.asFile.get()
+        .resolve("generated")
+        .resolve("uniffiKotlin")
+        .resolve(KOTLIN_SOURCE_SET)
+        .resolve("kotlin")
```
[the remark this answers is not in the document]
cedrickcooke: I think this is fine/already lazy due to the use of `get() =` instead of a plain `=` for this value

uniffi-plugin/settings.gradle.kts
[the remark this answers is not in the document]
cedrickcooke: Looks like it's a default for `pluginManagement`, but I still need it for `dependencyResolutionManagement`.

PR #974, opened 2025-07-26
Support `jvm` label in release drafts and use emojis for logos

twyatt:
Adds support for `jvm` platform label in release drafts. Replaces platform logos with emojis; this makes it easier to add additional platforms.


sdonn3 (approved):
🔥

Issue #1016, opened 2025-08-27
Support multiple characteristics with the same UUID


twyatt:
Issue to cover bug reported via #875.

PR #1017, opened 2025-08-27
Add `forceCharacteristicEqualityByUuid` peripheral builder option

twyatt:
Provides an option as a workaround until #1016 can be tackled.


fgroeger:
I checked out this solution and it works fine for our use case. Thanks! Looking forward to see it integrated into the next release. 🎉

cedrickcooke (approved):

sdonn3 (approved):

PR #1028, opened 2025-09-15
Expose internal/private platform specific fields

zerodea:
Expose platform specific fields to allow users to handle more advanced Bluetooth features after using Kable for discovery and/or connection

This change will allow users to implement their own advanced features, such as L2CAP,  while the features are added to Kable, or even if the features are not added.

I haven't exposed JVM and JS fields because I'm not familiar with the platforms, and whether it is of interest. If it is I could research it and see what fields to expose, on JS it seems to be BluetoothDevice, but I haven't been able to figure it out at a quick glance for JVM (btleplug).

---

> [!NOTE]
> **Medium Risk**
> Adds a public-but-opt-in surface to raw platform objects, which can bypass Kable’s abstractions and break if removed; core connection/GATT paths are otherwise unchanged.
> 
> **Overview**
> Adds a **`@KableInternalApi`** opt-in (ERROR level) and wires it into the module’s language settings so callers can deliberately use unstable escape hatches.
> 
> On **Android**, connected peripherals expose **`BluetoothDevice`** via `AndroidPeripheral.bluetoothDevice`, and scan advertisements expose the same through **`PlatformAdvertisement.bluetoothDevice`** (with the Android implementation promoting what was previously an internal property). On **Apple**, **`CoreBluetoothPeripheral.cbPeripheral`** and **`PlatformAdvertisement.cbPeripheral`** expose the underlying Core Bluetooth peripheral from connected devices and advertisements.
> 
> Implementations are updated to satisfy these interfaces; JVM/JS are unchanged in this diff.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8da0face9017db763e8b2c1779f0861105576afd. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


zerodea:
I understand and your hesitation makes sense, exposing internal fields is always a risk. I will keep working on the L2CAP PR. I'm working on it for a project and have a pretty clean Android implementation, but I still have to work on the iOS code and test both thoroughly.

cedrickcooke:
@twyatt how would you feel about an opt-in annotation like `InternalKableApi`, with clear messaging about both the opportunity to create bugs and also no promise of API stability?

twyatt:
> @twyatt how would you feel about an opt-in annotation like `InternalKableApi`, with clear messaging about both the opportunity to create bugs and also no promise of API stability?

Ultimately, I want Kable to be useful for as many people as possible, so I would like to make fields available if they might help some consumers.

`@InternalKableApi` is a good idea to create some barrier at the code level.

My only remaining concern would be getting bug reports after someone used an "internal" API. So maybe we log a very loud log statement upon first access of an "internal" API? That way when they share logs it becomes very apparent that their problems may be from the API usage vs. a Kable bug. 🤷

zerodea:
I have created an initial implementation of what the annotation could be like if that's the way you want to take it. 

I also added another commit with a possible implementation of the log warning. It's a bit messy, but I couldn't think of a way to only show the message once and to only show it when the property is accessed from outside the Kable library. If you like the way the log message was handled I could add it to the other places where @KableInternalApi is being used.

twyatt (approved):
@zerodea apologies for the very long delay getting to this. Ultimately decided the log message was unnecessary (can be added later if needed). Updated PR accordingly.

Phoenix7351 (approved):

Issue #765, opened 2024-09-29
Adding WASM target artifact 


vermasrijan19:
Similar to this https://github.com/JuulLabs/kable/pull/626

Currently is not possible to compile shared if wasm target exists (as there is no corresponding implementation in kable).
Solution: Provide an empty implementation that throws runtime exceptions for kable for wasm target target.

twyatt:
Good thinking. Unfortunately it will be some time before I can allocate energy to this effort.

yuroyami:
This would be greatly useful since we have to manually copy the source code to get the whole project to sync

PR #1099, opened 2026-01-20
Kotlin/Wasm support

cedrickcooke:
Adds support for Kotlin/Wasm. Tested that Sensortag (see: JuulLabs/sensortag#531) works on both JS and WasmJS targets.

Highest risk in this PR is probably around exception handling. I think I got the pattern correct, it's just the place I'm least confident about and hard to test.

Solves #765

### DIFF
--- a/kable-btleplug-ffi/Cargo.lock
+++ b/kable-btleplug-ffi/Cargo.lock
@@ -119,7 +119,7 @@ dependencies = [
  "log",
  "serde",
  "serde-xml-rs",
- "thiserror 2.0.17",
+ "thiserror 2.0.18",
  "tokio",
  "uuid",
 ]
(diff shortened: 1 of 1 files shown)


kable-core/src/androidMain/kotlin/BluetoothDeviceAndroidPeripheral.kt
```diff
@@ -41,7 +42,6 @@ import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
-import kotlin.coroutines.cancellation.CancellationException
```
twyatt: Is there a reason your changing which `CancellationException` we're using?

Gemini suggests:
Image

PR #1137, opened 2026-05-01
Honor caller cancellation during `connect`

### LINKED ISSUE #1136: IllegalStateException when I cancel a Peripheral.connect() call
If our app is part-way through connecting a Peripheral and someone exits the screen, we cancel the coroutine scope of the connect task. But inside Kable's connect call, it has a try-catch for `IllegalStateException` and rethrows its own `IllegalStateException: Cannot connect peripheral that has been cancelled `:

https://github.com/JuulLabs/kable/blob/368d4d5b16ae7ed46c89fac331b0788183a94ed7/kable-core/src/commonMain/kotlin/SharedRepeatableAction.awaitConnect.kt#L5-L10

But `CancellationException` inherits from `IllegalStateException`, so any time you cancel the connect call, it (seemingly unintentionally) catches that cancellation and turns it into `IllegalStateException`, causing our app to crash from that unexpected exception. Am I correct in thinking this is a bug?

I'm working around it with this:

```kotlin
try {
    connect()
} catch (e: IllegalStateException) {
    // Work around Kable seemingly accidentally catching CancellationException and throwing
    // an IllegalStateException. This call throws CancellationException if we're cancelled.
    currentCoroutineContext().ensureActive()

    throw e
}
```

twyatt:
Closes #1136

---

> [!NOTE]
> **Low Risk**
> Minimal change adding proper cancellation propagation; fixes a bug rather than introducing new behavior.
> 
> **Overview**
> Fixes a bug where coroutine cancellations were being swallowed during peripheral connection. `CancellationException` is a subclass of `IllegalStateException`, so the existing catch block was unintentionally masking caller cancellations. Adds `ensureActive()` check to properly propagate cancellation before wrapping other `IllegalStateException`s.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0b99ae1af8f1be90a2743bd08ead6a1d63d0361f. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/kable-core/src/commonMain/kotlin/SharedRepeatableAction.awaitConnect.kt
+++ b/kable-core/src/commonMain/kotlin/SharedRepeatableAction.awaitConnect.kt
@@ -1,10 +1,17 @@
 package com.juul.kable
 
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.currentCoroutineContext
+import kotlinx.coroutines.ensureActive
 
 internal suspend fun SharedRepeatableAction<CoroutineScope>.awaitConnect(): CoroutineScope =
     try {
         await()
     } catch (e: IllegalStateException) {
+        // If calling code cancels we'd unintentionally swallow the `CancellationException`, as it
+        // is a subclass of `IllegalStateException`. `ensureActive()` honors caller cancellation.
+        // https://github.com/JuulLabs/kable/issues/1136
+        currentCoroutineContext().ensureActive()
+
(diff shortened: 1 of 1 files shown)


twyatt:
@cursor review

davertay-j (approved):

cedrickcooke (approved):

QuantumRand (approved):

PR #1202, opened 2026-07-10
Restore `kotlin-parcelize-runtime` dependency in published Android metadata

### LINKED ISSUE #1201: 0.44.0 regression: `kotlin-parcelize-runtime` dropped from kable-core-android published metadata — R8 full mode fails with "Missing class kotlinx.parcelize.Parcelize"
## Summary

`kable-core-android` bytecode references `kotlinx.parcelize.Parcelize` (`ScanResultAndroidAdvertisement` is `@Parcelize`), but as of 0.44.0 the published Gradle module metadata and POM no longer declare a dependency on `org.jetbrains.kotlin:kotlin-parcelize-runtime`. Android apps consuming Kable that don't themselves apply the `kotlin-parcelize` plugin now fail release builds under R8 full mode (the default since AGP 8):

```
Missing class kotlinx.parcelize.Parcelize (referenced from: com.juul.kable.ScanResultAndroidAdvertisement)
```

## Regression in 0.44.0

Verified against the Maven Central artifacts:

| Published metadata | 0.43.1 | 0.44.0 |
|---|---|---|
| `.module` (what Gradle consumers resolve) | `releaseRuntimeElements-published` declares `kotlin-parcelize-runtime:2.4.0` | no parcelize entry in any variant |
| POM | `kotlin-parcelize-runtime` 2.4.0, runtime scope | dropped |
| AAR consumer ProGuard rules | none | none |

## Likely cause

#1124 migrated `kable-core` from `com.android.library` + `androidTarget().publishLibraryVariants("debug", "release")` to the new AGP KMP plugin `com.android.kotlin.multiplatform.library` (AGP 9.2.1). Under the old setup, the runtime dependency that the `kotlin-parcelize` plugin adds was exported into the published variants; under the new plugin it is silently lost. (Possibly an AGP/Kotlin interop bug worth escalating to Google/JetBrains, but the consumer-visible breakage is in the Kable artifact.)

## Reproduction

1. Android app with `minifyEnabled true` (R8 full mode) depending on `com.juul.kable:kable-core:0.44.0`, without applying `kotlin-parcelize` itself.
2. Build a release variant → fails with the missing-class error above.

Hit in [Meshtastic-Android #6196](https://github.com/meshtastic/Meshtastic-Android/pull/6196); worked around consumer-side with `-dontwarn kotlinx.parcelize.Parcelize`.

## Suggested fix (either)

For context: the only reference into `kotlinx.parcelize` across the entire shipped AAR is the `@Parcelize` annotation itself (BINARY retention — present in bytecode, never loaded at runtime), so both options below are safe:

- Restore the `kotlin-parcelize-runtime` dependency in the published metadata for the android target (as in ≤ 0.43.1). This is what the `kotlin-parcelize` plugin itself does for Android compilations; the artifact is a 15 KB annotations-only jar whose sole dependency is `kotlin-stdlib`, and R8 strips it from consumer APKs anyway. Or,
- Ship a consumer ProGuard rule in the AAR (`-dontwarn kotlinx.parcelize.Parcelize`) — precedent for this exact pattern: [stripe/stripe-android#11422](https://github.com/stripe/stripe-android/issues/11422). Avoids the dependency, at the cost of a blanket suppression in every consumer's R8 config that would mask a real missing class if Kable ever adopts `Parceler`/`TypeParceler`.

**Workaround for consumers:** add `-dontwarn kotlinx.parcelize.Parcelize` to your ProGuard rules, or apply the `kotlin-parcelize` plugin in the app.

I've opened companion PRs implementing each option — merge whichever you prefer and close the other:

- #1202 — restore the `kotlin-parcelize-runtime` dependency (matches ≤ 0.43.1 published shape)
- #1203 — ship the `-dontwarn` consumer keep rule in the AAR (no dependency)

jamesarich:
## Why

The 0.43.1 → 0.44.0 migration (#1124) from `com.android.library` to `com.android.kotlin.multiplatform.library` stopped propagating the runtime dependency that the `kotlin-parcelize` plugin adds into the published Gradle module metadata and POM for `kable-core-android`:

| Published metadata | 0.43.1 | 0.44.0 |
|---|---|---|
| `.module` | `releaseRuntimeElements-published` declares `kotlin-parcelize-runtime:2.4.0` | no parcelize entry in any variant |
| POM | `kotlin-parcelize-runtime`, runtime scope | dropped |

Since `ScanResultAndroidAdvertisement` is `@Parcelize`, consumers that don't apply `kotlin-parcelize` themselves now fail R8 full-mode builds with `Missing class kotlinx.parcelize.Parcelize`. Details in #1201.

## What

Declare the dependency explicitly in `androidMain`, restoring the ≤ 0.43.1 published shape. This is what the `kotlin-parcelize` plugin itself does for Android compilations; the artifact is a 15 KB annotations-only jar whose sole dependency is `kotlin-stdlib`, and R8 strips the unused annotation classes from consumer APKs. (The non-propagation may ultimately be a KGP/AGP interop bug worth escalating upstream, but the explicit declaration is harmless, self-documenting, and unbreaks consumers now.)

## Verification

`./gradlew :kable-core:generateMetadataFileForAndroidPublication :kable-core:generatePomFileForAndroidPublication` on this branch: the `.module` again declares `kotlin-parcelize-runtime` in `androidRuntimeElements-published`, and the POM carries it at runtime scope, version `2.4.0` (tracks the `kotlin` catalog version).

Alternative to #1203, which ships a `-dontwarn` consumer keep rule in the AAR instead of the dependency. Merge whichever you prefer and close the other.

Fixes #1201.

---

> [!NOTE]
> **Low Risk**
> Build-metadata and a small annotations-only runtime dependency on Android; no auth, data, or core BLE logic changes.
> 
> **Overview**
> After the move to `com.android.kotlin.multiplatform.library`, **`kotlin-parcelize-runtime` stopped appearing** in published `.module` / POM metadata for `kable-core-android`, so apps using `@Parcelize` types (e.g. `ScanResultAndroidAdvertisement`) without applying `kotlin-parcelize` themselves could hit R8 **Missing class kotlinx.parcelize.Parcelize**.
> 
> This PR **declares `kotlin-parcelize-runtime` explicitly** on `androidMain` (catalog entry + `implementation` in `kable-core/build.gradle.kts`), matching the ≤0.43.1 published dependency shape and fixing consumer builds.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d5bab061ff51f3a0649b512f77416192c5dbca46. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -26,6 +26,7 @@ jna = { module = "net.java.dev.jna:jna", version.ref = "jna" }
 kable = { module = "com.juul.kable:kable-core" }
 kable-permissions = { module = "com.juul.kable:kable-default-permissions" }
 khronicle = { module = "com.juul.khronicle:khronicle-core", version = "1.1.0" }
+kotlin-parcelize-runtime = { module = "org.jetbrains.kotlin:kotlin-parcelize-runtime", version.ref = "kotlin" }
 kotlin-plugin = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlin" }
 kotlinx-browser = { module = "org.jetbrains.kotlinx:kotlinx-browser", version = "0.5.0" }
 kotlinx-coroutines = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-core", version.ref = "coroutines" }
--- a/kable-core/build.gradle.kts
+++ b/kable-core/build.gradle.kts
@@ -83,6 +83,7 @@ kotlin {
             api(libs.kotlinx.coroutines.android)
             implementation(libs.androidx.core)
             implementation(libs.androidx.startup)
+            implementation(libs.kotlin.parcelize.runtime)
             implementation(libs.tuulbox.coroutines)
         }
 


cedrickcooke (approved):

twyatt (approved):

jamesarich:
Thanks kable gang - we at Meshtastic appreciate your work :green_heart:

PR #1214, opened 2026-07-15
Build Linux dynamic libraries with cross for older glibc compatibility

### LINKED ISSUE #990: Fix linked glibc version in Linux (JVM) builds
Spun up a Raspberry Pi to test #989, and Sensortag scans crash immediately. JNA fails to load kable-btleplug-ffi because the binary is linked with glibc 2.38, but Raspberry Pi OS only has glibc up to 2.36.

Easiest solution is probably building with [cross](https://github.com/cross-rs/cross) on the Linux runners. At time of writing, their default toolchain is on glibc 2.31 and they also provide a toolchain built on centos 7/glibc 2.17

twyatt:
Resolves https://github.com/JuulLabs/kable/issues/990

## Problem

The Linux `kable-btleplug-ffi` dynamic libraries are built directly on GitHub's Ubuntu runners, so they link against the runner's glibc (2.38+). On distributions with an older glibc (e.g. Raspberry Pi OS, glibc ≤ 2.36), JNA fails to load the library and Sensortag scans crash immediately.

## Solution

As suggested in the issue, Linux builds now go through [`cross`](https://github.com/cross-rs/cross), using its CentOS 7 based images (`main-centos` tags, pinned via `kable-btleplug-ffi/Cross.toml`) so the produced `.so` files link against **glibc 2.17**.

> [!NOTE]
> The issue mentions cross's default toolchain being on glibc 2.31, but cross's default images have since moved to Ubuntu 24.04 (glibc 2.39), which would reproduce the original problem — hence the CentOS 7 images, which the issue also mentions as an option.

### `uniffi-plugin`

- New optional `cargoCommand` extension property: command used to build the dynamic library (defaults to `cargo`). CI sets it to `cross` for Linux builds via the `UNIFFI_CARGO` environment variable.
- New optional `target` extension property: Rust target triple to build the dynamic library for (defaults to the host's target). CI sets it via the `UNIFFI_TARGET` environment variable. This was needed because cross does not publish arm64 host images ([cross-rs/cross#751](https://github.com/cross-rs/cross/issues/751)), so the `aarch64-unknown-linux-gnu` library is now cross-compiled from the x86_64 runner instead of being built natively on `ubuntu-24.04-arm`.
- `generateKotlinBindings` still runs `uniffi-bindgen` with host `cargo`; only the `--library` it parses points at the (possibly cross-compiled) target library, which works since uniffi-bindgen reads metadata from the binary statically.

### Workflows (`ci.yml` / `publish.yml`)

- The `assemble-rust` matrix replaces the `ubuntu-24.04-arm` + `ubuntu-latest` runners with two `ubuntu-latest` entries (one per Linux target) that install cross (from git `main`, matching the pinned `main-centos` image tags) and export `UNIFFI_CARGO`/`UNIFFI_TARGET`.
- Linux artifact names are now keyed by target triple instead of runner name; macOS/Windows artifacts are unchanged.

Local (non-CI) builds are unaffected: without the environment variables set, the plugin behaves exactly as before (`cargo build --target <host triple>`).

## Verification

Performed locally (Docker + cross installed from git main, same as CI):

- `UNIFFI_CARGO=cross UNIFFI_TARGET=aarch64-unknown-linux-gnu ./gradlew :kable-btleplug-ffi:assemble` → jar contains `linux-aarch64/libbtleplug_ffi.so`, `objdump -T` shows max versioned symbol `GLIBC_2.17`.
- `UNIFFI_CARGO=cross ./gradlew :kable-btleplug-ffi:cargoBuild` (x86_64) → max versioned symbol `GLIBC_2.16`.
- Plain `./gradlew :kable-btleplug-ffi:assemble` (no env vars) → unchanged native host build.

<div><a href="https://cursor.com/agents/bc-135a7b9f-73bd-4501-aa03-803093507b6a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-135a7b9f-73bd-4501-aa03-803093507b6a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,11 +14,20 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runner:
-          - macos-15-intel
-          - ubuntu-24.04-arm
-          - ubuntu-latest
-          - windows-latest
+        include:
+          - runner: macos-15-intel
+          - runner: windows-latest
+
+          # Linux dynamic libraries are built with `cross` (in Docker containers) to link against
+          # an older glibc than provided by the GitHub runners, for compatibility with older Linux
+          # distributions (e.g. Raspberry Pi OS).
+          # https://github.com/JuulLabs/kable/issues/990
+          - runner: ubuntu-latest
+            cargo: cross
+            target: x86_64-unknown-linux-gnu
+          - runner: ubuntu-latest
+            cargo: cross
+            target: aarch64-unknown-linux-gnu
     runs-on: ${{ matrix.runner }}
     steps:
       - uses: actions/checkout@v7
@@ -26,13 +35,18 @@ jobs:
         with:
           java-version-file: .sdkmanrc
       - uses: gradle/actions/setup-gradle@v6
+      - if: matrix.cargo == 'cross'
+        run: cargo install cross --git https://github.com/cross-rs/cross
       - run: ./gradlew :kable-core:compileKotlinJvm
+        env:
+          UNIFFI_CARGO: ${{ matrix.cargo }}
+          UNIFFI_TARGET: ${{ matrix.target }}
       - run: unzip kable-btleplug-ffi.jar && rm -rf META-INF com *.jar
         working-directory: kable-btleplug-ffi/build/libs
         shell: bash
       - uses: actions/upload-artifact@v7
         with:
-          name: kable-btleplug-ffi-${{ matrix.runner }}
+          name: kable-btleplug-ffi-${{ matrix.target || matrix.runner }}
           path: kable-btleplug-ffi/build/libs/
           if-no-files-found: error
           retention-days: 1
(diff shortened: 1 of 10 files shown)


jamesarich:
built this branch with cross - x86-64 and aarch64 `.so` both load where 0.45.0 fails on GLIBC_2.38 (debian 10/11/12, ubuntu 22.04, checked on a uConsole for arm64), no regression on 24.04 / debian 13. would unblock our deb/AppImage on Pi OS and 22.04

PR #1231, opened 2026-07-20
Add L2CAP channel support

### LINKED ISSUE #810: Add support for L2CAP
An alternative to GATT for BLE communication is to use L2CAP. This let's use use a connection-oriented channel towards the peripheral instead of the message-based method GATT uses. Think of this like TCP vs UDP. On an API level we two streams (InputStream and OutputStream), so it makes it much easier to work with. 

This is supported by both Android and iOS, but doesn't exist for Web.
https://developer.apple.com/documentation/corebluetooth/cbperipheral/openl2capchannel(_:)
https://developer.android.com/reference/kotlin/android/bluetooth/BluetoothDevice?hl=en#createl2capchannel

I suggest this is added as an additional API in Kable. If this sounds good (despite the API missing on Web) I can start on an initial implementation.

torfinnberset:
Adapted from #1023, and completing the implementation for both iOS and Android. Tested on both platforms with real devices, and also from macOS to iOS. 

Adds `L2CapSocket`, a connection-oriented channel opened from a connected peripheral: `AndroidPeripheral.openL2CapChannel` / `openInsecureL2CapChannel` and `CoreBluetoothPeripheral.openL2CapChannel`. The socket is a bidirectional byte stream (`read` into a buffer, `write` a packet, `close`) exposing `isConnected` / `hasReachedEof` state; failures surface as `L2CapException`.

Android wraps `BluetoothDevice.createL2capChannel`, connecting on `Dispatchers.IO`. Apple confines the `CBL2CAPChannel` NSStreams to a single dedicated run-loop thread (they are not safe to drive from a rotating dispatcher pool) and serializes opens through the connection's guard, since CoreBluetooth's `didOpenL2CAPChannel` callback carries no PSM and so cannot be correlated to a request; a channel delivered to a cancelled or disconnected open is torn down rather than leaked.

Closes #810 #588 #1023.

---

> [!NOTE]
> **Medium Risk**
> New BLE transport path with non-trivial lifecycle (cancellation, single in-flight Apple opens, dedicated run-loop threading), but changes are additive and isolated from existing GATT APIs.
> 
> **Overview**
> Adds **L2CAP CoC** as a bidirectional byte stream alongside GATT, with a shared **`L2CapSocket`** API (`incoming` `Flow`, suspend `write`/`close`, `isConnected`) and **`L2CapException`** for open/I/O failures.
> 
> **Android (API 29+):** `AndroidPeripheral` gains `openL2CapChannel` and `openInsecureL2CapChannel`, connecting via `BluetoothSocket` on IO and wrapping **`AndroidL2CapSocket`** (back-pressure-friendly unbuffered reads).
> 
> **Apple:** `CoreBluetoothPeripheral.openL2CapChannel` routes through **`Connection.openL2CapChannel`**, which serializes opens under the connection mutex because `didOpenL2CAPChannel` has no PSM; **`PeripheralDelegate`** now completes pending opens and disposes stray channels on cancel/disconnect. **`AppleL2CapSocket`** drives `CBL2CAPChannel` streams on a dedicated run-loop thread.
> 
> README documents usage; JVM `.api` is updated; Robolectric and Apple unit tests cover read pacing, EOF, errors, and close behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d9f6ae150fe6d7c5ef958b597f7e5275908ffa56. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -548,27 +548,21 @@ On Apple platforms:
 val socket = (peripheral as CoreBluetoothPeripheral).openL2CapChannel(psm = 0x0080)
 ```
 
-The returned `L2CapSocket` is already open. Read chunks of incoming bytes, and write whole packets:
+The returned `L2CapSocket` is already open. Incoming bytes arrive as a `Flow` of chunks that completes at
+end of stream, and packets are written whole:
 
 ```kotlin
-val chunk = socket.read() // suspends until data arrives; returns null at end-of-stream
+socket.incoming.collect { chunk ->
+    // Process chunk.
+}
 
 socket.write(byteArrayOf(1, 2, 3))
 
 socket.close()
 ```
 
-Incoming data is also available as a cold `Flow`, which completes at end-of-stream:
-
-```kotlin
-socket.incoming().collect { chunk ->
-    // Process chunk.
-}
-```
-
 > [!NOTE]
-> _Opening a channel throws `L2CapException` on failure. `read` should be called from a single coroutine
-> at a time; `write` may run concurrently with it._
+> _A failure while opening, reading or writing throws `L2CapException`._
 
 ### Observation
 
--- a/README.md
+++ b/README.md
@@ -548,15 +548,22 @@ On Apple platforms:
 val socket = (peripheral as CoreBluetoothPeripheral).openL2CapChannel(psm = 0x0080)
 ```
 
-The returned `L2CapSocket` is already open. Read into a caller-provided buffer, and write whole packets:
+The returned `L2CapSocket` is already open. Read chunks of incoming bytes, and write whole packets:
 
 ```kotlin
-val buffer = ByteArray(1024)
-val count = socket.read(buffer) // suspends until data arrives; returns -1 at end-of-stream
+val chunk = socket.read() // suspends until data arrives; returns null at end-of-stream
 
 socket.write(byteArrayOf(1, 2, 3))
 
 socket.close()
+```
+
+Incoming data is also available as a cold `Flow`, which completes at end-of-stream:
+
+```kotlin
+socket.incoming().collect { chunk ->
+    // Process chunk.
+}
 ```
 
 > [!NOTE]
--- a/README.md
+++ b/README.md
@@ -527,6 +527,42 @@ When a write type is not specified, [`WithoutResponse`] is used.
 > [!NOTE]
 > _Write type only applies to characteristic writes (descriptor writes are always acknowledged by the peripheral)._
 
+### L2CAP
+
+In addition to GATT, a connected peripheral can open an L2CAP connection-oriented channel (CoC) on a PSM
+(protocol/service multiplexer). Unlike GATT's message-based characteristics, a channel is a bidirectional
+byte stream, so callers are responsible for framing their own protocol. L2CAP is available on Android and
+Apple platforms (it has no Web Bluetooth equivalent), so a channel is opened from the platform-specific
+peripheral.
+
+On Android (requires API level 29 or higher):
+
+```kotlin
+val socket = (peripheral as AndroidPeripheral).openL2CapChannel(psm = 0x0080)
+// or `openInsecureL2CapChannel(...)` for a channel without authentication/encryption
+```
+
+On Apple platforms:
+
+```kotlin
+val socket = (peripheral as CoreBluetoothPeripheral).openL2CapChannel(psm = 0x0080)
+```
+
+The returned `L2CapSocket` is already open. Read into a caller-provided buffer, and write whole packets:
+
+```kotlin
+val buffer = ByteArray(1024)
+val count = socket.read(buffer) // suspends until data arrives; returns -1 at end-of-stream
+
+socket.write(byteArrayOf(1, 2, 3))
+
+socket.close()
+```
+
+> [!NOTE]
+> _Opening a channel throws `L2CapException` on failure. `read` should be called from a single coroutine
+> at a time; `write` may run concurrently with it._
+
 ### Observation
 
 Bluetooth Low Energy provides the capability of subscribing to characteristic changes by means of notifications and/or
--- a/kable-core/src/androidMain/kotlin/AndroidL2CapSocket.kt
+++ b/kable-core/src/androidMain/kotlin/AndroidL2CapSocket.kt
@@ -1,0 +1,83 @@
+package com.juul.kable
+
+import android.bluetooth.BluetoothSocket
+import android.bluetooth.BluetoothSocketException
+import android.os.Build
+import com.juul.kable.logs.Logger
+import com.juul.kable.logs.Logging
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.withContext
+import kotlin.coroutines.cancellation.CancellationException
+
+internal class AndroidL2CapSocket(
+    private val socket: BluetoothSocket,
+    logging: Logging,
+) : L2CapSocket {
+
+    private val logger = Logger(logging, "Kable/L2CapSocket", socket.remoteDevice.address)
+
+    init {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+            logger.info {
+                message = "L2CAP MTU tx=${socket.maxTransmitPacketSize}, rx=${socket.maxReceivePacketSize}"
+            }
+        }
+    }
+
+    private val inputStream = socket.inputStream
+    private val outputStream = socket.outputStream
+
+    private val _isConnected = MutableStateFlow(true)
+    override val isConnected: StateFlow<Boolean> = _isConnected.asStateFlow()
+
+    private val _hasReachedEof = MutableStateFlow(false)
+    override val hasReachedEof: StateFlow<Boolean> = _hasReachedEof.asStateFlow()
+
+    override suspend fun read(buffer: ByteArray): Int {
+        try {
+            val count = withContext(Dispatchers.IO) {
+                inputStream.read(buffer)
+            }
+            if (count < 0) {
+                _hasReachedEof.value = true
+                _isConnected.value = false
+            }
+            return count
+        } catch (e: CancellationException) {
+            throw e
+        } catch (e: Exception) {
+            _isConnected.value = false
+            logger.error(e) { message = "Failed to read bytes" }
+            throw e.toL2CapException()
+        }
+    }
+
+    override suspend fun write(packet: ByteArray) {
+        try {
+            withContext(Dispatchers.IO) {
+                outputStream.write(packet)
+            }
+        } catch (e: CancellationException) {
+            throw e
+        } catch (e: Exception) {
+            _isConnected.value = false
+            logger.error(e) { message = "Failed to write packet" }
+            throw e.toL2CapException()
+        }
+    }
+
+    override suspend fun close() {
+        _isConnected.value = false
+        socket.close()
+    }
+}
+
+internal fun Exception.toL2CapException(): L2CapException =
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE && this is BluetoothSocketException) {
+        L2CapException(message, this, errorCode.toLong())
+    } else {
+        L2CapException(message, this, 0)
+    }
--- a/kable-core/src/androidMain/kotlin/AndroidL2CapSocket.kt
+++ b/kable-core/src/androidMain/kotlin/AndroidL2CapSocket.kt
@@ -6,25 +6,24 @@ import android.os.Build
 import com.juul.kable.logs.Logger
 import com.juul.kable.logs.Logging
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.NonCancellable
-import kotlinx.coroutines.SupervisorJob
-import kotlinx.coroutines.async
-import kotlinx.coroutines.cancel
-import kotlinx.coroutines.currentCoroutineContext
-import kotlinx.coroutines.ensureActive
+import kotlinx.coroutines.cancelAndJoin
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
-import kotlinx.coroutines.job
+import kotlinx.coroutines.flow.receiveAsFlow
+import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
-import kotlin.coroutines.cancellation.CancellationException
+import java.io.IOException
 
 private const val READ_CHUNK_SIZE = 8192
 
 internal class AndroidL2CapSocket(
     private val socket: BluetoothSocket,
+    scope: CoroutineScope,
     logging: Logging,
 ) : L2CapSocket {
 
@@ -38,90 +37,59 @@ internal class AndroidL2CapSocket(
         }
     }
 
-    private val inputStream = socket.inputStream
-    private val outputStream = socket.outputStream
-
     private val _isConnected = MutableStateFlow(true)
     override val isConnected: StateFlow<Boolean> = _isConnected.asStateFlow()
 
-    // Lets the read path report the exception from a deliberate socket.close() as end-of-stream.
     @Volatile
-    private var closeRequested = false
+    private var closed = false
 
-    private val scope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
+    // Unbuffered, so the socket is only read as fast as `incoming` is collected.
+    private val chunks = Channel<ByteArray>()
+    override val incoming: Flow<ByteArray> = chunks.receiveAsFlow()
 
-    // The in-flight blocking read. It survives a cancelled read() — data the OS may already have handed
-    // over must not be lost — so the next read() awaits it instead of starting another. Only touched by
-    // read(), which the interface limits to a single coroutine at a time.
-    private var pending: Deferred<ByteArray?>? = null
-
-    override suspend fun read(): ByteArray? {
-        val chunk = pending ?: scope.async { readChunk() }.also { pending = it }
+    private val reader = scope.launch(Dispatchers.IO) {
+        val buffer = ByteArray(READ_CHUNK_SIZE)
         try {
-            val result = chunk.await()
-            // End-of-stream and failure are terminal: they stay in pending so every later read()
-            // observes the same outcome instead of starting another read on a dead socket.
-            if (result != null) pending = null
-            return result
-        } catch (e: CancellationException) {
-            // Rethrows when read() itself was cancelled (pending is kept for the next read());
-            // otherwise close() cancelled the scope, which is end-of-stream.
-            currentCoroutineContext().ensureActive()
-            return null
-        }
-    }
-
-    // Blocking read into a socket-owned array, so a cancelled read() never leaves the OS writing into a
-    // caller's buffer. Terminal state is recorded here so it is observed even when no read() is awaiting.
-    private fun readChunk(): ByteArray? {
-        val scratch = ByteArray(READ_CHUNK_SIZE)
-        val count = try {
-            inputStream.read(scratch)
-        } catch (e: Exception) {
+            while (true) {
+                val count = socket.inputStream.read(buffer)
+                if (count < 0) break
+                chunks.send(buffer.copyOf(count))
+            }
+        } catch (e: IOException) {
+            if (!closed) chunks.close(e.wrapInL2CapException())
+        } finally {
             _isConnected.value = false
-            if (closeRequested) return null
-            logger.error(e) { message = "Failed to read bytes" }
-            throw e.toL2CapException()
+            chunks.close()
         }
-        if (count < 0 || closeRequested) {
-            _isConnected.value = false
-            return null
-        }
-        return if (count == scratch.size) scratch else scratch.copyOf(count)
     }
 
     override suspend fun write(packet: ByteArray) {
-        if (packet.isEmpty()) return
+        require(packet.isNotEmpty()) { "Packet must not be empty" }
         try {
             withContext(Dispatchers.IO) {
-                outputStream.write(packet)
+                socket.outputStream.write(packet)
             }
-        } catch (e: CancellationException) {
-            throw e
-        } catch (e: Exception) {
+        } catch (e: IOException) {
             _isConnected.value = false
-            logger.error(e) { message = "Failed to write packet" }
-            throw e.toL2CapException()
+            throw e.wrapInL2CapException()
         }
     }
 
     override suspend fun close() {
-        closeRequested = true
+        closed = true
         _isConnected.value = false
-        scope.cancel() // Completes any awaiting read() with null, even if socket.close() below throws.
         withContext(NonCancellable + Dispatchers.IO) {
-            try {
-                socket.close() // Unblocks an in-flight readChunk.
-            } finally {
-                scope.coroutineContext.job.join()
-            }
+            socket.close()
+            reader.cancelAndJoin()
         }
     }
 }
 
-internal fun Exception.toL2CapException(): L2CapException =
-    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE && this is BluetoothSocketException) {
-        L2CapException(message, this, errorCode.toLong())
+internal fun IOException.wrapInL2CapException(): L2CapException {
+    val code = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE && this is BluetoothSocketException) {
+        errorCode.toLong()
     } else {
-        L2CapException(message, this, 0)
+        null
     }
+    return L2CapException(message, this, code)
+}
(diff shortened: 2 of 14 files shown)


kable-core/src/androidMain/kotlin/BluetoothDeviceAndroidPeripheral.kt
```diff
@@ -192,6 +196,20 @@ internal class BluetoothDeviceAndroidPeripheral(
         return connectionOrThrow().requestMtu(mtu)
     }
 
+    @RequiresApi(Build.VERSION_CODES.Q)
+    override suspend fun openL2CapChannel(psm: Int): L2CapSocket {
+        val socket = bluetoothDevice.createL2capChannel(psm)
+        withContext(Dispatchers.IO) { socket.connect() }
+        return AndroidL2CapSocket(socket, logging)
+    }
```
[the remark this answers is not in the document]
torfinnberset: Adressed in [974c8d9](https://github.com/JuulLabs/kable/commit/974c8d9486254fd1470b6c715f48b5c88f2799e9)

torfinnberset:
@cursor review

twyatt:
Thanks for the PR! I'll try to find time to review this soon.

torfinnberset:
Please  let me know if there's anything I can do to help move this forward, either by refactoring, providing more documentation, or helping you with other blocking tickets :) Thanks!

kable-core/src/androidMain/kotlin/AndroidL2CapSocket.kt
```diff
@@ -0,0 +1,83 @@
...
+    private val inputStream = socket.inputStream
+    private val outputStream = socket.outputStream
+
+    private val _isConnected = MutableStateFlow(true)
+    override val isConnected: StateFlow<Boolean> = _isConnected.asStateFlow()
+
+    private val _hasReachedEof = MutableStateFlow(false)
+    override val hasReachedEof: StateFlow<Boolean> = _hasReachedEof.asStateFlow()
+
+    override suspend fun read(buffer: ByteArray): Int {
```
twyatt: Passing in a buffer feels like a footgun for consumers. If a consumer modifies the buffer while this operation in in-flight then we'll either corrupt the data or hit concurrent I/O failures.

We'd either need to document these risks, or workaround them entirely. Protecting against the concurrency will likely incur object creation overhead, but I feel that is worth it for providing a less error prone API?

twyatt (changes requested):
Did a quick pass; I'm a little worried about how complex the implementation is.
#1023 provides similar functionality but seems far simpler. What is this implementation providing that #1023 doesn't offer?

Is there a reason a `Flow` based input of data wasn't done?

I'd definitely like to brainstorm use-cases and what a clean/friendly API might look like.

kable-core/src/androidMain/kotlin/AndroidL2CapSocket.kt
torfinnberset: I agree, I will try to simplify the API for the consumers, thanks for taking the time to review this

torfinnberset:
Thank you again for providing your feedback, I took another pass at the implementation based on it. I'm open to further changes or discussions if that helps.

On read(buffer): Yes this makes a lot of sense in retrospect. On Android a cancelled read could still write into your buffer afterwards. I've changed read() to return a new ByteArray each time, and null at EOF, so hasReachedEof is gone too.

On Flow: I added incoming(): Flow<ByteArray> as a convenience function. I didn't want a flow as the base API because it either drops data when the consumer is slow (I had this during my testing) or it buffers forever. With a plain read() a slow consumer just slows the sender down, which I think is acceptable. 

On #1023: This PR is  made by my colleague by the way :) The API is basically the same, the difference is mostly in the iOS implementation, which was not considered in the other PR. We ran into a few things there that needed handling, such as stream calls fail intermittently unless they stay on one thread, the PSM stays connected until the CBL2CAPChannel is deallocated (so close() has to force that), and open callbacks can't be matched to requests. That's where the extra code comes from.

I added a bunch more tests locally to test this, but I wasn't sure if you wanted something to cover every edge case, or just something to cover the majority of the code.

kable-core/src/commonMain/kotlin/L2CapException.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import kotlinx.io.IOException
+
+/** Represents a failure while opening or communicating over an [L2CapSocket]. */
+public class L2CapException(
+    message: String? = null,
+    cause: Throwable? = null,
+    /**
+     * Platform error code associated with the failure: `BluetoothSocketException.errorCode` on Android
+     * (API 34+), `NSError.code` on Apple platforms, or `0` when no code is available.
```
twyatt: A value (`0`) is not a good indicator of absence. `null` would be better.

kable-core/src/appleTest/kotlin/AppleL2CapSocketTests.kt
```diff
@@ -0,0 +1,155 @@
...
+                sender.join()
+            }
+            assertContentEquals(sent, received)
+        } finally {
+            socket.close()
+        }
+    }
+
+    @Test
+    fun read_afterEndOfStream_returnsNullAndStaysNull() = runBlocking {
```
twyatt:
```suggestion
    fun read_afterEndOfStream_returnsNullAndStaysNull() = runTest {
```

kable-core/src/androidMain/kotlin/AndroidL2CapSocket.kt
```diff
@@ -0,0 +1,127 @@
...
+    private val outputStream = socket.outputStream
+
+    private val _isConnected = MutableStateFlow(true)
+    override val isConnected: StateFlow<Boolean> = _isConnected.asStateFlow()
+
+    // Lets the read path report the exception from a deliberate socket.close() as end-of-stream.
+    @Volatile
+    private var closeRequested = false
+
+    private val scope = CoroutineScope(SupervisorJob() + Dispatchers.IO)
```
twyatt: The fact that this socket isn't coupled to its associated `Peripheral` feels like we're violating structured concurrency.

kable-core/src/commonMain/kotlin/L2CapSocket.kt
```diff
@@ -0,0 +1,63 @@
...
+    public suspend fun close()
+}
+
+/**
+ * The channel's incoming bytes as a cold [Flow] of chunks, completing at end-of-stream. Collection
+ * calls [read], so collect from at most one coroutine and do not call [read] while collecting.
+ */
+public fun L2CapSocket.incoming(): Flow<ByteArray> = flow {
+    while (true) emit(read() ?: break)
+}
```
twyatt: I'm not a fan of creating a new `Flow` on every invocation, we should be using the same flow, which can be resubscribed by multiple consumers.

kable-core/src/commonMain/kotlin/L2CapSocket.kt
```diff
@@ -0,0 +1,63 @@
...
+    public suspend fun write(packet: ByteArray)
+
+    /** Closes the channel and releases its resources, suspending until teardown is complete. */
+    @Throws(CancellationException::class, IOException::class)
+    public suspend fun close()
+}
+
+/**
+ * The channel's incoming bytes as a cold [Flow] of chunks, completing at end-of-stream. Collection
+ * calls [read], so collect from at most one coroutine and do not call [read] while collecting.
```
twyatt: Having a contract of not being allowed to call another function while a Flow is active feels like a footgun.
We should pick one based on a balance of ease of use and most common use-case.

kable-core/src/commonMain/kotlin/L2CapSocket.kt
```diff
@@ -0,0 +1,63 @@
...
+     *
+     * @return the next chunk (at least one byte), or `null` once end-of-stream has been reached.
+     * @throws L2CapException if the channel fails while reading.
+     */
+    @Throws(CancellationException::class, IOException::class)
+    public suspend fun read(): ByteArray?
+
+    /**
+     * Writes the entirety of [packet] to the channel, suspending until it has all been handed off to the
+     * operating system. A no-op if [packet] is empty.
```
twyatt: We should `throw` with `IllegalArgumentException` rather than no-op.

kable-core/src/appleMain/kotlin/CoreBluetoothPeripheral.kt
```diff
@@ -31,4 +31,19 @@ public interface CoreBluetoothPeripheral : Peripheral {
...
     public suspend fun readAsNSData(characteristic: Characteristic): NSData
+
+    /**
+     * Opens an L2CAP channel to the peripheral on the given [psm], suspending until CoreBluetooth
+     * reports the channel open. The returned [L2CapSocket] is already connected.
+     *
+     * The operating system may briefly report the [psm] as still connected ("L2CAP PSM already
+     * connected") when reopening shortly after a previous socket was closed — the disconnect handshake
+     * is asynchronous — or after a cancelled open. Callers reopening the same [psm] should be prepared
+     * to retry.
```
twyatt: This comment is a little unclear to me. Where is this reported? To Logcat, via exception?

kable-core/src/androidMain/kotlin/BluetoothDeviceAndroidPeripheral.kt
```diff
@@ -192,6 +198,46 @@ internal class BluetoothDeviceAndroidPeripheral(
...
+            throw e
+        } catch (e: IOException) {
+            socket.closeOrLog()
+            throw e.toL2CapException()
+        }
+        return AndroidL2CapSocket(socket, logging)
+    }
+
+    // BluetoothSocket.close() aborts an in-progress connect, so a failed or cancelled open never leaks
+    // the socket (connect() is a blocking call that a cancelled coroutine cannot interrupt).
```
twyatt: What is this talking about:
> connect() is a blocking call...

`Peripheral.connect` is not blocking. 😕

kable-core/src/androidMain/kotlin/AndroidL2CapSocket.kt
```diff
@@ -0,0 +1,127 @@
...
+            try {
+                socket.close() // Unblocks an in-flight readChunk.
+            } finally {
+                scope.coroutineContext.job.join()
+            }
+        }
+    }
+}
+
+internal fun Exception.toL2CapException(): L2CapException =
```
twyatt: This function is misleading. The `Exception` is not being converted to another `Exception` (as the name implies), it is setting the receiver as the `cause`.

kable-core/src/androidMain/kotlin/AndroidL2CapSocket.kt
```diff
@@ -0,0 +1,127 @@
...
+            _isConnected.value = false
+            logger.error(e) { message = "Failed to write packet" }
+            throw e.toL2CapException()
+        }
+    }
+
+    override suspend fun close() {
+        closeRequested = true
+        _isConnected.value = false
+        scope.cancel() // Completes any awaiting read() with null, even if socket.close() below throws.
```
twyatt: I don't follow this comment. Probably just poorly placed? It isn't clear from the code in this function how this will cause a `null` completion.

twyatt (changes requested):
I think we're getting there, but maybe we should tone down the code comments? I generally believe that comments should be used to explain _why_ something is being done (i.e. explaining an unexpected or abnormal process) but code otherwise should try to be as self-documenting as possible and need very few comments if clearly written.

kable-core/src/commonMain/kotlin/L2CapException.kt
torfinnberset: Updated!

kable-core/src/androidMain/kotlin/AndroidL2CapSocket.kt
torfinnberset: I took another pass at this, the reader now runs in `Peripheral.scope`, so that the socket is torn down with the peripheral itself. On Apple the stream thread ends when CoreBluetooth closes the channel.

kable-core/src/commonMain/kotlin/L2CapSocket.kt
torfinnberset: It's my first time using `Flow`, so I appreciate the feedback. I replaced it with one incoming property backed by a single channel.

kable-core/src/commonMain/kotlin/L2CapSocket.kt
torfinnberset: Per the previous comment, `incoming` is the only read and there is no rules about concurrent calls anymore.

kable-core/src/commonMain/kotlin/L2CapSocket.kt
torfinnberset: Done

kable-core/src/appleMain/kotlin/CoreBluetoothPeripheral.kt
torfinnberset: Updated comment, it fails with an L2CapException the caller can retry.

kable-core/src/androidMain/kotlin/BluetoothDeviceAndroidPeripheral.kt
torfinnberset: I meant `BluetoothSocket.connect` is blocking, but it's perhaps just clearer without the comment, so I removed it 😅

kable-core/src/androidMain/kotlin/AndroidL2CapSocket.kt
torfinnberset: Renamed to `wrapInL2CapException` since it sets the receiver as the cause, is that clearer?

kable-core/src/androidMain/kotlin/AndroidL2CapSocket.kt
torfinnberset: This code is gone with the latest refactor

torfinnberset:
Thanks for the detailed feedback, I think we got to a much better place now, but let me know what you think!

torfinnberset:
@cursor review

kable-core/src/androidMain/kotlin/AndroidL2CapSocket.kt
```diff
@@ -0,0 +1,95 @@
...
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.receiveAsFlow
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import java.io.IOException
+
+private const val READ_CHUNK_SIZE = 8192
```
twyatt, line 22: How/why was this value chosen?

torfinnberset:
@twyatt please let me know if this is the direction you were thinking of, otherwise I'm still happy to iterate on this until we settle on the desired API

twyatt:
> @twyatt please let me know if this is the direction you were thinking of, otherwise I'm still happy to iterate on this until we settle on the desired API

Apologies, its been challenging to keep up on PRs with the flood of AI code (both on internal and external projects). I will try to get back to reviewing this again soon. Thank you for your patience.

kable-core/src/androidMain/kotlin/AndroidL2CapSocket.kt
```diff
@@ -0,0 +1,95 @@
...
+) : L2CapSocket {
+
+    private val logger = Logger(logging, "Kable/L2CapSocket", socket.remoteDevice.address)
+
+    init {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+            logger.info {
+                message = "L2CAP MTU tx=${socket.maxTransmitPacketSize}, rx=${socket.maxReceivePacketSize}"
+            }
+        }
```
twyatt, line 37: I'm not super familiar with L2CAP, is there a reason we'd only log the MTU only on certain Android versions?

kable-core/src/androidMain/kotlin/AndroidL2CapSocket.kt
```diff
@@ -0,0 +1,95 @@
...
+        try {
+            while (true) {
+                val count = socket.inputStream.read(buffer)
+                if (count < 0) break
+                chunks.send(buffer.copyOf(count))
+            }
+        } catch (e: IOException) {
+            if (!closed) chunks.close(e.wrapInL2CapException())
+        } finally {
+            _isConnected.value = false
```
twyatt, line 61: Despite adding a bit of complexity, we should probably monitor and intent to get instant feedback about connection state. This will allow for the state to change even if I/O operations are not being performed.

This will allow make the behavior of Android and iOS more consistent.

AI says we can do something like:

```kotlin
val filter = IntentFilter(BluetoothDevice.ACTION_ACL_DISCONNECTED)
context.registerReceiver(object : BroadcastReceiver() {
    override fun onReceive(context: Context, intent: Intent) {
        val device: BluetoothDevice? = intent.getParcelableExtra(BluetoothDevice.EXTRA_DEVICE)
        if (device?.address == connectedDeviceAddress) {
            // todo: update connected state
        }
    }
}, filter)
```

kable-core/src/androidMain/kotlin/AndroidL2CapSocket.kt
```diff
@@ -0,0 +1,95 @@
...
+        require(packet.isNotEmpty()) { "Packet must not be empty" }
+        try {
+            withContext(Dispatchers.IO) {
+                socket.outputStream.write(packet)
+            }
+        } catch (e: IOException) {
+            _isConnected.value = false
+            throw e.wrapInL2CapException()
+        }
+    }
```
twyatt, line 76: We need some kind of protection to ensure concurrent writes don't interleave the data being written.

README.md
```diff
@@ -527,6 +527,43 @@ When a write type is not specified, [`WithoutResponse`] is used.
...
+In addition to GATT, a connected peripheral can open an L2CAP connection-oriented channel (CoC) on a PSM
+(protocol/service multiplexer). Unlike GATT's message-based characteristics, a channel is a bidirectional
+byte stream, so callers are responsible for framing their own protocol. L2CAP is available on Android and
+Apple platforms (it has no Web Bluetooth equivalent), so a channel is opened from the platform-specific
+peripheral.
+
+On Android (requires API level 29 or higher):
+
+```kotlin
+val socket = (peripheral as AndroidPeripheral).openL2CapChannel(psm = 0x0080)
```
twyatt, line 541: We should expose the API in a way that does not require casting. That may mean providing no-op stubs on platforms that don't support L2CAP.

README.md
```diff
@@ -527,6 +527,43 @@ When a write type is not specified, [`WithoutResponse`] is used.
...
+end of stream, and packets are written whole:
+
+```kotlin
+socket.incoming.collect { chunk ->
+    // Process chunk.
+}
+
+socket.write(byteArrayOf(1, 2, 3))
+
+socket.close()
```
twyatt, line 561: In this example, `collect` is non-terminal, so the `write` and `close` are unreachable.

kable-core/src/appleMain/kotlin/CBPeripheralCoreBluetoothPeripheral.kt
```diff
@@ -392,6 +392,11 @@ internal class CBPeripheralCoreBluetoothPeripheral(
         cbPeripheral.identifier.UUIDString,
     )
 
+    override suspend fun openL2CapChannel(psm: Int): L2CapSocket {
+        require(psm in 0..UShort.MAX_VALUE.toInt()) { "psm $psm is outside the valid range 0..65535" }
```
twyatt, line 396:
```suggestion
        require(psm in 0..UShort.MAX_VALUE.toInt()) { "PSM of $psm is outside the valid range 0..65535" }
```

kable-core/src/androidMain/kotlin/AndroidL2CapSocket.kt
```diff
@@ -0,0 +1,95 @@
...
+    override val isConnected: StateFlow<Boolean> = _isConnected.asStateFlow()
+
+    @Volatile
+    private var closed = false
+
+    // Unbuffered, so the socket is only read as fast as `incoming` is collected.
+    private val chunks = Channel<ByteArray>()
+    override val incoming: Flow<ByteArray> = chunks.receiveAsFlow()
+
+    private val reader = scope.launch(Dispatchers.IO) {
```
twyatt, line 50: Are we sure `Dispatchers.IO` is what we want? Is it documented somewhere that we should be using an IO thread? Or, how was this dispatcher decided on?

kable-core/src/androidMain/kotlin/AndroidL2CapSocket.kt
```diff
@@ -0,0 +1,95 @@
...
+            throw e.wrapInL2CapException()
+        }
+    }
+
+    override suspend fun close() {
+        closed = true
+        _isConnected.value = false
+        withContext(NonCancellable + Dispatchers.IO) {
+            socket.close()
+            reader.cancelAndJoin()
```
twyatt, line 83: We can leak the reader if `socker.close` throws.

kable-core/src/androidMain/kotlin/AndroidL2CapSocket.kt
```diff
@@ -0,0 +1,95 @@
...
+        } catch (e: IOException) {
+            _isConnected.value = false
+            throw e.wrapInL2CapException()
+        }
+    }
+
+    override suspend fun close() {
+        closed = true
+        _isConnected.value = false
+        withContext(NonCancellable + Dispatchers.IO) {
```
twyatt, line 81: Why `NonCancellable`?

kable-core/src/androidMain/kotlin/AndroidL2CapSocket.kt
```diff
@@ -0,0 +1,95 @@
...
+                if (count < 0) break
+                chunks.send(buffer.copyOf(count))
+            }
+        } catch (e: IOException) {
+            if (!closed) chunks.close(e.wrapInL2CapException())
+        } finally {
+            _isConnected.value = false
+            chunks.close()
+        }
+    }
```
twyatt, line 64: If reading completes should we not close the socket? Otherwise it will require consumer to close it explicitly; or is there a reason that would be desired?

kable-core/src/androidMain/kotlin/AndroidL2CapSocket.kt
```diff
@@ -0,0 +1,95 @@
...
+    private val reader = scope.launch(Dispatchers.IO) {
+        val buffer = ByteArray(READ_CHUNK_SIZE)
+        try {
+            while (true) {
+                val count = socket.inputStream.read(buffer)
+                if (count < 0) break
+                chunks.send(buffer.copyOf(count))
+            }
+        } catch (e: IOException) {
+            if (!closed) chunks.close(e.wrapInL2CapException())
```
twyatt, line 59: What happens if a non-`IOException` is thrown? (how will it be propagated)?

kable-core/src/appleMain/kotlin/Connection.kt
```diff
@@ -169,6 +171,61 @@ internal class Connection(
...
+                delegate.cancelL2CapOpen(pending, e.unwrapCancellationException())
+            }
+            throw e.unwrapCancellationException()
+        } catch (e: Throwable) {
+            delegate.cancelL2CapOpen(pending, e)
+            throw e
+        }
+
+        val result = try {
+            withContext(NonCancellable) { pending.await() }
```
twyatt, line 198: Why `NonCancellable`?

kable-core/src/appleMain/kotlin/AppleL2CapSocket.kt
```diff
@@ -0,0 +1,310 @@
...
+    private var retainedChannel: CBL2CAPChannel? = channel
+
+    private val inputStream: NSInputStream = channel.inputStream ?: error("L2CAP channel has no input stream")
+    private val outputStream: NSOutputStream = channel.outputStream ?: error("L2CAP channel has no output stream")
+    private val streamDelegate = StreamDelegate()
+
+    private val _isConnected = MutableStateFlow(true)
+    override val isConnected: StateFlow<Boolean> = _isConnected.asStateFlow()
+
+    private val chunks = Channel<ByteArray>(Channel.UNLIMITED)
```
twyatt, line 63: On Android a rendezvous channel is used (allowing for backpressure) but here we're using an unlimited buffer. This would make the behavior between the two platforms inconsistent.

kable-core/src/commonMain/kotlin/L2CapException.kt
```diff
@@ -0,0 +1,14 @@
...
+
+/** A failure while opening or communicating over an [L2CapSocket]. */
+public class L2CapException(
+    message: String? = null,
+    cause: Throwable? = null,
+    /**
+     * Platform error code, if any: `BluetoothSocketException.errorCode` on Android (API 34+) or
+     * `NSError.code` on Apple platforms.
+     */
+    public val code: Long? = null,
```
twyatt, line 13: For consumers (in `commonMain`) this `code` can have overlapping values from the two platforms that mean different things. This should be commonized (for example, could be mapped to an error `enum`).

kable-core/src/commonMain/kotlin/L2CapSocket.kt
```diff
@@ -0,0 +1,38 @@
...
+public interface L2CapSocket {
+
+    /** Whether the channel is open. Becomes `false` once [closed][close], at end of stream, or on failure. */
+    public val isConnected: StateFlow<Boolean>
+
+    /**
+     * Bytes received over the channel, in chunks of arbitrary size. Completes at end of stream and
+     * throws [L2CapException] if the channel fails. Each chunk is delivered to a single collector.
+     */
+    public val incoming: Flow<ByteArray>
```
twyatt, line 24: This should be a `SharedFlow`, otherwise multiple consumers could each get different chunks (rather than all consumers getting every chunk).

twyatt (changes requested):

torfinnberset:
Thanks for your comments, I'll try to get back to you later this week!

Issue #485, opened 2023-05-20
Android ANR when collecting advertisements


miwright2:
My Android app is in production and uses kable.  It's working great for us.  But I'm getting ANRs from Crashlytics that backtrace to [this line](https://github.com/JuulLabs/kable/blob/2326f416a81fd0af07ac26b45a16276c8013168e/core/src/androidMain/kotlin/Scanner.kt#L49) in kable.

One of the most interesting parts is that the issue effects only Samsung users.  I have ~244 events over 94 users in the last 90 days.

My code essentially looks like this...

```
allMatchesScanner
    .advertisements
    .buffer(capacity = 5, onBufferOverflow = BufferOverflow.DROP_OLDEST)
    .flowOn(Dispatchers.IO)
    .catch { 
        stop()
    }
    .onCompletion {
        Log.d(LOG_TAG, "Scanning stopped ...")
    }
    .collect { advertisement ->
        // process advertising packet
    }
```

And this is the backtrace...

```
[inc.combustion.app_issue_4ef85018dda05154ffababd15f0888f2_ANR_session_6466959903A400015B7E1A496343B900_DNE_0_v2_stacktrace.txt](https://github.com/JuulLabs/kable/files/11521287/inc.combustion.app_issue_4ef85018dda05154ffababd15f0888f2_ANR_session_6466959903A400015B7E1A496343B900_DNE_0_v2_stacktrace.txt)

main (timed waiting):tid=1 systid=23422 
       at jdk.internal.misc.Unsafe.park(Native method)
       at java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:234)
       at kotlinx.coroutines.BlockingCoroutine.joinBlocking(Builders.kt:88)
       at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking(Builders.kt:59)
       at kotlinx.coroutines.BuildersKt.runBlocking(unavailable:1)
       at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking$default(Builders.kt:38)
       at kotlinx.coroutines.BuildersKt.runBlocking$default(unavailable:1)
       at kotlinx.coroutines.channels.ChannelsKt__ChannelsKt.trySendBlocking(Channels.kt:38)
       at kotlinx.coroutines.channels.ChannelsKt.trySendBlocking(unavailable:1)
       at com.juul.kable.AndroidScanner$advertisements$1$callback$1.onScanResult(Scanner.kt:48)
[...]

PR #1253, opened 2026-08-12
Never block scan callback threads when delivering scan results

### LINKED ISSUE #654: Android ANR when collecting advertisements
Still trying to figure out this one. I've been running into an ANR that I seem to consistently hit after anywhere from 15-45 minutes of successful scanning. I've been testing on an Galaxy S22 running Android 13. 

For context, I have an ongoing BLE scan running in a foreground service to connect to certain devices as they appear. This scan is meant to run alongside the app, indefinitely. I have only observed this when I leave the phone unplugged with the screen off for a while (with the foreground service still running in the background).

The ANR is as follows:

```Cmd line: <package>
Build fingerprint: 'samsung/b0qsqw/b0q:13/TP1A.220624.014/S908USQS2BWA2:user/release-keys'
ABI: 'arm64'
Build type: optimized
suspend all histogram:	Sum: 41.270ms 99% C.I. 0.494us-948.800us Avg: 78.311us Max: 2379us
DALVIK THREADS (168):
"main" prio=5 tid=1 TimedWaiting
  | group="main" sCount=1 ucsCount=0 flags=1 obj=0x7334ef58 self=0xb4000079a6fc6380
  | sysTid=6202 nice=0 cgrp=background sched=0/0 handle=0x7b521c3500
  | state=S schedstat=( 427635411952 12382667868 187478 ) utm=34641 stm=8121 core=2 HZ=100
  | stack=0x7fceb7f000-0x7fceb81000 stackSize=8188KB
  | held mutexes=
  at jdk.internal.misc.Unsafe.park(Native method)
  - waiting on an unknown object
  at java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:252)
  at kotlinx.coroutines.BlockingCoroutine.joinBlocking(Builders.kt:88)
  at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking(Builders.kt:59)
  at kotlinx.coroutines.BuildersKt.runBlocking(unavailable:1)
  at kotlinx.coroutines.BuildersKt__BuildersKt.runBlocking$default(Builders.kt:38)
  at kotlinx.coroutines.BuildersKt.runBlocking$default(unavailable:1)
  at kotlinx.coroutines.channels.ChannelsKt__ChannelsKt.trySendBlocking(Channels.kt:38)
  at kotlinx.coroutines.channels.ChannelsKt.trySendBlocking(unavailable:1)
  at com.juul.kable.BluetoothLeScannerAndroidScanner$advertisements$1$callback$1.onBatchScanResults(BluetoothLeScannerAndroidScanner.kt:54)
  at android.bluetooth.le.BluetoothLeScanner$BleScanCallbackWrapper$2.run(BluetoothLeScanner.java:681)
  ...
```

This can be followed to this line in `BluetoothLeScannerAndroidScanner.kt:54`:

https://github.com/JuulLabs/kable/blob/b69d0ff84e0a5e1075c695d35bb7de32f364280d/core/src/androidMain/kotlin/BluetoothLeScannerAndroidScanner.kt#L54

jamesarich:
## Problem

`BluetoothLeScannerAndroidScanner` delivers scan results to the `advertisements` `callbackFlow` via `trySendBlocking` (unless `preConflate` is enabled). Android invokes `ScanCallback` functions from a binder thread — and on some phones, the main thread. When scan results arrive faster than the collector drains the channel (dense BLE environments), the callback thread parks until buffer space frees up. When that thread is the main thread, the OS declares an ANR.

This is the same failure mode previously reported in #485 and #654. We (Meshtastic Android, Kable 0.44.3) hit it in production: 8 ANRs across distinct users during DEF CON 34 (an extremely dense BLE environment), all with the main thread parked in:

```
kotlinx.coroutines.channels.ChannelsKt__ChannelsKt.trySendBlocking
com.juul.kable.BluetoothLeScannerAndroidScanner$advertisements$1$callback$1.onScanResult
android.bluetooth.le.BluetoothLeScanner$BleScanCallbackWrapper$1.run
```

## Change

- Use non-blocking `trySend` unconditionally in the scan callback.
- Add a `bufferCapacity` property to `ScannerBuilder` (Android), applied as `buffer(capacity, DROP_OLDEST)` on the `advertisements` `callbackFlow`:
  - `UNLIMITED` (the default) is **lossless**: `trySend` cannot fail on an unlimited channel (except after close), so no scan results are dropped - the previous blocking behavior's guarantee is preserved without ever parking the callback thread.
  - Any capacity of at least `1` drops the oldest buffered result to make room for the newest; `1` is equivalent to `conflate`.
  - `BUFFERED`, `CONFLATED` and `0` are rejected, since none of them keep their usual meaning alongside `DROP_OLDEST` (`BUFFERED` becomes a capacity of 1, not the default depth).
- Deprecate `preConflate`; `preConflate = true` maps to `bufferCapacity = 1`, `false` to `UNLIMITED`.

The trade-off at the default is that a slow collector now grows the buffer (memory) instead of exerting backpressure on the binder thread. Backpressure via blocking an Android-owned callback thread was never safe - the "pressure" propagated to the OS as an ANR rather than to the producer, since the producer (the Bluetooth stack) doesn't slow down anyway.

The public API change is additive (`bufferCapacity` getter/setter); the Android API dump is updated.

Closes #654 (the default-behavior remnant of it, at least).

---

*Disclosure: this change was drafted with an AI assistant (Claude Code), then reviewed by me. It is motivated by first-party production ANR reports from [Meshtastic-Android](https://github.com/meshtastic/Meshtastic-Android). Verified locally with `lintKotlin`, `apiCheck`, and `testAndroidHostTest`.*

---

> [!NOTE]
> **Medium Risk**
> Touches the hot-path Android scan callback pipeline and shifts backpressure from thread blocking to unbounded buffering; behavior change is intentional but worth validating under heavy BLE load.
> 
> **Overview**
> Fixes **ANRs** when BLE scan results outpace collectors by stopping use of **`trySendBlocking`** on Android `ScanCallback` threads (sometimes the main thread).
> 
> Scan delivery now always uses non-blocking **`trySend`**, with a **`buffer`** on the advertisements flow: **`UNLIMITED`** by default so results stay lossless without parking the callback thread; **`BUFFERED`** when **`preConflate`** is enabled (drops when full, same as before). **`preConflate`** docs are updated to describe buffering/drops instead of blocking/`runBlocking`.
> 
> Trade-off: a slow collector can grow memory instead of blocking the binder thread.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 73d9299c9b89695de74ea7173a755b5824d1b4ba. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/kable-core/src/androidHostTest/kotlin/com/juul/kable/ScannerBuilderTests.kt
+++ b/kable-core/src/androidHostTest/kotlin/com/juul/kable/ScannerBuilderTests.kt
@@ -1,0 +1,71 @@
+package com.juul.kable
+
+import android.os.Build
+import kotlinx.coroutines.channels.Channel
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+@RunWith(RobolectricTestRunner::class)
+@Config(sdk = [Build.VERSION_CODES.S])
+class ScannerBuilderTests {
+
+    @Test
+    fun bufferCapacity_default_isUnlimited() {
+        assertEquals(Channel.UNLIMITED, ScannerBuilder().bufferCapacity)
+    }
+
+    @Test
+    fun bufferCapacity_positive_isRetained() {
+        val builder = ScannerBuilder()
+        builder.bufferCapacity = 64
+        assertEquals(64, builder.bufferCapacity)
+    }
+
+    @Test
+    fun bufferCapacity_buffered_isRejected() {
+        // `Channel.BUFFERED` behaves as a capacity of 1 (not the default channel capacity) when
+        // combined with an overflow strategy, so it is rejected rather than silently misleading.
+        assertFailsWith<IllegalArgumentException> {
+            ScannerBuilder().bufferCapacity = Channel.BUFFERED
+        }
+    }
+
+    @Test
+    fun bufferCapacity_conflated_isRejected() {
+        assertFailsWith<IllegalArgumentException> {
+            ScannerBuilder().bufferCapacity = Channel.CONFLATED
+        }
+    }
+
+    @Test
+    fun bufferCapacity_rendezvous_isRejected() {
+        assertFailsWith<IllegalArgumentException> {
+            ScannerBuilder().bufferCapacity = 0
+        }
+    }
+
+    @Suppress("DEPRECATION")
+    @Test
+    fun preConflate_true_conflatesBufferCapacity() {
+        val builder = ScannerBuilder()
+        builder.preConflate = true
+        assertEquals(1, builder.bufferCapacity)
+        assertTrue(builder.preConflate)
+    }
+
+    @Suppress("DEPRECATION")
+    @Test
+    fun preConflate_false_isUnlimitedBufferCapacity() {
+        val builder = ScannerBuilder()
+        builder.preConflate = true
+        builder.preConflate = false
+        assertEquals(Channel.UNLIMITED, builder.bufferCapacity)
+        assertFalse(builder.preConflate)
+    }
+}
--- a/kable-core/src/androidMain/kotlin/BluetoothLeScannerAndroidScanner.kt
+++ b/kable-core/src/androidMain/kotlin/BluetoothLeScannerAndroidScanner.kt
@@ -20,10 +20,12 @@ import com.juul.kable.scan.message
 import com.juul.kable.scan.requirements.checkLocationServicesEnabled
 import com.juul.kable.scan.requirements.checkScanPermissions
 import com.juul.kable.scan.requirements.requireBluetoothLeScanner
+import kotlinx.coroutines.channels.Channel.Factory.BUFFERED
+import kotlinx.coroutines.channels.Channel.Factory.UNLIMITED
 import kotlinx.coroutines.channels.awaitClose
 import kotlinx.coroutines.channels.onFailure
-import kotlinx.coroutines.channels.trySendBlocking
 import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.buffer
 import kotlinx.coroutines.flow.callbackFlow
 import kotlinx.coroutines.flow.filter
 import kotlin.reflect.KClass
@@ -59,12 +61,12 @@ internal class BluetoothLeScannerAndroidScanner(
         logger.verbose { message = "Checking permissions for scanning" }
         checkScanPermissions()
 
+        // `trySend` is used (rather than `trySendBlocking`) because scan callbacks are invoked
+        // from a binder thread (on some phones, the main thread), where blocking can trigger an
+        // ANR. See https://github.com/JuulLabs/kable/issues/654 for more details.
         fun sendResult(scanResult: ScanResult) {
             val advertisement = ScanResultAndroidAdvertisement(scanResult)
-            when {
-                preConflate -> trySend(advertisement)
-                else -> trySendBlocking(advertisement)
-            }.onFailure {
+            trySend(advertisement).onFailure {
                 logger.warn { message = "Unable to deliver scan result due to failure in flow or premature closing." }
             }
         }
@@ -115,7 +117,13 @@ internal class BluetoothLeScannerAndroidScanner(
                 logger.warn(e) { message = "Failed to stop scan. " }
             }
         }
-    }.filter { advertisement ->
+    }.buffer(
+        // When pre-conflating, the default (BUFFERED) capacity is used, whereas scan results that
+        // arrive while the buffer is full are dropped (via failed `trySend`). Otherwise, an
+        // UNLIMITED capacity ensures no scan results are dropped, while never blocking the
+        // (Android provided) thread that scan callbacks are invoked from.
+        capacity = if (preConflate) BUFFERED else UNLIMITED,
+    ).filter { advertisement ->
         if (scanFilters.flow.isEmpty()) {
             true
         } else {
--- a/kable-core/src/androidMain/kotlin/BluetoothLeScannerAndroidScanner.kt
+++ b/kable-core/src/androidMain/kotlin/BluetoothLeScannerAndroidScanner.kt
@@ -20,7 +20,7 @@ import com.juul.kable.scan.message
 import com.juul.kable.scan.requirements.checkLocationServicesEnabled
 import com.juul.kable.scan.requirements.checkScanPermissions
 import com.juul.kable.scan.requirements.requireBluetoothLeScanner
-import kotlinx.coroutines.channels.Channel.Factory.BUFFERED
+import kotlinx.coroutines.channels.BufferOverflow.DROP_OLDEST
 import kotlinx.coroutines.channels.Channel.Factory.UNLIMITED
 import kotlinx.coroutines.channels.awaitClose
 import kotlinx.coroutines.channels.onFailure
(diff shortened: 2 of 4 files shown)


twyatt:
Thanks for submitting this PR, I'll try to review soon.

kable-core/src/androidMain/kotlin/BluetoothLeScannerAndroidScanner.kt
```diff
@@ -115,7 +117,13 @@ internal class BluetoothLeScannerAndroidScanner(
                 logger.warn(e) { message = "Failed to stop scan. " }
             }
         }
-    }.filter { advertisement ->
+    }.buffer(
+        // When pre-conflating, the default (BUFFERED) capacity is used, whereas scan results that
+        // arrive while the buffer is full are dropped (via failed `trySend`). Otherwise, an
+        // UNLIMITED capacity ensures no scan results are dropped, while never blocking the
+        // (Android provided) thread that scan callbacks are invoked from.
+        capacity = if (preConflate) BUFFERED else UNLIMITED,
```
twyatt:
```suggestion
        capacity = if (preConflate) BUFFERED else UNLIMITED,
        onBufferOverflow = DROP_OLDEST,
```

twyatt (commented):
These changes look good, but I think we should allow the consumer a little more control. As some consumers may want to allow dropping of some advertisements.

Let's mark the `preConflate` builder property as deprecated, and introduce a new `bufferCapacity: Int` builder property.

jamesarich:
`BUFFERED` doesn't survive `DROP_OLDEST` - the `Channel` factory only uses the default capacity when `onBufferOverflow` is `SUSPEND`, otherwise you get `ConflatedBufferedChannel(1, ...)`. 100 elements into a channel nothing is draining, coroutines 1.11.0:

```
BUFFERED  -> 1
0         -> 1
64        -> 64
UNLIMITED -> 100
```

So `bufferCapacity` takes `UNLIMITED` (the default) or anything >= 1, and rejects `BUFFERED`, `CONFLATED` and `0` since none of them mean what they say here. Deprecated `preConflate = true` maps to `1` - that's what the name promised anyway, and 64 can't be named alongside `DROP_OLDEST` without hardcoding it. Happy to hardcode it if you'd rather the old depth stick.

One side effect - `trySend` evicts instead of failing now, so finite-capacity drops are silent. No `onUndeliveredElement` hook on `callbackFlow` to log them.

Rebased, added `ScannerBuilderTests`. @twyatt another look?

kable-core/src/androidMain/kotlin/BluetoothLeScannerAndroidScanner.kt
```diff
@@ -115,7 +117,18 @@ internal class BluetoothLeScannerAndroidScanner(
                 logger.warn(e) { message = "Failed to stop scan. " }
             }
         }
-    }.filter { advertisement ->
+    }.buffer(
+        // Scan results are delivered via (non-blocking) `trySend`, so this buffer — rather than
+        // backpressure on the thread that scan callbacks are invoked from — is what absorbs a slow
+        // collector. `DROP_OLDEST` keeps `trySend` from failing at a finite capacity; it is
+        // ignored at an UNLIMITED capacity, which never overflows and so never drops.
+        //
```
twyatt, line 125: This describes typical semantics of the flow operator so isn't adding much value, IMHO.


```suggestion
```

kable-core/src/androidMain/kotlin/BluetoothLeScannerAndroidScanner.kt
```diff
@@ -59,12 +61,12 @@ internal class BluetoothLeScannerAndroidScanner(
         logger.verbose { message = "Checking permissions for scanning" }
         checkScanPermissions()
 
+        // `trySend` is used (rather than `trySendBlocking`) because scan callbacks are invoked
+        // from a binder thread (on some phones, the main thread), where blocking can trigger an
+        // ANR. See https://github.com/JuulLabs/kable/issues/654 for more details.
```
twyatt, line 66: Let's move this comment just above the `trySend` call.

kable-core/src/androidHostTest/kotlin/com/juul/kable/ScannerBuilderTests.kt
```diff
@@ -0,0 +1,71 @@
...
+import kotlin.test.assertTrue
+
+@RunWith(RobolectricTestRunner::class)
+@Config(sdk = [Build.VERSION_CODES.S])
+class ScannerBuilderTests {
+
+    @Test
+    fun bufferCapacity_default_isUnlimited() {
+        assertEquals(Channel.UNLIMITED, ScannerBuilder().bufferCapacity)
+    }
```
twyatt, line 21: This isn't testing any business logic so isn't adding any value.


```suggestion
```

kable-core/src/androidHostTest/kotlin/com/juul/kable/ScannerBuilderTests.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    @Test
+    fun bufferCapacity_positive_isRetained() {
+        val builder = ScannerBuilder()
+        builder.bufferCapacity = 64
+        assertEquals(64, builder.bufferCapacity)
+    }
```
twyatt, line 28: Testing variable assignment, so isn't adding any value.


```suggestion
```

kable-core/src/androidHostTest/kotlin/com/juul/kable/ScannerBuilderTests.kt
```diff
@@ -0,0 +1,71 @@
...
+        builder.preConflate = true
+        assertEquals(1, builder.bufferCapacity)
+        assertTrue(builder.preConflate)
+    }
+
+    @Suppress("DEPRECATION")
+    @Test
+    fun preConflate_false_isUnlimitedBufferCapacity() {
+        val builder = ScannerBuilder()
+        builder.preConflate = true
```
twyatt, line 66:
```suggestion
```

twyatt (commented):
Looks great! Just some very minor comments remaining.

kable-core/src/androidMain/kotlin/BluetoothLeScannerAndroidScanner.kt
jamesarich: Done. Kept the fusing paragraph, dropped the rest.

kable-core/src/androidMain/kotlin/BluetoothLeScannerAndroidScanner.kt
jamesarich: Moved above the `trySend` call.

kable-core/src/androidHostTest/kotlin/com/juul/kable/ScannerBuilderTests.kt
jamesarich: Removed.

kable-core/src/androidHostTest/kotlin/com/juul/kable/ScannerBuilderTests.kt
jamesarich: Removed.

kable-core/src/androidHostTest/kotlin/com/juul/kable/ScannerBuilderTests.kt
jamesarich: Removed.

jamesarich:
Applied the remaining comments in 32712bc. @twyatt another look?

twyatt (approved):

twyatt:
@jamesarich looks great! I'll ping the team to get another review and then merge it.

cedrickcooke (approved):

PR #1263, opened 2026-08-30
Add mingwX64 target support

xxfast:
Kable has no Windows target today. This wires up `mingwX64` by reusing the `kable-btleplug-ffi` Rust crate through a new `kable-btleplug-ffi-native` module, with Kotlin/Native cinterop bindings generated by the Ubique uniffi plugin. Went with Ubique over gobley because gobley's bindgen is pinned in lockstep with uniffi 0.29, and adopting it would have meant downgrading the crate (currently on uniffi 0.32) that the JVM desktop client already builds against. The new target is gated to Windows hosts so existing CI stays green.

- [x] Add mingwX64 target support
- [x] Generate Kotlin/Native bindings with the Ubique uniffi plugin

Still pending before this is ready for review: mingw toolchain on the macOS publish job, `rustup target add x86_64-pc-windows-gnu` on the windows-latest runner, and a release-variant staticlib config (the klib currently embeds a debug build).


kable-btleplug-ffi-native/src/mingwX64Test/kotlin/FfiSmokeTest.kt
```diff
@@ -0,0 +1,67 @@
...
+                try {
+                    delay(5.seconds)
+                } finally {
+                    handle.cancel()
+                    handle.destroy()
+                }
+                println("Advertisement callbacks received: ${updates.value}")
+            }
+        }
+    }
```
twyatt: This test is not deterministic/repeatable so isn't adding value, IMHO.

kable-btleplug-ffi-native/src/mingwX64Test/kotlin/FfiSmokeTest.kt
```diff
@@ -0,0 +1,67 @@
...
+     * on the Bluetooth hardware of the machine running the test.
+     */
+    @Test
+    fun isSupported_returnsWithoutCrashing() {
+        runBlocking {
+            withTimeout(30.seconds) {
+                println("isSupported: ${isSupported()}")
+            }
+        }
+    }
```
twyatt: This is not adding value if it doesn't perform any assertions.

kable-core/src/commonTest/kotlin/ObservationTest.kt
```diff
@@ -32,16 +28,6 @@ class ObservationTest {
...
-    @BeforeTest
-    fun setup() {
-        Log.dispatcher.install(ConsoleLogger)
-    }
-
-    @AfterTest
-    fun tearDown() {
-        Log.dispatcher.clear()
-    }
-
```
twyatt: I agree that this logging is not necessary, but definitely outside the scope of this PR.

kable-core/src/jvmMain/kotlin/com/juul/kable/IoDispatcher.kt
```diff
@@ -0,0 +1,6 @@
+package com.juul.kable
+
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Dispatchers
+
+internal val ioDispatcher: CoroutineDispatcher = Dispatchers.IO
```
twyatt: Why do we need this syntax sugar rather than using `Dispatchers.IO` directly?

kable-btleplug-ffi-native/build.gradle.kts
```diff
@@ -0,0 +1,70 @@
...
+    bindgenFromGitTag(
+        repository = "https://github.com/UbiqueInnovation/uniffi-kotlin-multiplatform-bindings.git",
+        tag = "v1.2.1",
+    )
+    generateFromLibrary()
+}
+
+// On Windows, `cargo install` produces `uniffi-bindgen-kotlin-multiplatform.exe`, but the plugin
+// resolves the installed binary without the `.exe` extension. Provide the binary under the name
+// the plugin expects. https://github.com/UbiqueInnovation/uniffi-kotlin-multiplatform-bindings/issues
```
twyatt: The link is for the issues page of that library; is there a specific issue that was intended to be linked?

kable-btleplug-ffi-native/gradle.properties
```diff
@@ -0,0 +1,3 @@
+# The generated bindings live in the shared `nativeMain` source set and reference the cinterop
+# declarations, which are only visible from a shared source set once the commonizer has run.
+kotlin.mpp.enableCInteropCommonization=true
```
twyatt: I may be wrong, but I thought this would only be needed with multiple source sets, and we only have a single source set for this module. Is this really needed?

twyatt (commented):
We have Windows CI runners, we should exercise this new module to verify that it is working.

I'm also curious, did the Windows JVM target not fit your needs/use-case?

PR #1270, opened 2026-09-09
Workaround for trailing `Unit` bug in Kotlin compiler

### LINKED ISSUE #1267: write can resume the caller's continuation with the raw Response instead of Unit
Consumers compiled with Kotlin 2.3 can get a `ClassCastException` out of `AndroidPeripheral.write`:

```
java.lang.ClassCastException: com.juul.kable.gatt.Response$OnCharacteristicWrite cannot be cast to kotlin.Unit
    at com.example.MyPeripheral.requestConnection(MyGatt.kt:183)
    at com.example.MyPeripheral$requestConnection$1.invokeSuspend(MyGatt.kt:15)
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:34)
```

### Cause

`write` is a `Unit` returning suspend fun whose last expression is `connection.execute<OnCharacteristicWrite> { ... }`. Up to and including 0.43.0 that compiles to a suspend tail call: `write` hands the caller's `$completion` straight to `execute` and generates no resume label of its own. From `kable-core-android-0.43.0`:

```
174: invokevirtual Connection.execute:(...)Ljava/lang/Object;
177: dup
178: getCOROUTINE_SUSPENDED
181: if_acmpne 185
184: areturn                      // suspended: the caller resumes with the raw Response
185: pop
186: getstatic kotlin/Unit.INSTANCE
189: areturn                      // only the non-suspending path yields Unit
```

So on the suspending path, which is every real GATT write since it waits for `onCharacteristicWrite`, the caller's continuation is resumed with `Response.OnCharacteristicWrite` rather than `Unit`.

That is harmless as long as callers discard the value. Kotlin 2.3 emits a `checkcast kotlin/Unit` on the resumed value in some caller shapes, and then it isn't:

```kotlin
suspend fun requestConnection(peripheral: Peripheral, characteristic: Characteristic, data: ByteArray) {
  try {
    peripheral.write(characteristic, data, WriteType.WithResponse)
  } catch (e: NotConnectedException) {
    logger?.error { "failed: ${e.message}" }
  }
}
```

The `?.` types that catch branch as `Unit?`, so the `try`/`catch` expression is `Unit?` and the compiler coerces it. The `checkcast` throws, 10 times out of 10. Adding any statement after the `write` inside the `try` makes it disappear, which makes it an unpleasant one to land on.

`write(descriptor, data)` has the same shape, two tail calls deep, via the private `write(BluetoothGattDescriptor, ByteArray)`.

### Versions

I checked the published bytecode:

| kable | built with (metadata `mv`) | `execute` in tail position |
| --- | --- | --- |
| 0.42.0 | 2.0 | yes |
| 0.43.0 | 2.3 | yes |
| 0.44.3 | 2.4 | no |

In 0.44.3 `write` has its own `write$1` state machine and returns `Unit.INSTANCE` on the resume path. The source is unchanged between 0.43.0 and 0.44.3, so Kotlin 2.4 appears to have fixed the codegen.

So 0.44.3 does not reproduce. Anyone on 0.43.0 or older with a Kotlin 2.3 consumer does, and the crash lands in their code with nothing in the stack pointing at kable.

### Suggestion

Worth not leaving `execute` in tail position of the `Unit` returning members, so this doesn't depend on which Kotlin version built kable.

twyatt:
Closes #1267

---

> [!NOTE]
> **Low Risk**
> Narrow Android-only change to two write helpers with no API or behavior change beyond fixing incorrect return-type propagation at compile/runtime on affected Kotlin versions.
> 
> **Overview**
> Adds a Kotlin compiler workaround on the Android peripheral **write** paths for GATT characteristics and descriptors.
> 
> After `connectionOrThrow().execute { … }`, those `suspend` methods now call a new non-inline **`discard()`** helper so the resumed value is forced to `Unit`. This addresses [KT-72710](https://youtrack.jetbrains.com/issue/KT-72710) / [#1267](https://github.com/JuulLabs/kable/issues/1267), where older compilers could incorrectly propagate a non-`Unit` result from the tail call inside `execute` to callers of functions that should return `Unit`. Read and RSSI paths are unchanged because they already consume the `execute` result explicitly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a1c8bfdd530d01bcb35c78628d6eac41447a7ddf. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


twyatt:
Bug fixed upstream, consumers should upgrade to Kable 0.44.3 or newer.

PR #1274, opened 2026-09-11
Fix JVM maximum write length using negotiated MTU

### LINKED ISSUE #1248: `maximumWriteValueLengthForType` returns a hardcoded 20 on JVM/btleplug, ignoring the negotiated MTU
#### Summary

On the JVM (btleplug) target, `Peripheral.maximumWriteValueLengthForType` always returns `20`, regardless of the
MTU actually negotiated with the peripheral. It also ignores its `writeType` argument.

This contradicts the documented contract of the function, and unlike the JavaScript target — where the same
limitation exists and *is* documented — there is nothing in the KDoc telling a caller that JVM behaves this way.

#### Environment

- Kable `0.42.0`, and unchanged on `main` as of `0.44.3` (2026-07-22)
- Target: JVM (btleplug), observed on Windows

#### Detail

`kable-core/src/jvmMain/kotlin/com/juul/kable/btleplug/BtleplugPeripheral.kt`:

```kotlin
private const val DEFAULT_ATT_MTU = 23
private const val ATT_MTU_HEADER_SIZE = 3

// ...

override suspend fun maximumWriteValueLengthForType(writeType: WriteType): Int =
    DEFAULT_ATT_MTU - ATT_MTU_HEADER_SIZE
```

The declared contract in `kable-core/src/commonMain/kotlin/Peripheral.kt` is:

> Return the current ATT MTU size, minus the size of the ATT headers (3 bytes).
>
> On Android, this will be the default (23 - 3) unless you called `requestMtu` when connecting.
> For iOS, this is automatically negotiated, and can also vary depending on the writeType.
> On JavaScript, this will return the default (23 - 3) every time as there is no ATT MTU property available.

JVM is absent from that list. Since three of the four targets are described and the JS caveat is called out
explicitly, the omission reads as "JVM behaves like Android/iOS" — which is what makes this quiet rather than
obvious.

**Expected:** either the current negotiated MTU minus 3, or a documented statement that JVM returns the default,
as JavaScript already has.

**Actual:** `20`, always.

#### Why this is fixable rather than an inherent btleplug limitation

btleplug's `Peripheral` trait carries the value already — `fn mtu(&self) -> u16`, *"Returns the currently
negotiated mtu size"* — and it is a **required** trait method as of btleplug 0.12.0, which Kable already tracks
(#1120). So this looks like surfacing one existing FFI call rather than adding capability:

```kotlin
override suspend fun maximumWriteValueLengthForType(writeType: WriteType): Int =
    ffi.mtu().toInt() - ATT_MTU_HEADER_SIZE
```

(`kable-btleplug-ffi` 0.42.0 does not expose `mtu` at all, so the FFI side needs the binding too.)

#### Repro

```kotlin
val peripheral = Peripheral(identifier)
peripheral.connect()
// Prints 20 on JVM regardless of what the OS and peripheral negotiated.
println(peripheral.maximumWriteValueLengthForType(WriteType.WithResponse))
```

#### Why it matters to us

We use `maximumWriteValueLengthForType` to decide whether to fragment outbound writes. The peripheral we talk to
prefers an MTU of 247, and the very first client write of every session is a ~132-byte authentication packet.

A hardcoded `20` is worse than no answer here: it is indistinguishable from a genuinely negotiated 20, so a caller
that trusts it fragments messages that never needed fragmenting. In our case whole writes of ~141 bytes were
observed succeeding on this path against real hardware, while splitting to the reported 20 did not work — the
reported value was not merely conservative, it was actively misleading. We currently special-case the JVM target
to treat the number as unknown and skip fragmentation entirely, which is the only safe reading of it.

This also interacts with #298 (throw when a write exceeds the MTU): implemented against the current JVM value,
that would reject every write over 20 bytes on a link that can carry far more.

#### Secondary, related: no way to request an MTU on JVM

There is no `requestMtu` on the JVM path — `ServicesDiscoveredPeripheral.requestMtu` is Android-only — so a JVM
caller can neither ask for a larger MTU nor find out what it got. Whether that belongs in this issue or in the
builder-configuration work already under way in #867 / #811 is yours to decide; happy to split it out.

#### Related

- #809 — added `maximumWriteValueLength` to `Peripheral` (iOS/Android/Web); JVM appears to have received the
  constant stub at that point
- #867 / #811 — moving MTU configuration onto the peripheral builder, the natural home for a JVM request path
- #298 — validating writes against the MTU, which depends on this value being real
- #1120 — btleplug 0.12.0 upgrade, which is what makes `mtu()` available

AvikMakwana:
On JVM, \BtleplugPeripheral.maximumWriteValueLengthForType\ previously returned a hardcoded \20\ (\23 - 3\) because the negotiated MTU was not exposed through the btleplug FFI layer.

This PR:
- Exposes the current MTU from btleplug through \kable-btleplug-ffi\.
- Updates \BtleplugPeripheral.maximumWriteValueLengthForType\ to calculate the maximum write length as the negotiated ATT MTU minus the 3-byte ATT header.
- Documents automatic JVM/Desktop MTU negotiation in \Peripheral.maximumWriteValueLengthForType\.

Fixes #1248

### Validation

- \git diff --check\ passes.
- Inspected btleplug 0.13.0 platform implementations to verify MTU behavior across WinRT, BlueZ, and CoreBluetooth.
- No public API surface changes.
- Local Rust compilation and UniFFI binding generation were unavailable because the local environment does not have a Rust toolchain; the FFI integration will therefore rely on project CI validation.

---

> [!NOTE]
> **Medium Risk**
> Changes GATT write sizing on JVM/Desktop; incorrect MTU could cause oversized writes or under-reported limits, but scope is limited to the btleplug path with no public API changes.
> 
> **Overview**
> **JVM/Desktop** `maximumWriteValueLengthForType` no longer assumes a fixed ATT MTU of 23 (20 bytes payload). It now reads the **negotiated MTU** from btleplug and still subtracts the 3-byte ATT header.
> 
> The Rust **btleplug FFI** gains an `mtu()` export on `Peripheral` that forwards to the platform peripheral. Shared `Peripheral` KDoc notes that JVM/Desktop MTU is **automatically negotiated**, matching iOS behavior in the docs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8f4957f7a70af78d0e7242414a7ff5de6be45510. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kable-core/src/jvmMain/kotlin/com/juul/kable/btleplug/BtleplugPeripheral.kt
```diff
@@ -182,7 +181,9 @@ internal class BtleplugPeripheral(
     }
 
     override suspend fun maximumWriteValueLengthForType(writeType: WriteType): Int =
-        DEFAULT_ATT_MTU - ATT_MTU_HEADER_SIZE
+        withContext(Dispatchers.IO) {
+            ffi.mtu().toInt() - ATT_MTU_HEADER_SIZE
+        }
```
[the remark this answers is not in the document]
AvikMakwana: Fixed in d9682bc. JVM now falls back to the default ATT MTU payload size when the btleplug peripheral can’t be resolved, while preserving coroutine cancellation. Negotiated MTU is still used whenever it’s available.

kable-core/src/jvmMain/kotlin/com/juul/kable/btleplug/BtleplugPeripheral.kt
```diff
@@ -182,7 +182,15 @@ internal class BtleplugPeripheral(
...
-        DEFAULT_ATT_MTU - ATT_MTU_HEADER_SIZE
+        try {
+            withContext(Dispatchers.IO) {
+                ffi.mtu().toInt() - ATT_MTU_HEADER_SIZE
+            }
+        } catch (e: CancellationException) {
+            throw e
+        } catch (_: Exception) {
+            DEFAULT_ATT_MTU - ATT_MTU_HEADER_SIZE
+        }
```
twyatt: Although roughly equivalent, I think I prefer the following:

- using ensure active
- log the failure

```suggestion
        } catch (e: Exception) {
            currentCoroutineContext().ensureActive()
            logger.warn(e) { message = "Failed to get MTU, using default" }
            DEFAULT_ATT_MTU - ATT_MTU_HEADER_SIZE
        }
```

AvikMakwana: Updated in 4bc4e7d — switched to ensureActive() and added the warning before falling back to the default MTU. Thanks.

twyatt:
Looks like this is addressing the same thing as #1277.
I don't have a device to test/validate against at this time, but have you observed the behavior that #1277 describes? Where you have to wait until services are discovered before querying MTU?

PR #1277, opened 2026-09-13
Report the negotiated ATT MTU from the JVM peripheral

jamesarich:
## Problem

`BtleplugPeripheral.maximumWriteValueLengthForType` returns a constant `23 - 3`, so every JVM consumer sizes writes for the default MTU no matter what the link negotiated. A peripheral that negotiated 247 (or 517, which the Meshtastic firmware asks for) still gets 20-byte chunks, and a fragmenting protocol spends many times the round trips it needs.

btleplug already reports the value: `btleplug::api::Peripheral::mtu()` is populated by every backend Kable ships natives for - BlueZ reads the characteristic `MTU` property during service discovery, CoreBluetooth derives it from `maximumWriteValueLength(for:)` when `ServicesDiscovered` arrives, and WinRT stores it from `MaxPduSizeChanged`. It just wasn't exported through the ffi.

## Change

- `kable-btleplug-ffi`: export `Peripheral::mtu()`, returning the platform peripheral's negotiated MTU (btleplug's `DEFAULT_MTU_SIZE`, 23, until services have been discovered).
- `BtleplugPeripheral.maximumWriteValueLengthForType` forwards the query from `Connecting.Services` up (btleplug reports the default itself until discovery has run) and returns the default outside a connection, matching the Android backend's `mtu.value ?: DEFAULT_ATT_MTU`. An ffi error from a disconnect racing the query is folded into the default, so a plain query never throws.
- `Peripheral.maximumWriteValueLengthForType` KDoc gains the JVM line alongside the Android, iOS and JavaScript ones.

No public Kotlin API change; the JVM API dump is unchanged. The ffi gains one method, which the generated bindings pick up at build.

One sharp edge, in btleplug rather than here: on BlueZ older than 5.62 the `MTU` property is absent and btleplug's BlueZ backend `unwrap()`s it (`bluez/peripheral.rs:145`). That path was unreachable from Kable before this change. Current LTS distributions ship well past 5.62, so it is a concern for legacy hosts only; happy to add a `catch_unwind` guard in the ffi if you'd rather not expose it at all.

Not covered by tests: the JVM backend has no test source set, and the value is only observable against a real peripheral. Observed on Linux (BlueZ 5.87) against a Meshtastic radio, which asks for an MTU of 517, via a throwaway `jvmTest` that scans, connects and queries `maximumWriteValueLengthForType(WithoutResponse)` at each stage:

| Stage | Before this change | After |
|---|---|---|
| Before `connect()` | 20 | 20 |
| `Connecting.Services`, as `services` is first published | 20 | 514 |
| `Connected` | 20 | 514 |
| After `disconnect()` | 20 | 20 |

BlueZ agrees: polling the characteristic's `MTU` property over D-Bus on a `bluetoothctl` connection reads 23 until `ServicesResolved` flips to true and 517 from then on.

---

*Disclosure: drafted with an AI assistant (Claude Code), then reviewed by me. Motivated by [meshtastic-node-kmp](https://github.com/meshtastic/meshtastic-node-kmp), whose GATT transport sizes fragments from the live MTU per send. Verified locally with `cargoLintRust`, `lintKotlin`, `compileKotlinJvm`, `jvmApiCheck` and the hardware run above; `cargoLintClippy` fails on Linux `main` before this change, on an unused `serde` import in `peripheral_id.rs` that is only compiled under `cfg(target_os = "linux")`, so it never fires on the macOS runner where CI runs `check`.*

Suggested labels: `patch`, `jvm`.


kable-core/src/jvmMain/kotlin/com/juul/kable/btleplug/BtleplugPeripheral.kt
```diff
@@ -181,8 +181,14 @@ internal class BtleplugPeripheral(
         connectAction.cancelAndJoin(CancellationException(NotConnectedException("Disconnect requested")))
     }
 
-    override suspend fun maximumWriteValueLengthForType(writeType: WriteType): Int =
-        DEFAULT_ATT_MTU - ATT_MTU_HEADER_SIZE
+    override suspend fun maximumWriteValueLengthForType(writeType: WriteType): Int {
+        // btleplug learns the MTU during service discovery, so it is only meaningful while connected.
+        val mtu = when (state.value) {
+            is Connecting.Observes, is State.Connected -> withContext(Dispatchers.IO) { ffi.mtu().toInt() }
+            else -> DEFAULT_ATT_MTU
```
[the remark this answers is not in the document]
jamesarich: Fixed in e705d89: the query is now forwarded from `Connecting.Services` up (btleplug reports the default itself until discovery has run), and a sample taken the moment `services` is published reads 514 during `Connecting.Services` on a Meshtastic radio.

kable-core/src/jvmMain/kotlin/com/juul/kable/btleplug/BtleplugPeripheral.kt
```diff
@@ -181,8 +181,14 @@ internal class BtleplugPeripheral(
         connectAction.cancelAndJoin(CancellationException(NotConnectedException("Disconnect requested")))
     }
 
-    override suspend fun maximumWriteValueLengthForType(writeType: WriteType): Int =
-        DEFAULT_ATT_MTU - ATT_MTU_HEADER_SIZE
+    override suspend fun maximumWriteValueLengthForType(writeType: WriteType): Int {
+        // btleplug learns the MTU during service discovery, so it is only meaningful while connected.
+        val mtu = when (state.value) {
+            is Connecting.Observes, is State.Connected -> withContext(Dispatchers.IO) { ffi.mtu().toInt() }
```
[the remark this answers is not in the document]
jamesarich: Fixed in e705d89: an ffi error from a disconnect racing the query is folded into the default, so the query never throws.

kable-core/src/jvmMain/kotlin/com/juul/kable/btleplug/BtleplugPeripheral.kt
```diff
@@ -181,8 +183,21 @@ internal class BtleplugPeripheral(
         connectAction.cancelAndJoin(CancellationException(NotConnectedException("Disconnect requested")))
     }
 
-    override suspend fun maximumWriteValueLengthForType(writeType: WriteType): Int =
-        DEFAULT_ATT_MTU - ATT_MTU_HEADER_SIZE
+    override suspend fun maximumWriteValueLengthForType(writeType: WriteType): Int {
+        // btleplug reports the default until service discovery has run, so the query is forwarded for
+        // the whole life of a connection. A disconnect racing the query drops the ffi handle before
```
twyatt: > so the query is forwarded for the whole life of a connection

Is this a limitation/bug in btleplug?

jamesarich: Partly. On BlueZ it's the platform: the MTU is a property on the GATT characteristics, so there's nothing to read until services resolve. On CoreBluetooth it's btleplug, which samples `maximumWriteValueLength(for:)` once at `ServicesDiscovered` though CoreBluetooth would answer earlier. WinRT updates on `MaxPduSizeChanged`. So no backend has a real value before `Connecting.Services`, and each call goes to btleplug rather than being cached. Reworded the comment in eac531a.

kable-core/src/jvmMain/kotlin/com/juul/kable/btleplug/BtleplugPeripheral.kt
```diff
@@ -181,8 +183,21 @@ internal class BtleplugPeripheral(
         connectAction.cancelAndJoin(CancellationException(NotConnectedException("Disconnect requested")))
     }
 
-    override suspend fun maximumWriteValueLengthForType(writeType: WriteType): Int =
-        DEFAULT_ATT_MTU - ATT_MTU_HEADER_SIZE
+    override suspend fun maximumWriteValueLengthForType(writeType: WriteType): Int {
+        // btleplug only knows the negotiated MTU once services are discovered, so it is queried from
+        // `Connecting.Services` on rather than cached. A disconnect racing the query drops the ffi handle
+        // before `state` moves, which surfaces as an error folded into the default.
```
twyatt: Minor grammatical changes:

```suggestion
        // btleplug only knows the negotiated MTU once services are discovered, so it is queried from
        // `Connecting.Services` on, rather than being cached. A disconnect racing the query drops the ffi
        // handle before `state` moves, which surfaces as an error folded into the default.
```

twyatt (approved):
Looks good. I'll try to find time to test/validate these changes before merging.

kable-btleplug-ffi/src/peripheral.rs
```diff
@@ -215,6 +215,11 @@ impl Peripheral {
             .map(|p| p.services().into_iter().map(Into::into).collect())
     }
 
+    /// Negotiated ATT MTU, or the default (23) until services have been discovered.
+    async fn mtu(&self) -> Result<u16> {
+        self.get_platform().await.map(|p| p.mtu())
```
twyatt: nit


```suggestion
        self.get_platform()
            .await
            .map(|p| p.mtu())
```

kable-core/src/jvmMain/kotlin/com/juul/kable/btleplug/BtleplugPeripheral.kt
```diff
@@ -181,8 +183,21 @@ internal class BtleplugPeripheral(
...
+        val mtu = if (state.value.isAtLeast<Connecting.Services>()) {
+            try {
+                withContext(Dispatchers.IO) { ffi.mtu().toInt() }
+            } catch (e: FfiException) {
+                DEFAULT_ATT_MTU
```
twyatt: Can you add logging (`warn` level is probably appropriate) that the default is being fallen back to?

kable-core/src/jvmMain/kotlin/com/juul/kable/btleplug/BtleplugPeripheral.kt
```diff
@@ -181,8 +183,21 @@ internal class BtleplugPeripheral(
...
+            }
+        } else {
+            DEFAULT_ATT_MTU
```
twyatt: Can you add logging (`warn` level is probably appropriate) that the default is being fallen back to?

kable-core/src/jvmMain/kotlin/com/juul/kable/btleplug/BtleplugPeripheral.kt
twyatt: Also, should we still be subtracting `ATT_MTU_HEADER_SIZE`?

kable-core/src/jvmMain/kotlin/com/juul/kable/btleplug/BtleplugPeripheral.kt
twyatt: Also, should we still be subtracting `ATT_MTU_HEADER_SIZE`?

kable-btleplug-ffi/src/peripheral.rs
jamesarich: Done in b0fbd0e.

kable-core/src/jvmMain/kotlin/com/juul/kable/btleplug/BtleplugPeripheral.kt
jamesarich: Added in b0fbd0e.

kable-core/src/jvmMain/kotlin/com/juul/kable/btleplug/BtleplugPeripheral.kt
jamesarich: Added in b0fbd0e.

kable-core/src/jvmMain/kotlin/com/juul/kable/btleplug/BtleplugPeripheral.kt
jamesarich: Yes - btleplug returns the full ATT MTU (its CoreBluetooth backend adds the 3 back when converting `maximumWriteValueLength`), so the 517 the Meshtastic radio negotiates comes out as 514, same as the Android backend.

kable-btleplug-ffi/src/peripheral.rs
jamesarich: Reverted in fe6f3a8, `cargo fmt --check` folds it back onto one line and fails `cargoLintRust`.

